PR #4, opened 2020-12-31
Attempt one to implement superclassOf

matentzn:
This is probably not the right design, but I am not sure at the moment where exactly we want to handle transformations of the tsv values to rdf - for now I assume this goes in `io.py::to_rdf()`; so this also where I would 
1. handle conversion to existential restrictions and other graph patterns
2. handle external property mappings (say we want to annotate using standard properties, like dc:source, dc:contributor instead of sssom:source etc.

Let me know what you think the best way is here. I want to finish a feature complete prototype SSSOMtime in January :D

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,6 @@
 venv/
+sssom/__pycache__/
+.idea
+sssom.egg-info
+build/
+dist/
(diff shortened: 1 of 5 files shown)


matentzn:
Related ticket: https://github.com/cmungall/sssom-py/pull/4/files

cmungall (approved):

PR #15, opened 2021-02-08
Some small fixes to correlations and crosstab

matentzn:
- added verbose option to correlations (there is a lot being printed there -> would you want those tuples in the end to be somehow exported?
- change order of click options (duplicate -o/--output entries).


sssom/cli.py
```diff
@@ -114,38 +113,41 @@ def crosstab(input, output, transpose, fields):
 @main.command()
 @click.option('-o', '--output')
 @click.option('-t', '--transpose/--no-transpose', default=False)
+@click.option('-v', '--verbose/--no-verbose', default=False)
```
cmungall, line 116: there is a way in click to make verbose a global option.. we could do this.. or it may just complicate

cmungall (approved):

PR #29, opened 2021-04-29
Add filter method

matentzn:
No reviews yet, just junk.


matentzn:
Will redo this when the time comes

Issue #31, opened 2021-04-29
Potential problem with read_pandas: # in rows


matentzn:
Currently we ar using

```
pd.read_csv(filename, comment='#', sep=sep).fillna("")
```

To read a pandas data frame. This can be risky, due to the potential of hash symbols in the middle of rows. It would be more robust to manually skip rows which start with # and read the rest.

matentzn:
@hrshdhgd can you check this is actually aproblem? if so just try and implement the above?

hrshdhgd:
You are correct! When a '#' occurs in the data, pandas just turns everything after that as a 'NaN'

e.g.
I took part of your [test file](https://github.com/mapping-commons/sssom-py/blob/master/tests/validate_data/basic.tsv) as an input and added 2 '#'s in the first and third row.
```
#comment: .
#creator_id: cjm
#curie_map:
#  a: http://example.org/a/
#  rdfs: http://www.w3.org/2000/01/rdf-schema#
#  x: http://example.org/x/
#  y: http://example.org/y/
#  z: http://example.org/z/
#license: https://creativecommons.org/publicdomain/zero/1.0/
#mapping_date: '2020-05-30'
#mapping_tool: https://github.com/cmungall/rdf_matcher
#object_match_field: rdfs:label
#object_source: y
#subject_match_field: rdfs:label
#subject_source: z
subject_id	subject_label	subject_category	predicate_id	object_id	object_label	object_category	match_type	subject_source	object_source	mapping_tool	confidence	subject_match_field	object_match_field	match_string	comment
x:appendage	appendage	#biolink:AnatomicalEntity	owl:equivalentClass	y:appendage	appendages	biolink:AnatomicalEntity	Lexical|Stemming	x	y	rdf_matcher	0.8407144059077161	rdfs:label	rdfs:label	appendag	.
x:appendage	appendage	biolink:AnatomicalEntity	owl:equivalentClass	z:appendage	APPENDAGE	biolink:AnatomicalEntity	Lexical|Stemming	x	z	rdf_matcher	0.8407144059077161	rdfs:label	rdfs:label	appendag	.
x:appendage	appendage	#biolink:AnatomicalEntity	owl:equivalentClass	z:appendage	APPENDAGE	biolink:AnatomicalEntity	Lexical	x	z	rdf_matcher	0.8818562360532485	rdfs:label	rdfs:label	appendage	.
[...]

PR #35, opened 2021-04-29
Issue31 fix

hrshdhgd:
refactored pandas' `pd.read_csv` to our customized `read_pandas` in an effort to address #31


matentzn (changes requested):
Awesome, thanks! maybe double check my comments :) Thank you!

Also the qc is failing! You can test this locally by running pytest.

PR #37, opened 2021-05-04
Issue31

hrshdhgd:
Lines that begin with a '#' are treated as comments and the rest as considered to be part of the data.


hrshdhgd:
Addresses #38 too.

matentzn (approved):
Great! Thanks!

PR #50, opened 2021-05-24
Cli edit

hrshdhgd:
- Documentation for the click options.


sssom/cli.py
```diff
@@ -31,51 +52,86 @@ def main(verbose, quiet):
...
+@click.option('-o', '--output', type=click.Path(), help= help_output)
 def parse(input: str, input_format: str, metadata:str, curie_map_mode: str, output: str):
-    """
-    parse file (currently only supports conversion to RDF)
-    """
+    """parse file (currently only supports conversion to RDF)
+
+    Args:
+
+        input (str): Input filename.
```
cmungall, line 92: this is redundant with what is declared above. Is it necessary?

Args will be shown anyway by click if you say `--help`

I think it is better to use this space to give examples, etc

sssom/cli.py
```diff
@@ -31,51 +52,86 @@ def main(verbose, quiet):
...
+@click.option('-o', '--output', help= help_output)
+@click.option('-t', '--to-format', help=help_format)
+@click.option('-c', '--context', help=help_context)
 def convert(input: str, output: str, format: str, to_format: str, context: str):
-    """
-    convert file (currently only supports conversion to RDF)
-    """
+    """Convert file (currently only supports conversion to RDF)
+
+    Args:
```
cmungall, line 63: use consider `\b` for formatting

matentzn: also, we should use one docstring scheme throughout; this looks like the google one, which is different from the one used elsewhere.. I don't care which, just make sure its the same throughout.

hrshdhgd: I agree. But since we will be using this space for examples as Chris suggested before, this would look like:

```
[Description]

[Example]
```

Does the doctoring scheme differentiate how it displays?

matentzn: I dont know - I wouldnt think so, but probably worth a google (especially for systems that use docstrings that autogenerate cool yaml docs).

hrshdhgd: Tried `\b`, does not work:
Screen Shot 2021-05-27 at 10 56 15 AM

I am replacing everything under `Args` (including Args) by `Examples`. I will need help from @cmungall and/or @matentzn to fill in the blanks [A B C ... as placeholders below]

Screen Shot 2021-05-27 at 10 59 48 AM

sssom/cli.py
```diff
@@ -15,11 +15,33 @@
 import logging
 from pandasql import sqldf
 
+# Click 'help' variabes
+help_input = 'Input file. For e.g.: SSSOM tsv file'
```
cthoyt, line 19: if you leave the actual help text so far away from the options themselves, they can't serve as documentation for the code. If you need to reuse an option in several functions, you could always save the option into a variable.

hrshdhgd: Thank you so much for the suggestion but I do not follow. Would it be possible to direct me to an example to do this the right way?

matentzn: @cthoyt this is a constant question mark for me in CLI development, see debates above. IDEs like pycharm wont read the click help text in any case, so we wont get the nice autocomplete help text when typing code no matter where the help text actually is. On the other hand, the CLI code (in this case click) needs to have the help so the user can get it when `--help`. Now: Is there any way to prevent this redundancy?

cthoyt: A CLI should implement as little logic as possible. Offload as much as you can to a helper function

cthoyt: Then the helper function can be idiomatic and pythonic, and the cli can just focus on click

matentzn: I dont understan. Consider this method:

```
@main.command()
@click.option('-i', '--input', required=True, type=click.Path(), help='The input file (URL or local) in SSSOM format.')
@click.option('-d', '--output-directory', type=click.Path(),help='The directory to which the separate mappings will be exported')
def split(input: str, output_directory: str):
    """
    Reading a single table in SSSOM format and splitting int by prefixes.
    
    :param input: The input file (URL or local) in SSSOM format.
    :param output_directory: The directory to which the separate mappings will be exported
    """
    split_file(input_path=input, output_directory=output_directory)
```

How can I avoid the redundancy of the docstring and click help text?

cthoyt: The first line of the docstring is propagated to the command line help, but the rest of the docstring should be dropped.

sssom/cli.py
```diff
@@ -15,11 +15,33 @@
 import logging
 from pandasql import sqldf
```
cthoyt, line 17: @hrshdhgd here, you can do something like this:

```suggestion
input_option = click.option('-i', '--input', help=help_input)
```

Later, inside your commands, you can use `@input_option` instead of `@click.option('-i', '--input', help=help_input)`

sssom/cli.py
matentzn: And tools like pycharm and others would recognise click options as docs, or is it just that because you know that you will never use the CLI command in pycharm, you can omit them here in favour of the docstrings?

cthoyt: It's because you know you will never call a click function from any other python code or in PyCharm, exactly

matentzn: Thank you @cthoyt super interesting and lesson learned :)

hrshdhgd: How about showing an example there? 

so instead of 
```
:param input: The input file (URL or local) in SSSOM format.
:param output_directory: The directory to which the separate mappings will be exported
```

Show an example of how the command is used correctly? Do you think that would be useful for users?

cthoyt: I don't think the docstring of a command line is really the place where people show examples. You have manpages for most CLI-only tools, and SSSOM should also have user documentation with Sphinx. This is where usage and examples should go.

hrshdhgd: Thank you so much @cthoyt !!

hrshdhgd:
@matentzn , All I've done so far is changed docstrings. Any idea why the build would be failing with the error `ModuleNotFoundError: No module named 'linkml.utils.slot'` I have neither added any library requirement nor changed any dependencies in the project.

matentzn (changes requested):
Before we merge this I would like:

At least one, better two or three tests per method in here (make TESTDIR configurable):
https://github.com/mapping-commons/sssom-py/blob/cli_edit/tests/tests.sh

Make sure all CLI methods are tested here: 
https://github.com/mapping-commons/sssom-py/blob/cli_edit/tests/test_cli.py

I made a bunch of change to the annotations, but we wont make this perfect now; the goal here is to have cli in an ok state that we can iterate over.

tests/tests.sh
```diff
@@ -1,4 +1,22 @@
 set -e
 
-sssom parse --input /Users/matentzn/ws/sssom-py/tests/data/basic.tsv --output /Users/matentzn/ws/sssom-py/tests/tmp/basic.tsv.tsv --input-format tsv --curie-map-mode merged
```
cthoyt: It's possible to write proper unit tests for CLIs using the CliRunner in click
https://click.palletsprojects.com/en/8.0.x/testing/

but it's better to write tests for the functions the CLI wraps

sssom/cli.py
```diff
@@ -31,77 +56,149 @@ def main(verbose, quiet):
...
     """
-    df = parse(input)
-    df = collapse(df)
+    msdf = from_tsv(input)
+    # df = parse(input)
+    df = collapse(msdf.df)
     # , priors=list(priors)
     rows = dataframe_to_ptable(df)
     for row in rows:
         logging.info("\t".join(row))
```
cmungall, line 171: I guess I am responsible for this... what was I thinking. This isn't logging output, it's the primary output

tests/tests.sh
matentzn: Oh we have that as well - I just think we want both :) 
https://github.com/mapping-commons/sssom-py/blob/cli_edit/tests/test_cli.py

matentzn: We also have tests for the functions that the CLI wraps: 
https://github.com/mapping-commons/sssom-py/blob/cli_edit/tests/test_conversion.py

sssom/cli.py
matentzn: This is my mistake, not yours.. I want to generally be able to stream outputs from all commands in that way; what is the best way to do this then - using print() in conjunction with a stream (either a file output stream or stdout)?

matentzn: @hrshdhgd I will fix this

sssom/cli.py
matentzn: We will do this bit by bit in a different PR

hrshdhgd:
I have committed tests as well.

Things that are pending:

1. `tests.sh` : test for `sparql`
2. 'test_cli.py`: tests for the following
    - `dosql` : requires a query. I'm not sure if the queries in `test_config.yaml` actually apply here. Please correct me if I'm wrong.
    - `sparql`
3. `test_config.yaml`: Lines 11-19. I just copy-pasted `basic.tsv` properties to `basic2.tsv` **and hence the CI is failing**. I would like some guidance on how those values are determined.
   - ct_json_elements
   - query_count_equivalent_classes
   - query_count_equivalent_classes

See below:

```
  - filename: "basic2.tsv"
    inputformat: "tsv"
    multiple_input: True
    ct_json_elements: 176 **#need to confirm how to get this**
    ct_data_frame_rows: 120
    ct_graph_queries_owl:
      query_count_equivalent_classes: 86 **#need to confirm how to get this**
    ct_graph_queries_rdf:
      query_count_equivalent_classes: 86 **#need to confirm how to get this**
```

sssom/cli.py
matentzn: for ptable we are now using print, but I want to ensure with the architecture so will review with @cmungall on Monday and merge.

matentzn:
Alright this has become too massive, sorry about that @cmungall and I will review this on Monday together and merge it in.

cmungall:
go ahead and merge!

PR #52, opened 2021-05-31
Trying to add mapping_cardinality enum

matentzn:
There are quite a few odditities here, also discussed on slack.

@cmungall check how the enums are represented throughout the board, especially OWL and python data classes. Is that what you would expect?


matentzn:
Made ticket

PR #71, opened 2021-06-22
Apply `black` and add tox configuration

### LINKED ISSUE #64: Apply `black` code style
Even though it takes a bit of getting used to, the [`black`](https://github.com/psf/black) code formatter is proving to be a godsend on projects with multiple authors that all write python with slightly different styles.

I'm using it in all of my collaborative projects now, I'd suggest setting it up here. It's also possible to use a pre-commit hook so people can write code however they want, then it automatically fixes it just before pushing to github. Then you can also add a CI action to check that black has been applied before accepting PRs.

Let me know if you want me to add the config/docs that would help you get started. In the mean time, you could always see what happens when you do:

```bash
$ pip install black
$ black sssom/ tests/
```

cthoyt:
Closes #64 

If desired, I could also add a GitHub Actions workflow to check that black has been applied.


matentzn:
Awesome, thank you! Can you
1) Document how exactly we can apply black to everything? Could you perhaps add a goal to the makefile like 

```
lint:
      black *.py
```

so that we can run this ourselves?
2) Document what Tox is and how we are expected to interact with it?

You can use our existing GH workflow if you want to do checking for `black`, I think this is a good idea. Thank you again!

matentzn:
Oh my god this is _so_ cool! Love it! Let me know when you are happy with the PR to be merged! I am already!!

cmungall:
This is awesome

Why shouldn't we check autogen code?

Does flake8 or black check typing or should that be done separately with
mypy?

matentzn:
@hrshdhgd I will do this - needed to check some related things..

matentzn (approved):

Issue #51, opened 2021-05-31
New command: sort


matentzn:
See  https://github.com/mapping-commons/SSSOM/issues/39

We need a command that uses the canonical column order to sort an entire mapping set data frame correctly.

cmungall:
I think this lower priority now we can just do `sssom dosql SELECT * FROM match ORDER BY ...`

matentzn:
We need to be able to sort the columns by their canonical order, with a method like
https://github.com/mapping-commons/sssom-py/pull/158/files#diff-8e0eb781f6bf2cf74b9c6a904555d8bd7f214fef45ea8e55da8527f834e600e4L191

I am happy if the sort function could simply build on the the dosql command,  in combination with a canonical column ordering.

hrshdhgd:
Does this sequence seem correct?

```
['subject_id', 'subject_label', 'subject_category', 'predicate_id', 'predicate_modifier', 'object_id', 'object_label', 'object_category', 'match_type', 'subject_source', 'object_source', 'mapping_tool', 'confidence', 'subject_match_field', 'object_match_field', 'match_string', 'comment']
```

This is the sequence in which slots are declared in [sssom.yaml](https://github.com/mapping-commons/sssom-py/blob/master/sssom/sssom.yaml)

hrshdhgd:
It seems like `Mapping(**mdict)` already sorts the column names in the order it appears in `sssom.yaml` as slots thus keeping it a standard sequence. So a `sort` method is not needed(?). 

cc: @matentzn , @cmungall

matentzn:
Sort needs to sort also the rows!

This here seems wrong:
https://github.com/monarch-initiative/mondo/blob/master/src/ontology/mappings/mondo_broadmatch_icd10cm.sssom.tsv

You can see it has a subject_label right at the end, while I would expect it to be after subject_id!

hrshdhgd:
Aah .. I see why that's happening. Easy fix!

hrshdhgd:
Now for rows, what would be the sort-by rule?

hrshdhgd:
[Columns have been sorted!](https://github.com/monarch-initiative/mondo/blob/do-sql-makefile/src/ontology/mappings/mondo_broadmatch_icd10cm.sssom.tsv)

PR #72, opened 2021-06-22
Issue 56

hrshdhgd:


hrshdhgd:
Still have to write tests for `merge_msdf` as [seen here](https://github.com/mapping-commons/sssom-py/blob/c033ec81db4a017dbb72fb24de7f041d51a366f1/sssom/util.py#L230).


@matentzn , I had to place this code in `utils.py` to avoid circular imports and other errors. Please review. I have commented out the `inplace` parameter for now. Need some discussion around it which we can address during our next meeting.

sssom/util.py
```diff
@@ -21,6 +22,10 @@
 OBJECT_SOURCE = 'object_source'
```
matentzn, line 22: All of these should probably be in a separate python file just for static variables.

sssom/util.py
```diff
@@ -187,7 +192,9 @@ def dataframe_to_ptable(df: pd.DataFrame, priors=[0.02, 0.02, 0.02, 0.02], inver
         elif p == 'dbpedia-owl:different':
             pi = 3
         else:
-            raise Exception(f'Unknown predicate {p}')
```
matentzn, line 190: Why?

sssom/util.py
```diff
@@ -219,3 +226,153 @@ def sha256sum(filename):
         for n in iter(lambda: f.readinto(mv), 0):
             h.update(mv[:n])
     return h.hexdigest()
+
+def merge_msdf(msdf1:MappingSetDataFrame, msdf2:MappingSetDataFrame, reconcile:bool=True, inplace:bool=False) -> MappingSetDataFrame:
```
matentzn, line 230: didnt we have this being a method in the msdf rather than as a utility method? similar to pd.merge()?

matentzn (commented):
I made a few comments but looking forward to dealing with this in more depths in our next call!

sssom/util.py
hrshdhgd: Discussed this with Chris. As of now the lines above are addressing `p = ` hardcoded predicates (handful) . I do not completely understand that part of the code completely. This is a temporary fix I made because I had used new predicates as a test due to which it was `throw`ing errors. For e.g. `!owl:equivalentClass`

sssom/util.py
hrshdhgd: I started it as being part of Class `msdf` but when I had to import `filter_redundant_rows`, it threw an error (I think something about importing x before y and hence may behave unexpectedly .. ) and not doing anything. So I moved the methods where `filter_redundant_rows` was located. We could always put it somewhere else.

hrshdhgd:
Yes sir!

sssom/util.py
```diff
@@ -219,3 +415,246 @@ def sha256sum(filename):
...
+        negation_df = df.loc[df[PREDICATE_ID].str.startswith('!')]# or df.loc[df['predicate_modifier'] == 'NOT']
+
+        # #####This step ONLY if 'NOT' is expressed by the symbol '!' in 'predicate_id' #####
+        normalized_negation_df = negation_df.reset_index()
+        normalized_negation_df[PREDICATE_ID] = normalized_negation_df[PREDICATE_ID].str.replace('!','')
+        ########################################################
+        normalized_negation_df = normalized_negation_df.drop(['index'], axis = 1)
+
+        # remove the NOT rows from the main DataFrame
+        condition = negation_df.isin(df)
```
matentzn, line 498: Shouldnt this use "_is_defining_feature?"

matentzn: i see now what this is doing, scratch that.

hrshdhgd:
@matentzn , I think this PR is ready to merge. I will work on #51 on a separate branch and PR. Thoughts?

sssom/util.py
```diff
@@ -24,6 +31,192 @@
...
+    def merge(self, msdf2):
+        """Merges two MappingSetDataframes
+
+        Args:
+            msdf2 (MappingSetDataFrame): Secondary MappingSetDataFrame (self => primary)
+
+        Returns:
+            MappingSetDataFrame: Merged MappingSetDataFrame
+        """
+        merge_msdf(msdf1=self, msdf2=msdf2)
```
matentzn, line 57: missing:

```
 msdf = merge_msdf(msdf1=self, msdf2=msdf2)
self.df = msdf.df
self.prefixmap = msdf.prefixmap
self.metadata - msdf.metadata
```

sssom/util.py
```diff
@@ -237,3 +445,275 @@ def sha256sum(filename):
...
+        # If msdf2 has a DataFrame
+        if msdf2.df is not None:
+            # 'outer' join in pandas == FULL JOIN in SQL
+            merged_msdf.df = msdf1.df.merge(msdf2.df, how='outer')
+        else:
+            merged_msdf.df = msdf1.df
+        #merge the non DataFrame elements
+        merged_msdf.prefixmap = dict_merge(msdf2.prefixmap, msdf1.prefixmap, 'prefixmap')
+        # After a Slack convo with @matentz, commented out below.
+        #merged_msdf.metadata = dict_merge(msdf2.metadata, msdf1.metadata, 'metadata')
```
matentzn, line 476: The metadata needs to be added as column values in though? Else we lose it!

matentzn (changes requested):

sssom/util.py
hrshdhgd: The last three lines already happen in the function `merge_msdf`

hrshdhgd: Added.

sssom/util.py
hrshdhgd: As per our discussion on Slack or rather as far as I understood from it (and the test example `basic.tsv`), was that metadata was generated from the columns of the `msdf.DataFrame` itself, correct? So they are already in the `msdf.DataFrame`?

sssom/util.py
hrshdhgd: Done! https://github.com/mapping-commons/sssom-py/pull/72/commits/eb884e4a543da7a0ad872569f2c711922281c8b9

sssom/util.py
hrshdhgd: Done! https://github.com/mapping-commons/sssom-py/pull/72/commits/eb884e4a543da7a0ad872569f2c711922281c8b9

sssom/util.py
hrshdhgd: Where do you suggest these be?

cmungall:
these should be introspectable from the gen py classes

matentzn (approved):

Issue #70, opened 2021-06-17
Create function to clean prefix map according to actual prefix usage in dataframe


matentzn:
Sometimes, there are more prefixes than necessary in the prefix map and we may want to clean them out.

I propose to add a method like this to MappingSetDataFrame:

```
    def clean_prefix_map(self):
        prefixes_in_map = get_prefixes_used_in_table(self.df)
        new_prefixes = dict()
        missing_prefix = False
        for prefix in prefixes_in_map:
            if prefix in self.prefixmap:
                new_prefixes[prefix]=self.prefixmap[prefix]
            else:
                logging.warning(f"{prefix} is used in the data frame but does not exist in prefix map")
                missing_prefix = True
        if not missing_prefix:
            self.prefixmap=missing_prefix
```

PR #82, opened 2021-07-08
Adding "clean prefixes" parameter

matentzn:
fixes https://github.com/mapping-commons/sssom-py/issues/70

### DIFF
--- a/sssom/cli.py
+++ b/sssom/cli.py
@@ -142,11 +142,22 @@ def convert(input: str, output: str, output_format: str):
     help="Defines wether the curie map in the metadata should be extended or replaced with "
     "the SSSOM default curie map. Must be one of metadata_only, sssom_default_only, merged",
 )
-@click.option('-p', '--clean-prefixes', default=True, is_flag=True, required=True,
-              help='If True (default), records with unknown prefixes are removed from the SSSOM file.')
+@click.option(
+    "-p",
+    "--clean-prefixes",
+    default=True,
+    is_flag=True,
+    required=True,
+    help="If True (default), records with unknown prefixes are removed from the SSSOM file.",
+)
 @output_option
 def parse(
-    input: str, input_format: str, metadata: str, curie_map_mode: str, clean_prefixes: bool, output: str
+    input: str,
+    input_format: str,
+    metadata: str,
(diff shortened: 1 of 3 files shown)


sssom/cli.py
```diff
@@ -139,9 +139,22 @@ def convert(input: str, output: str, output_format: str):
     help="Defines wether the curie map in the metadata should be extended or replaced with "
     "the SSSOM default curie map. Must be one of metadata_only, sssom_default_only, merged",
 )
+@click.option(
+    "-p",
+    "--clean-prefixes",
+    default=True,
+    is_flag=True,
```
hrshdhgd, line 146: Where is this parameter (`is_flag`) going? Shouldn't this be part of `parse` below?

sssom/util.py
```diff
@@ -815,3 +816,21 @@ def get_prefixes_used_in_table(df: pd.DataFrame):
...
+            prefix = get_prefix_from_curie(v)
+            if prefix in filter_prefixes:
+                ok = False
+                break
+        if ok:
+            rows.append(row)
+    if rows:
+        return pd.DataFrame(rows)
+    else:
+        return pd.DataFrame(columns=_defining_features)
```
hrshdhgd, line 836: Could we rewrite this like this ? Just a thought - less code.

```
def filter_out_prefixes(df: pd.DataFrame, filter_prefixes) -> pd.DataFrame:
     rows = []
     for index, row in df.iterrows():
         for col in _defining_features:
             v = row[col]
             prefix = get_prefix_from_curie(v)
             if prefix not in filter_prefixes:
                 rows.append(row)
     if rows:
         return pd.DataFrame(rows)
     else:
         return pd.DataFrame(columns=_defining_features)
```

hrshdhgd (commented):
A question and a suggestion. Please let me know if I'm misunderstanding anything.

sssom/cli.py
matentzn: `is_flag` means that you can just say `--clean-prefixes` instead of `--clean-prefixes true`. Does this make sense?

sssom/util.py
matentzn: Ok that makes sense. Done

PR #83, opened 2021-07-08
Adding toString method to msdf

matentzn:
Not sure this is the right way to do this; or even if we would want this.


hrshdhgd (approved):
Looks good to me. Not sure when we'll need this but it may be good to have.

cmungall (approved):

PR #86, opened 2021-07-08
Running tests.sh when runniong make test

matentzn:


hrshdhgd (approved):
Great catch! I missed these, sorry about that!

PR #93, opened 2021-07-09
Make QC run on pull requests

cthoyt:
This PR changes the github actions config so it also runs on pull requests


matentzn (approved):
Thank you @cthoyt

PR #94, opened 2021-07-09
OLD Cliquesummary

matentzn:
From @cmungall maybe he wants to delete this


matentzn:
@cmungall needs to decide what to do with this code. Shall we delete it? work it in?

cmungall:
Seems like this pr is for a one line change?

matentzn:
huuu! It didnt used to be :D

PR #96, opened 2021-07-11
General refactor (renaming, import/export)

matentzn:
renaming methods etc


tests/test_cli.py
```diff
@@ -74,12 +75,13 @@ def run_convert(self, runner, test_case: SSSOMTestCase):
             "owl",
         ]
         result = runner.invoke(convert, params)
-        self.run_successful(result)
+        print(params)
```
hrshdhgd, line 78: do we need this `print`?

tests/test_cli.py
```diff
@@ -132,29 +134,35 @@ def run_dedupe(self, runner, test_case: SSSOMTestCase):
...
-        result = runner.invoke(diff, params)
-        print(result)
-        self.run_successful(result)
-        return result
-
-    def run_partition(self, runner, test_case: SSSOMTestCase):
+        if out_file:
+            params.extend(["--output", out_file.get_out_file("tsv")])
+            result = runner.invoke(diff, params)
+            print(result)
```
hrshdhgd, line 148: `print`

tests/tests.sh
```diff
@@ -18,6 +18,13 @@ sssom split $INPUT_URL_1 --output-directory $OUTPUT_DIR
 sssom convert $INPUT_FILE_1 --output $OUTPUT_DIR/converted_basic.tsv --output-format tsv
 sssom convert $INPUT_URL_1 --output $OUTPUT_DIR/converted_basic.tsv --output-format tsv
 
+for fmt in tsv json owl rdf
+do
+  sssom convert $INPUT_FILE_1 --output $OUTPUT_DIR/converted_basic.tsv --output-format $fmt
```
hrshdhgd, line 23: This will output all files with extension '.tsv'. 
Proposed change:
```
sssom convert $INPUT_FILE_1 --output $OUTPUT_DIR/converted_basic.$fmt --output-format $Mt
```

I tried this but the .rdf file shows an error in Mozilla.

hrshdhgd: Committed this change.

hrshdhgd:
`correlation` and `crosstab` test outputs (from `tests.sh`) are the same. Is that expected?

tests/test_cli.py
matentzn: Thanks :) fixed

hrshdhgd:
Found a new issue where the `#` in `owl: "http://www.w3.org/2002/07/owl#"` located in the test file [here](https://github.com/mapping-commons/sssom-py/blob/46c4b4d609418edd4fbe32f20bde538e426efd98/tests/data/basic.tsv#L8) appears when we use the local file but disappears when using the URL for the test: 

`sssom merge $INPUT_URL_1 $INPUT_URL_2 $INPUT_URL_3 -o $OUTPUT_DIR/merged_msdf_url.tsv`

Due to this it is throwing an error: `ValueError: prefixmap values in both MappingSetDataFrames for the same key [owl] are different.` Not sure of the '#' is important to be present or no?

hrshdhgd:
This is due to the code present [here](https://github.com/mapping-commons/sssom-py/blob/100cfeb873d7e4a71cd023700b5a3de7e3c62276/sssom/util.py#L747) where `v = 'http://www.w3.org/2002/07/owl#'` and `target[k] = 'http://www.w3.org/2002/07/owl'` where `k = owl `

matentzn:
`http://www.w3.org/2002/07/owl` is totally wrong! Thanks for catching that!

matentzn:
`http://www.w3.org/2002/07/owl` is totally wrong! Thanks for catching that!

hrshdhgd:
So `sssom parse` using a URL does not capture the `#`

matentzn:
I think that this is a more general issue: Given

```
# curie_map:
#  owl: http://www.w3.org/2002/07/owl#
#  rdfs: "http://www.w3.org/2000/01/rdf-schema#"
#  skos: "http://www.w3.org/2004/02/skos/core#"
```

The parser will think that the trailing "#" is a comment!

matentzn:
Cna you verify that?

hrshdhgd:
But then it is also in the tsv present locally but it reads it properly. I'll double check that.

hrshdhgd:
[This ](https://github.com/mapping-commons/sssom-py/blob/100cfeb873d7e4a71cd023700b5a3de7e3c62276/sssom/parsers.py#L514)is causing it.

PR #98, opened 2021-07-12
accommodating for blank confidence #97

### LINKED ISSUE #97: merge method does not work when no confidence present
the merge method should allow the case that a data frame provides no confidence. We assume confidence 1 when there is no confidence given.

hrshdhgd:
Fixes #97

### DIFF
--- a/sssom/parsers.py
+++ b/sssom/parsers.py
@@ -37,6 +37,8 @@ def from_tsv(
     """
     if validators.url(file_path) or os.path.exists(file_path):
         df = read_pandas(file_path)
+        if 'confidence' in df.columns:
+            df['confidence'].replace(r'^\s*$', 1, regex=True, inplace=True)
         if meta is None:
             meta = _read_metadata_from_table(file_path)
         if "curie_map" in meta:
--- a/sssom/parsers.py
+++ b/sssom/parsers.py
@@ -22,6 +22,7 @@ from .util import (
     is_curie,
 )
 from .util import RDF_FORMATS
+import numpy as np
 
 cwd = os.path.abspath(os.path.dirname(__file__))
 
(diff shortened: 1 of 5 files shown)


sssom/sssom_datamodel.py
```diff
@@ -299,7 +299,9 @@ def __post_init__(self, *_: List[str], **kwargs: Dict[str, Any]):
         if self.mapping_date is not None and not isinstance(self.mapping_date, str):
             self.mapping_date = str(self.mapping_date)
 
-        if self.confidence is not None and not isinstance(self.confidence, float):
+        if self.confidence == '' or self.confidence is None:
```
matentzn, line 302: Unfortunately sssom_datamodel.py cannot be manually editted, the other comment..

matentzn, line 302: Unfortunately you cant edit sssom/sssom_datamodel.py manually, since its generated, see other comments.

matentzn (changes requested):

sssom/sssom_datamodel.py
hrshdhgd: rolling it back.Thanks!

hrshdhgd:
This fails in `diff` as well. Do you think there is a way to autogenerate this logic in `sssom_datamodel.py` such that it is taken care of for all commands (even for the ones in the future) rather than individually doing it for every command (`merge`, `diff` etc.).

matentzn:
I tried using LinkML default values - that _could_ do it, but seems extremely risky.. Like assuming _in general_ that confidence is 1.0.. The default confidence should be unknown..

cant we just add an if else to the merge() function? what would be the risk?

hrshdhgd:
thoughts on the latest commit? https://github.com/mapping-commons/sssom-py/issues/97

hrshdhgd:
Edited code to reflect our Slack conversation. Thoughts?

hrshdhgd:
Tests pass now.

matentzn:
Can you resolve the conflicts please (rebasing with master)?

hrshdhgd:
Done

hrshdhgd:
Done

matentzn:
What is the state of this pull request - did you convince yourself that the merging still works well? If so, what is the reason for changing the tests exactly (see slack discussion?)

hrshdhgd:
Yes, this works well. The test was changed because when I had introduced the blank confidence value, it was done in another branch. We rolled it back since it wasn't part of that PR but this PR. So I made that change here.

matentzn (approved):

matentzn:
Thank you very much @hrshdhgd

PR #103, opened 2021-07-20
--input should accept URLs along with local paths

hrshdhgd:
For testing , I just use the GitHub url to the same files that exist locally.


matentzn:
I will put this to work in progress while the piping does not work yet :) Let me know when its good for review.

hrshdhgd:
For now, it just does a `sys.stdout(str(meta))` and `sys.stdout(str(df))` id output parameter is not provided.

PR #104, opened 2021-07-20
Bugfix: Missing 'validators' install

joeflack4:
Added missing dependency 'validators' to pipfile and setup.cfg. This is currently in requirements.txt and being imported.


matentzn:
Thank you @joeflack4 !

PR #108, opened 2021-07-27
Sphinx docs

hrshdhgd:
New sphinx docs for the project


matentzn (approved):
This is really nice @hrshdhgd 

I have two questions:
1. Do all this static files _static, _modules etc belong on master, or should they be on a different branch (gh-pages) - think of this question as a general one for all repos.
2. for this:


```
.. click:: sssom.cli:parse
    :prog: parse
    :nested: full
```

Can you find out how we would document _all_, not just the `parse`, methods at once? Thank you!

hrshdhgd:
For Q1, these files are autogenerated by sphinx. I believe they are supporting files for the HTML. So having them in the main branch would be advisable. I'll try taking them out and see if they make any difference to the docs or no.

As for Q2, I will dig deeper and figure it out.

matentzn:
They definitely have to be on your gh-pages branch, that's certain; I usually .gitignore autogenerated HTML content on the main branch. I just don't know whether that is possible here.

matentzn:
I am so happy. Merge at your own discretion!

hrshdhgd:
Q2 taken care of : https://mapping-commons.github.io/sssom-py/cli_usage.html

I'll look into the .gitignore portion. I'm glad this is to your satisfaction!

hrshdhgd:
So are we planning on keeping the `sphinx_docs` branch active throughout just so that the documentation is hosted from that branch? The reason I ask this is if we add the folders within docs [_modules, _static, _source] , then it messes the webpage since it contains supporting files needed for the webpage to display as it does now. Else it looks like this:

Screen Shot 2021-07-28 at 12 55 18 PM

hrshdhgd:
I will wait on your response before merging this branch.

PR #113, opened 2021-08-23
Migrated files from charlies PR

matentzn:
This is the exact same pull request as #90 

it was erroneously created because of my inability to work with a forked repo i didn't have write rights to - and which had conflicts with master. Now I know better :)

PR #114, opened 2021-08-23
Update installation (redux)

cthoyt:
This is a re-do of #90, which got merged into #113 and created a bit of confusion.  The original text read:

> This PR updates the way requirements are installed. Notably, it removes the `requirements.txt`, `Pipfile`, and `Pipfile.lock` since those are typically only used for web applications and one-off scripts. There's the little caveat where this means some installation dependencies can get mixed up, but Python is provably bad at this, and it's a crapshoot anyway (especially if you expect someone to install your software).
>
> The requirements for docs and tests get put in their own `extras` inside the `setup.cfg`. All installation is changed to use pip. Tox allows you to specify which extras you want (should be obvious from the diff)


cthoyt:
@matentzn @hrshdhgd can you please define what the acceptance/merge criteria for this PR are? I'm worried that this will sit too long and become stale again

matentzn:
@cthoyt I was just working on it.. :D Mind reading.

cthoyt:
@matentzn woohoo!!

Issue #121, opened 2021-08-30
Is cliquesummary.py autogenerated?


matentzn:
If so, we need to take care that this is reinstaed in the deploy-dm goal of the makefile.

In this PR I had to fix something in cliquesummary.py that was no longer compliant with linkml for some reason:
https://github.com/mapping-commons/sssom-py/pull/96/files

PR #127, opened 2021-08-31
Fixing flake8 and black issues

matentzn:


sssom/cliquesummary.py
```diff
@@ -47,111 +69,125 @@ class Count(int):
...
+    """a date (year, month and day) in an idealized calendar"""
+
     type_class_uri = XSD.date
     type_class_curie = "xsd:date"
     type_name = "date"
     type_model_uri = SSSOM.CS.Date
 
 
 class Datetime(XSDDateTime):
-    """ The combination of a date and time """
```
cthoyt, line 114: @matentzn how are these generated? It would be good to fix this upstream so it's already flake8 compliant

matentzn: There is a related issue: https://github.com/mapping-commons/sssom-py/issues/121

cthoyt (approved):
LGTM

PR #128, opened 2021-08-31
Keep imports properly sorted with isort

cthoyt:
This PR adds the `isort` utility to the `lint` environment in tox as well as the `flake8-isort` plugin to the flake8 quality assurance testing. Since #127 was merged, this means that this check will be run on all PRs. As you can see by the big diff, there's quite a bit of cleanup it already did!

One thing to consider later (not for accepting this PR) is that there is a mix of absolute and relative imports. While I'm definitely going against the grain here, I think that packages should _only_ be using relative imports, to remind you never to rely on being able to run python files in packages as scripts (instead, the solution is to pip install then use `python -m ...`)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,7 @@
+[tool.black]
+# line-length = 100
+target-version = ['py36', 'py37', 'py38', 'py39']
+
+[tool.isort]
+profile = "black"
+multi_line_output = 3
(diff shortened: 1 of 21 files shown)


matentzn:
Great - so when I run "tox" now it will sort the imports automatically correctly?

matentzn (approved):

cthoyt:
Yes

PR #129, opened 2021-08-31
Add easy flake8 plugins

cthoyt:
This PR adds some easy to comply with flake8 plugins.

- [x] pep8-naming (93c773e)
- [x] flake8-colors (d526d10)
- [x] flake8-bugbear (c1f39e4; or common mistakes that could lead to bugs)
- [x] flake8-bandit (9c373b8; for security stuff)

### DIFF
--- a/sssom/cli.py
+++ b/sssom/cli.py
@@ -564,7 +564,7 @@ def correlations(input: str, output: str, transpose: bool, fields: Tuple):
 
     tups = []
     for i, row in corr.iterrows():
-        for j, v in row.iteritems():
+        for j, v in row.items():
             logging.info(f"{i} x {j} = {v}")
             tups.append((v, i, j))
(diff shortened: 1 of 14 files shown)


sssom/cli.py
```diff
@@ -564,7 +564,7 @@ def correlations(input: str, output: str, transpose: bool, fields: Tuple):
 
     tups = []
     for i, row in corr.iterrows():
-        for j, v in row.iteritems():
+        for j, v in row.items():
```
matentzn, line 567: Is this save? Why that change?

matentzn (approved):
WOW man!! All of these come to your attention from flake8 warnings?? This is insanely awesome!

sssom/cli.py
cthoyt: iteritems() is not necessarily forwards compatible with python3.

PR #132, opened 2021-08-31
Simplify loading data from inside package

cthoyt:


matentzn:
This seems to work great, thanks @cthoyt

PR #140, opened 2021-09-12
include equivalent properties in rewire

cmungall:


sssom/rdf_util.py
```diff
@@ -28,7 +28,7 @@ def expand_curie(curie: str):
 
     for m in mdoc.mapping_set.mappings:
         m: Mapping
-        if m.predicate_id == "owl:equivalentClass":
+        if m.predicate_id == "owl:equivalentClass" or m.predicate_id == "owl:equivalentProperty":
```
cthoyt, line 31:
```suggestion
        if m.predicate_id in {"owl:equivalentClass", "owl:equivalentProperty"}:
```

Perhaps even save this set as a module-level constant

matentzn: I agree with @cthoyt - I have tried to collect such predicate lists up top in utils.py, just so we keep potentially changing stuff closely side-by-side..

cmungall: agreed!

matentzn (approved):

PR #141, opened 2021-09-12
Add better exception when directory does not exist

cthoyt:
This PR does two things:

1. It removes the default none value for `convert_file()`. This didn't make any sense, as all usages of it always had strings
2. Gets the directory from the output file path and checks that it exists. If it doesn't, throw a meaningful exception with a more helpful message.

This should provide a more useful error message than what I encountered in https://github.com/OBOFoundry/COB/issues/176#issuecomment-917671960

### DIFF
--- a/sssom/io.py
+++ b/sssom/io.py
@@ -1,5 +1,6 @@
 import logging
 import os
+import pathlib
 
 import validators
 
@@ -12,7 +13,7 @@ from .writers import get_writer_function, write_table, write_tables
 cwd = os.path.abspath(os.path.dirname(__file__))
 
 
-def convert_file(input_path: str, output_path: str, output_format: str = None):
+def convert_file(input_path: str, output_path: str = None, output_format: str = None):
     """
 
     Args:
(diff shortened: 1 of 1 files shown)


sssom/io.py
```diff
@@ -23,6 +23,8 @@ def convert_file(input_path: str, output_path: str = None, output_format: str =
     Returns:
 
     """
+    if not os.path.exists(os.path.dirname(output_path)):
```
matentzn, line 26: It is not necessary for output path to be anything, because the correct behaviour here should be: if legal_filepath(output_path): write fo file, else: write to stdout.. Not sure whether we did this for convert, but this should be the consistent behaviour for all methods.. So if we want this new and improved condition plus error message, we need to first make sure that output path is actually something other than None or `"-"`.

matentzn (changes requested):

sssom/io.py
cthoyt: Okay, but this is not obvious from any of the documentation or the code itself.

I made a fix in 3c8e046 to check that it's a pathlike before doing this check

matentzn: Would `-` pass as a "path-like object"?

cthoyt: No, but this sort of points out that the entire way this is being handled is broken. Only the CLI should be worrying about parsing that stuff. There's a "smart open" function, which is also a huge anti-pattern. I will send an alternate PR to show how this should be done instead

matentzn: Ok, thank you!

cthoyt: See #145

cthoyt:
This PR is probably superseded by https://github.com/mapping-commons/sssom-py/pull/145

matentzn:
@cthoyt I moved this to draft until you decide its final! :) Thanks

cthoyt:
@matentzn thanks for the reminder. Best to clean this one up, the alternate solution was much nicer :)

PR #142, opened 2021-09-12
Code cleanup

### LINKED ISSUE #130: Use relative (.xyz) and absolute (sssom.xyz) imports consistently
See @cthoyt comment [here](https://github.com/mapping-commons/sssom-py/pull/128#issue-723261390).

cthoyt:
1. Closes #130 by switching to all relative imports
2. Removes some dead code
3. Updates docstrings and type annotations
4. Updates exceptions to be more specific

@hrshdhgd please read the annotations of the changes on this PR

### DIFF
--- a/sssom/cli.py
+++ b/sssom/cli.py
@@ -10,20 +10,15 @@ from pandasql import sqldf
 from rdflib import Graph
 from scipy.stats import chi2_contingency
 
-from sssom.rdf_util import rewire_graph
-from sssom.sparql_util import EndpointConfig, query_mappings
-from sssom.util import (
-    SSSOM_EXPORT_FORMATS,
-    SSSOM_READ_FORMATS,
-    MappingSetDataFrame,
-    to_mapping_set_dataframe,
-)
-from sssom.writers import write_table
-
 from .cliques import split_into_cliques, summarize_cliques
 from .io import convert_file, parse_file, split_file, validate_file
 from .parsers import read_sssom_table
+from .rdf_util import rewire_graph
+from .sparql_util import EndpointConfig, query_mappings
 from .util import (
+    SSSOM_EXPORT_FORMATS,
+    SSSOM_READ_FORMATS,
+    MappingSetDataFrame,
     collapse,
     compare_dataframes,
     dataframe_to_ptable,
@@ -31,7 +26,9 @@ from .util import (
     merge_msdf,
     remove_unmatched,
     smart_open,
+    to_mapping_set_dataframe,
 )
+from .writers import write_table
 
 # Click input options common across commands
 input_argument = click.argument("input", required=True, type=click.Path())
--- a/sssom/cli.py
+++ b/sssom/cli.py
@@ -70,18 +70,7 @@ fields_option = click.option(
 @click.option("-v", "--verbose", count=True)
 @click.option("-q", "--quiet")
 def main(verbose: int, quiet: bool):
-    """Main
-
-    Args:
-
-        verbose (int): Verbose.
-        quiet (bool): Quiet.
-
-    Returns:
-
-        None.
-
-    """
+    """Main."""
     if verbose >= 2:
         logging.basicConfig(level=logging.DEBUG)
     elif verbose == 1:
@@ -113,7 +102,6 @@ def convert(input: str, output: str, output_format: str):
         None.
 
     """
-
     convert_file(input_path=input, output_path=output, output_format=output_format)
 
 
--- a/sssom/cliques.py
+++ b/sssom/cliques.py
@@ -5,11 +5,10 @@ import statistics
 import networkx as nx
 import pandas as pd
 
-from sssom.parsers import to_mapping_set_document
-from sssom.util import MappingSetDataFrame
-
+from .parsers import to_mapping_set_document
 from .sssom_datamodel import MappingSet
 from .sssom_document import MappingSetDocument
+from .util import MappingSetDataFrame
 
 
 def to_networkx(msdf: MappingSetDataFrame) -> nx.DiGraph:
--- a/sssom/cliques.py
+++ b/sssom/cliques.py
@@ -12,10 +12,7 @@ from .util import MappingSetDataFrame
 
 
 def to_networkx(msdf: MappingSetDataFrame) -> nx.DiGraph:
-    """
-    converts a MappingSetDocument to a networkx DiGraph
-    """
-
+    """Convert a MappingSetDocument to a networkx DiGraph."""
     doc = to_mapping_set_document(msdf)
     g = nx.DiGraph()
     # m = {
@@ -102,9 +99,7 @@ def get_src(src, cid):
 
 
 def summarize_cliques(doc: MappingSetDataFrame):
-    """
-    summary stats on a clique doc
-    """
+    """Summarize stats on a clique doc."""
     cliquedocs = split_into_cliques(doc)
     items = []
     for cdoc in cliquedocs:
--- a/sssom/io.py
+++ b/sssom/io.py
@@ -1,12 +1,9 @@
 import logging
-import os
 from typing import Optional
-
-import validators
 
 from .context import get_default_metadata
 from .parsers import get_parsing_function, read_sssom_table, split_dataframe
-from .util import read_metadata
+from .util import raise_for_bad_path, read_metadata
 from .writers import get_writer_function, write_table, write_tables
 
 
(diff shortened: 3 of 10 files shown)


sssom/cli.py
```diff
@@ -73,18 +70,7 @@
 @click.option("-v", "--verbose", count=True)
 @click.option("-q", "--quiet")
 def main(verbose: int, quiet: bool):
-    """Main
```
cthoyt, line 76: Again - it makes no sense to write docs for CLIs like this since click autogenerates them

sssom/cli.py
```diff
@@ -116,7 +102,6 @@ def convert(input: str, output: str, output_format: str):
         None.
 
     """
-
```
cthoyt, line 119: this should have been caught by flake8...

sssom/cliques.py
```diff
@@ -5,18 +5,14 @@
...
-
+from .parsers import to_mapping_set_document
 from .sssom_datamodel import MappingSet
 from .sssom_document import MappingSetDocument
+from .util import MappingSetDataFrame
 
 
 def to_networkx(msdf: MappingSetDataFrame) -> nx.DiGraph:
-    """
-    converts a MappingSetDocument to a networkx DiGraph
```
cthoyt, line 17: Use imperative mood in docstrings and always end with a period

sssom/cliques.py
```diff
@@ -103,9 +99,7 @@ def get_src(src, cid):
 
 
 def summarize_cliques(doc: MappingSetDataFrame):
-    """
-    summary stats on a clique doc
-    """
+    """Summarize stats on a clique doc."""
```
cthoyt, line 102: Prefer one-line docstrings if it all fits rather than breaking up a logical one-line docstring into 3

sssom/io.py
```diff
@@ -1,78 +1,70 @@
...
-
-from sssom.util import read_metadata
+from typing import Optional
 
 from .context import get_default_metadata
 from .parsers import get_parsing_function, read_sssom_table, split_dataframe
+from .util import raise_for_bad_path, read_metadata
 from .writers import get_writer_function, write_table, write_tables
 
-cwd = os.path.abspath(os.path.dirname(__file__))
```
cthoyt, line 12: this appeared in several files but was not used

sssom/io.py
```diff
@@ -1,78 +1,70 @@
...
+        input_path: The path to the input file in one of the legal formats, eg obographs, aligmentapi-xml
+        output_path: The path to the output file.
+        input_format: The string denoting the input format.
+        metadata_path: The path to a file containing the sssom metadata (including curie_map)
+            to be used during parse.
+        curie_map_mode: Defines whether the curie map in the metadata should be extended or replaced with
+            the SSSOM default curie map. Must be one of metadata_only, sssom_default_only, merged
+        clean_prefixes: If True (default), records with unknown prefixes are removed from the SSSOM file.
     """
-    if validators.url(input_path) or os.path.exists(input_path):
```
cthoyt, line 56: any time you have

```python
if cond:
    ...  # big code block
else:
    raise ...
```

Switch to `if not cond`, then you can dedent the entire block. If you turn your head 90 degrees and your code has many mountainous peaks and valleys, there's a problem

sssom/io.py
```diff
@@ -85,42 +77,41 @@ def validate_file(input_path: str):
...
-        raise Exception(f"{input_path} is not a valid file path or url.")
+    raise_for_bad_path(input_path)
+    msdf = read_sssom_table(input_path)
+    splitted = split_dataframe(msdf)
+    write_tables(splitted, output_directory)
 
 
-def get_metadata_and_curie_map(metadata_path, curie_map_mode: str = "metadata_only"):
+def get_metadata_and_curie_map(
+    metadata_path: Optional[str] = None, curie_map_mode: str = "metadata_only"
```
cthoyt, line 95: type annotations give a lot of context to these functions, especially since most are missing docs ATM

sssom/util.py
```diff
@@ -881,18 +883,21 @@ class NoCURIEException(ValueError):
     pass
 
 
-def is_curie(string: str):
-    return re.match(r"[A-Za-z0-9_]+[:][A-Za-z0-9_]", string)
+CURIE_RE = re.compile(r"[A-Za-z0-9_]+[:][A-Za-z0-9_]")
```
cthoyt, line 886: better to pre-compile a regex that is used many times

sssom/writers.py
```diff
@@ -77,13 +76,14 @@ def write_table(msdf: MappingSetDataFrame, filename: str, serialisation="tsv") -
 def write_rdf(
     msdf: MappingSetDataFrame,
     filename: str,
-    serialisation=SSSOM_DEFAULT_RDF_SERIALISATION,
+    serialisation: Optional[str] = None,
```
cthoyt, line 79: rather than leave defaults in the function definiton, use None then assign the default inside. This means that any function calling this one doesn't have to repeat importing/using the default.

sssom/writers.py
```diff
@@ -107,7 +107,7 @@ def write_json(msdf: MappingSetDataFrame, filename: str, serialisation="json") -
             json.dump(data, outfile, indent="  ")
 
     else:
-        raise Exception(
```
cthoyt, line 110: never ever throw Exception in your code - always pick one of its subclasses, otherwise you can't reliably handle it later

sssom/writers.py
```diff
@@ -328,7 +328,7 @@ def to_json(msdf: MappingSetDataFrame) -> JsonObj:
 # Support methods
 
 
-def get_writer_function(output_format, output):
+def get_writer_function(*, output_format: Optional[str] = None, output: str):
```
cthoyt, line 331: using `*` to make all arguments keyword only is a pretty good way to force that you write code that looks more consistent (e.g., the name of the variable that's passed around should probably be the same, and not change from function to function)

sssom/util.py
```diff
@@ -465,72 +465,88 @@ def dataframe_to_ptable(df: pd.DataFrame, priors=None, inverse_factor: float = 0
...
-    )
+    else:
+        logging.warning(
+            f"Priors given ({priors}), but not being used by dataframe_to_ptable() method."
+        )
 
     df = collapse(df)
     rows = []
     for _, row in df.iterrows():
-        s = row[SUBJECT_ID]
```
cthoyt, line 479: should go without saying, one letter variable names are bad because they are not conveying any information to the reader

sssom/parsers.py
```diff
@@ -247,6 +234,7 @@ def from_sssom_rdf(
     curie_map = _check_curie_map(curie_map)
 
     if mapping_predicates is None:
+        # FIXME unused
```
hrshdhgd, line 237: @matentzn , thoughts? Would this be unused all the time?

hrshdhgd (commented):
Everything looks good. I did learn a few new things for which thank you sincerely @cthoyt !. Just one comment where I need @matentzn 's opinion and I think this branch is ready to be merged. All tests pass!

sssom/parsers.py
cthoyt: This doesn't necessarily have to be addressed in this PR - I only added the FIXME tag for us to check in on later

sssom/cli.py
matentzn: I am happy for all docstrings to be removed from cli.py

sssom/cli.py
matentzn: Anything we can do something about, or just a comment to self?

sssom/io.py
matentzn: Hard to remember this great advice, can this be checked by the code readbility checkers?

sssom/io.py
matentzn: will add extensive docs everywhere.

sssom/sparql_util.py
```diff
@@ -4,12 +4,10 @@
 
 import pandas as pd
 from rdflib import URIRef
-
-# noinspection PyProtectedMember
```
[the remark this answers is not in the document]
matentzn: oh. They are to tell my pycharm to shut up. not wedded to them.

sssom/writers.py
matentzn: I have never seen this, can you link some docs that explain this?

matentzn (approved):
This PR is truly awesome amazing. I feel we need to add a checklist though to write up the most common mistakes not caught by QC.. Do you have such a list somewhere @cthoyt 

THANK YOU

sssom/cli.py
cthoyt: It might have been from a plug-in i didn’t  add yet

sssom/io.py
cthoyt: It could be added as a flake8 plugin in theory, but I don’t know enough on how to make one myself

sssom/io.py
cthoyt: Definitely on the todo.

sssom/writers.py
cthoyt: https://www.python.org/dev/peps/pep-3102/ gave the spec, this is a short good blog post explaining it for humans https://lukeplant.me.uk/blog/posts/keyword-only-arguments-in-python/

cthoyt:
I don't have a list of these things, but I guess I should start compiling them. Here looks like a good place: https://cthoyt.com/2021/09/13/beyond-linters.html

sssom/writers.py
matentzn: Wow! Never heard of this. Thanks

Issue #138, opened 2021-09-03
Migrate to rdflib 6


matentzn:

cmungall:
I believe you can make the code work with either. One annoying
incompatibility is in rdflib6 you drop the decode() after saving a graph.
@hsolbrig is trying to get a shim layer into a new rdflib6 release

matentzn:
your new code (rdf_utils) is incompatible with 6 :D that's how I noticed it in the first place.

PR #145, opened 2021-09-14
Improve file handling

cthoyt:
The lesson learned for this PR is that you should split the functionality of opening files from the functionality that reads/writes them. Now, the CLI takes care of opening files for read and write, and the functions in the package are only responsible for interacting with files (and not paths).

This is better because click already gives you tools for making nice unix-like CLIs (e.g., give dash and it sends to standard out). It's also better because it makes the code generic to people who might have file-like objects or other sneaky files that are gzipped.

I'm not sure if RDFlib is supporting this out of the box, so I hope there are tests for the code that I touched.

### DIFF
--- a/sssom/cli.py
+++ b/sssom/cli.py
@@ -26,7 +26,6 @@ from .util import (
     filter_redundant_rows,
     merge_msdf,
     remove_unmatched,
-    smart_open,
     to_mapping_set_dataframe,
 )
 from .writers import write_table
@@ -40,9 +39,6 @@ input_format_option = click.option(
     help=f'The string denoting the input format, e.g. {",".join(SSSOM_READ_FORMATS)}',
 )
 output_option = click.option(
-    "-o", "--output", help="Output file, e.g. a SSSOM tsv file."
-)
-improved_output_option = click.option(
     "-o",
     "--output",
     help="Output file, e.g. a SSSOM tsv file.",
@@ -91,7 +87,7 @@ def main(verbose: int, quiet: bool):
 
 @main.command()
 @input_argument
-@improved_output_option
+@output_option
 @output_format_option
 def convert(input: str, output: TextIO, output_format: str):
     """Convert file (currently only supports conversion to RDF)
@@ -127,7 +123,7 @@ def convert(input: str, output: TextIO, output_format: str):
     required=True,
     help="If True (default), records with unknown prefixes are removed from the SSSOM file.",
 )
-@improved_output_option
+@output_option
 def parse(
     input: str,
     input_format: str,
@@ -202,7 +198,7 @@ def split(input: str, output_directory: str):
 @input_argument
 @output_option
 @click.option("-W", "--inverse-factor", help="Inverse factor.")
-def ptable(input=None, output=None, inverse_factor=None):
+def ptable(input, output: TextIO, inverse_factor):
     """Write ptable (kboom/boomer input) should maybe move to boomer (but for now it can live here, so cjm can tweak
 
     Args:
@@ -224,15 +220,13 @@ def ptable(input=None, output=None, inverse_factor=None):
     df = collapse(msdf.df)
     # , priors=list(priors)
     rows = dataframe_to_ptable(df)
-
-    with smart_open(output) as fh:
-        for row in rows:
-            print("\t".join(row), file=fh)
+    for row in rows:
+        print(row, sep="\t", file=output)
 
 
 @main.command()
 @input_argument
-@improved_output_option
+@output_option
 def dedupe(input: str, output: TextIO):
     """Remove lower confidence duplicate lines.
 
@@ -259,7 +253,7 @@ def dedupe(input: str, output: TextIO):
 @click.option("-q", "--query", help='SQL query. Use "df" as table name.')
 @click.argument("inputs", nargs=-1)
 @output_option
-def dosql(query: str, inputs: List[str], output: str):
+def dosql(query: str, inputs: List[str], output: TextIO):
     """
     Run a SQL query over one or more sssom files.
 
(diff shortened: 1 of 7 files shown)


sssom/cli.py
```diff
@@ -41,6 +42,13 @@
 output_option = click.option(
     "-o", "--output", help="Output file, e.g. a SSSOM tsv file."
 )
+improved_output_option = click.option(
+    "-o",
+    "--output",
+    help="Output file, e.g. a SSSOM tsv file.",
+    type=click.File(mode="w"),
```
cthoyt, line 49: note that you can have click do all of the hard work for opening files for you. below, you can even set stdout as the default. Then, later, you never have to write code that worries about file paths, handling, parsing, etc.

sssom/cli.py
```diff
@@ -315,15 +312,15 @@ def dosql(query: str, inputs: List[str], output: str):
...
 @click.option("-l", "--limit", type=int)
 @click.option("-P", "--prefix", type=click.Tuple([str, str]), multiple=True)
-@output_option
+@improved_output_option
 def sparql(
-    url: str = None,
-    config=None,
-    graph: str = None,
-    limit: int = None,
-    object_labels: bool = None,
```
cthoyt, line 324: Emphatically do not add default values in `click` command line functions - click will always provide all of them

cthoyt: click options get none by default anyway via the framework

sssom/cli.py
```diff
@@ -425,7 +422,8 @@ def partition(inputs: List[str], output_directory: str):
         # logging.info(f'Example: {cdoc.mapping_set.mappings[0].subject_id}')
         # logging.info(f'Writing to {ofn}. Size={len(cdoc)}')
         msdf = to_mapping_set_dataframe(cdoc)
-        write_table(msdf, ofn)
+        with open(ofn, "w") as file:
+            write_table(msdf, file)
```
cthoyt, line 426: as good 'ol mando says, this is the way.

Better for the function not to have to think about what kind of input this is - if you _must_, you can always write a wrapper around a function that interacts with files to make it interpret a path string in a certain way

sssom/cli.py
```diff
@@ -494,18 +492,15 @@ def crosstab(input: str, output: str, transpose: bool, fields: Tuple):
     ct = pd.crosstab(df[f1], df[f2])
     if transpose:
         ct = ct.transpose()
-    if output is not None:
-        ct.to_csv(output, sep="\t")
-    else:
-        logging.info(ct)
+    ct.to_csv(output, sep="\t")
```
cthoyt, line 495: to_csv accepts an IO like sys.stdout so this can all be combined

sssom/cli.py
```diff
@@ -627,11 +619,7 @@ def rewire(
     g.parse(input, format=input_format)
     rewire_graph(g, msdf, precedence=precedence)
     rdfstr = g.serialize(format=output_format).decode()
-    if output:
-        with open(output, "w") as stream:
-            stream.write(rdfstr)
-    else:
-        print(rdfstr)
+    print(rdfstr, file=output)
```
cthoyt, line 622: print lets you say where to print via the `file` argument, use this to deduplicate code like this a ton

sssom/writers.py
```diff
@@ -59,23 +59,13 @@ def write_table(msdf: MappingSetDataFrame, filename: str, serialisation="tsv") -
     lines = [f"# {line}" for line in lines if line != ""]
     s = msdf.df.to_csv(sep=sep, index=False)
     lines = lines + [s]
-
-    if filename and filename != "-":
```
cthoyt, line 63: ewwwwww all of this is no longer necessary because click does it for you!

sssom/writers.py
```diff
@@ -125,7 +115,8 @@ def write_owl(
         serialisation = SSSOM_DEFAULT_RDF_SERIALISATION
 
     graph = to_owl_graph(msdf)
-    graph.serialize(destination=filename, format=serialisation)
+    t = graph.serialize(format=serialisation, encoding="utf-8")
```
cthoyt, line 118: this is annoying - rdflib only writes to bytes (though I heard they got rid of this in rdflib 6 #138!!!)

sssom/writers.py
```diff
@@ -347,11 +339,11 @@ def get_writer_function(*, output_format: Optional[str] = None, output: str):
 
 
 def write_tables(sssom_dict, output_dir) -> None:
-    for split_id in sssom_dict:
-        sssom_file = os.path.join(output_dir, f"{split_id}.sssom.tsv")
-        msdf = sssom_dict[split_id]
```
cthoyt, line 352: any time you do this in a loop, you can use the `.items()` iterator over the dict instead

cthoyt:
@hrshdhgd same drill as last time, please read through this PR and check out the comments I made about the general philosophy of the changes

sssom/cli.py
```diff
@@ -227,16 +220,14 @@ def ptable(input=None, output=None, inverse_factor=None):
     df = collapse(msdf.df)
     # , priors=list(priors)
     rows = dataframe_to_ptable(df)
-
-    with smart_open(output) as fh:
-        for row in rows:
-            print("\t".join(row), file=fh)
+    for row in rows:
+        print(row, sep="\t", file=output)
```
cthoyt, line 224: note that there's a `sep` option which makes this much more simple

sssom/cli.py
hrshdhgd: Excellent! mypy was complaining about this too! I've never initialized variables to `None` like this ever but again, this is code I inherited so no idea where this style came from.

hrshdhgd (approved):
Looks good and as usual learnt a few things. Thanks again!

PR #143, opened 2021-09-13
Add static type checking with MyPy

cthoyt:
Blocked by #146 

This PR does the following:

- [x] Adds `mypy` environment to `tox.ini`
- [x] Adds relevant configuration to `pyproject.toml` to exclude auto-generated modules
- [x] Adds `mypy` testing to CI configuration
- [x] Fixes all type annotations
- [x]  Fix remaining tests

PR #146, opened 2021-09-15
Remove deprecated MetaTSVConverter

cthoyt:
This code is unused and is marked as deprecated. Rather than trying to fix it in #143, it might just be better to delete it now.


cthoyt:
@matentzn I think the failing tests are due to something unrelated to this PR with RDFlib's JSON-LD plugin

matentzn:
I assigned @cmungall  to this to take a look, but I have no objections myself. If we don't hear from him in 2 days, we can merge.

matentzn (approved):

matentzn:
Is it possible we need: https://github.com/mapping-commons/sssom-py/pull/148

cthoyt:
@matentzn please mark as approved so we can merge

matentzn (approved):

PR #148, opened 2021-09-20
Add rdflib-jsonld module?

matentzn:


cthoyt:
Seems like this didn't do the trick

cthoyt:
Weird, the use of https://pypi.org/project/rdflib-jsonld/0.5.0/ didn't work. Maybe it has to be marked as <= instead of ==. I'm testing locally now

cthoyt:
It might be okay to not restrict rdflib-jsonld, it seems to work with rdflib==5.0.0 when it installs rdflib-jsonld==0.6.2 (contrary to what their readme says, which seems to be specifically about the old version python3.6)

matentzn:
Super odd.. Locally the tests all work for me, using the exact versions indicated here.

cthoyt:
I'm experiencing mostly the same thing, except `rdflib-pyld-compat` isn't necessary to get it to work

matentzn:
could it be that setup.cfg somehow works offs a different version of pip, one that doesn't know old versions?

cthoyt:
Setuptools >= 58 (released september 5th/three weeks ago) breaks usage of 2to3, so I think we have to figure out how to downgrade setuptools

Ref: https://exerror.com/error-in-mongoengine-setup-command-use_2to3-is-invalid/

setup.cfg
```diff
@@ -64,7 +64,7 @@ install_requires =
     recommonmark>=0.7
     scikit_learn
     scipy
-    setuptools
+    setuptools<=57.5.0
```
cthoyt, line 67: oh weird, typically setuptools isn't actually a requirement for using the code. I'm surprised this is listed here

matentzn:
The primary problem is that some of our core dependencies use rdflib 5. This is what I believe causes the whole problem now - rdflib-jsonld has merged into rdflib 6 and is now obsolete.

cthoyt:
Oh, yes. This is the case. rdflib-jsonld has been "tombstoned" as of version 0.6.0 which means you can download it but now it doesn't do anything. Should we just put more effort into upgrading to rdflib 6 instead?

matentzn:
@hrshdhgd can you look into this? 

The last thing we want to do, but we will if we must, is go try migrating linkml to rdflib 6...

I pinged them a bit on that front: https://github.com/linkml/linkml/issues/291

 will also talk to @cmungall later today about this issue.

matentzn:
I also left a comment at rdflib, just in case: https://github.com/RDFLib/rdflib/issues/1417

hrshdhgd:
```
ERROR: Could not find a version that satisfies the requirement rdflib-jsonld==0.5.0 (from sssom[test]) (from versions: 0.2, 0.3, 0.4.0, 0.5.0, 0.6.0, 0.6.1, 0.6.2)
59
ERROR: No matching distribution found for rdflib-jsonld==0.5.0
```

So weird .... the version is listed right there

cthoyt:
@hrshdhgd scoot up a bit - the issue is that it can't install rdflib-jsonld==0.5.0 properly, not that it can't find it. it's the 2to3 issue

hrshdhgd:
0.6.0 did the trick!

setup.cfg
```diff
@@ -57,11 +57,12 @@ install_requires =
...
     pandasql
     pyyaml
-    rdflib
+    rdflib==5.0.0
+    rdflib-jsonld==0.6.0
     recommonmark>=0.7
     scikit_learn
     scipy
-    setuptools
+    setuptools<58.0
```
matentzn, line 65: is this still necessary?

tox.ini
```diff
@@ -8,6 +8,8 @@ envlist =
     lint
     flake8
     py
+requires =
+   setuptools<58.0
```
matentzn, line 12: make sure this is still necessary/

tox.ini
```diff
@@ -16,6 +18,7 @@ extras =
     test
 description = Run unit tests with pytest. This is a special environment that does not get a name, and
               can be referenced with "py".
+skipsdist = False
```
matentzn, line 21: Is this necessary, who put it there?

matentzn:
Thank you both :) Awesome. Made a few comments!

tox.ini
hrshdhgd: I put it there. Was testing it..not required. Taking it off.

tox.ini
hrshdhgd: nope, taking it out!

setup.cfg
hrshdhgd: Yes.

matentzn: Why? Just so we have it documented :) thank you!

cmungall:
Is this tombstoning standard practice?
https://github.com/RDFLib/rdflib-jsonld/releases/tag/0.6.2

So they scooped out all functionality and made a new.. minor release? To force people to switch?

matentzn:
@hrshdhgd why did you remove the version on rdf lib jsonld? This should fail. The question is why we need the version on setuptools

hrshdhgd:
> @hrshdhgd why did you remove the version on rdf lib jsonld? This should fail. The question is why we need the version on setuptools

Just to document the repercussions:

On removing version for redflib-jsonld, we get the error:
```
No module named 'rdflib_jsonld.parser'
```

On removing the versioning pin for setuptools, it all works.

setup.cfg
hrshdhgd: My bad, it is not needed. Took it off.

matentzn:
Perfect thanks :)

matentzn:
You can now merge this into the other open PRs to get them to pass again..

matentzn:
Nice one man.

cmungall:
also excluded tombstoned on linkml: https://github.com/linkml/linkml/pull/352

PR #151, opened 2021-09-20
remove unused linkml dependencies

cmungall:
you only need runtime


cthoyt:
You mean, “all you need is runtime” right?

cthoyt (approved):

matentzn:
This was not always true, but I guess its now :) Thank you @cmungall

PR #159, opened 2021-09-27
Add sphinx-autodoc-typehints

cthoyt:
This automatically turns type hints into documentation so you don't have to include those verbose `:type ...: ....` lines

It's already included in the `docs` extra, but it's super non-obvious how the docs actually get built in this repo, so I haven't been able to test that they can be re-built properly

Is there a real advantage to building and maintaining the docs inside this repo? Wouldn't using RTD be better?


cthoyt:
Having an error-free documentation build currently trips on

` /home/runner/work/sssom-py/sssom-py/.tox/docs/lib/python3.7/site-packages/sssom/sssom_datamodel.py:docstring of jsonasobj2._jsonobj.JsonObj.__new__:6:Field list ends without a blank line; unexpected unindent.`

This is caused by one of the auto-generated files, even worse, it's to one of the dependencies of the auto-generated file. I wonder if adding an `__all__` to make sure that only relevant stuff is exported could be a good idea (though this would involve an upstream patch to [`pythongen.py`](https://github.com/linkml/linkml/blob/main/linkml/generators/pythongen.py) in the linkml generator

cthoyt:
I dug deeper and found that everything in `sssom.sssom_datamodels` that inherits from `EnumDefinitionImpl` which includes `PredicateModifierEnum`, etc. all have a very deep inheritance hierarchy that eventually ends up at `JsonObj`, and inherits the docstrings for `__new__` which are problematic in the upstream library. I tried to turn off documentation of the `__new__` methods but haven't figured it out yet

matentzn:
@cthoyt thank you, I converted this to Draft for the time being.

matentzn:
@hrshdhgd please harvest what you think is useful from this PR, then close without merging.

matentzn:
What about the changes to sphinx/conf.py and sphinx/tests.rst?

matentzn:
Obsolete by https://github.com/mapping-commons/sssom-py/pull/249

Issue #124, opened 2021-08-30
Make sure multivalued elements in the model are parsed correctly


matentzn:
This needs some treatment on both the parsing and converting side.

Basically, it goes like this:
1. All `multivalued` fields in the sssom spec (https://github.com/mapping-commons/sssom-py/blob/master/schema/sssom.yaml) can be represented as `|`-separated values in a table cell. (this should be in the dataclasses, _not_ determined by parsing the spec!)
2. When they are parsed into the metamodel (to_mapping_set_document, parsers.py), we need to ensure that they are parsed as multi-valued.

matentzn:
My guess: 

| ... | creator_id |
| -- | -- |
| .. | orcid:001\|orcid:002 |

When parsing the value, we neet to do something like value.spit("|") and pass a list (ideally whitespace `strip`-ped) to the linkml object.

hrshdhgd:
Fixed by #163

PR #163, opened 2021-09-29
Make sure multivalued elements in the model are parsed correctly

hrshdhgd:
Issue #124


matentzn:
Can we add a test for this? I would like to be convinced that this change actually does what we think it does :D

hrshdhgd:
Done!

sssom/parsers.py
```diff
@@ -618,6 +618,14 @@ def to_mapping_set_document(msdf: MappingSetDataFrame) -> MappingSetDocument:
                 ok = False
                 if k:
                     k = str(k)
+                if (
+                    (hasattr(Mapping, k) or hasattr(MappingSet, k))
+                    and v is not None
+                    and "|" in str(v)
```
cthoyt, line 624: is coercing into a string so smart? what type is `v` supposed to be? maybe it would be better to add a check for whether it's multivalued into that class (and also the function for parsing)

matentzn: I dont know what your comment exactly refers to, but if it would be better, we could add another intsanceof str check to be sure.. What class should the multivalued check be in?

cthoyt: Idk - it's not obvious what a `v` is by reading this code. Obviously a better variable name is required as a first step. Second, if it's ever possible a `v` isn't a string already, I would be incredibly worried to just do a string coercion, since every object can do this and it will get totally meaningless results in most cases. Better to make a helper function that is a "is_multivalued()` that can do instance checks

matentzn: Ok, I will work this out with @hrshdhgd later today.

Makefile
```diff
@@ -47,6 +47,8 @@ sssom/internal_context.py: schema/sssom.context.jsonld
 	echo "sssom_context = \"\"\""  >> $@
 	cat $< >> $@
 	echo "\"\"\""  >> $@
+	echo "multivalued_slots = [\"match_type\", \"creator_id\", \"creator_label\", \"author_id\", \"author_label\", \"reviewer_id\", \"reviewer_label\", \"license\", \"mapping_set_source\", \"subject_match_field\", \"object_match_field\", \"match_string\", \"subject_preprocessing\", \"object_preprocessing\", \"see_also\"]"  >> $@
+
```
cthoyt, line 51: can you make it run black on this file immediately after it's been written?

hrshdhgd: Does black run over a `Makefile`? I ran black and it only formatted `*.py` files

sssom/util.py
```diff
@@ -921,3 +922,19 @@ def raise_for_bad_path(file_path: str) -> None:
...
+    """Check whether the slot is multivalued according to the SSSOM specification.
+
+    :param slot: Slot name
+    :return: Slot is multivalued or no
+    """
+    # Ideally:
+    # view = SchemaView('schema/sssom.yaml')
+    # return view.get_slot(slot).multivalued
+
+    if slot not in multivalued_slots:
```
cthoyt, line 937: for now you could just `return slot in multivalued_slots`

sssom/parsers.py
```diff
@@ -86,7 +87,9 @@ def read_sssom_rdf(
 
 
 def read_sssom_json(
-    file_path: str, prefix_map: Dict[str, str] = None, meta: Dict[str, str] = None
```
cthoyt, line 89: did black cause these spurious documentation changes?

sphinx/conf.py
```diff
@@ -12,17 +12,18 @@
 #
 import os
 import sys
-sys.path.insert(0, os.path.abspath('../'))
```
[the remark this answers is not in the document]
hrshdhgd: This is a result of black+flake8  formatting

sssom/parsers.py
cthoyt: Weird

hrshdhgd:
The one error I get here is form the package [rdflib_shim](https://github.com/hsolbrig/rdflib-shim) which is apparently used by both `linkml` and `linkml-runtime`. The error is:

```
    def serialize_shim(*args, **kwargs) -> DecodableStr:
>       return DecodableStr(orig_serialize(*args, **kwargs).decode())
E       AttributeError: 'NoneType' object has no attribute 'decode'

.tox/py/lib/python3.9/site-packages/rdflib_shim/__init__.py:30: AttributeError
```

it seems like the library `rdflib_shim` had it's latest update 3 days ago.

hrshdhgd:
Pinned linkml-runtime to 1.1.10 for the time-being.

matentzn:
I am not diametrically opposed to pinning versions, but we should have at least a new issue "Unpin LinkML version" with a link to the respective LinkML issue?

matentzn:
Looks good:

```
{
      "subject_id": "x:bone_element",
      "predicate_id": "owl:equivalentClass",
      "object_id": "y:bone",
      "match_type": [
        "Lexical"
      ],
      "subject_label": "bone element",
      "subject_category": "biolink:AnatomicalEntity",
      "object_label": "bones",
      "object_category": "biolink:AnatomicalEntity",
      "subject_source": "x",
      "object_source": "y",
      "mapping_tool": "rdf_matcher",
      "confidence": 0.738796125,
      "comment": "."
    },
```

matentzn (approved):

hrshdhgd:
> I am not diametrically opposed to pinning versions, but we should have at least a new issue "Unpin LinkML version" with a link to the respective LinkML issue?

[Done](https://github.com/mapping-commons/sssom-py/issues/188#issue-1085927208)

cmungall:
I commented in the issue. Was not able to reproduce issue

It's important that if we think we encounter an error there is a reproducible way to test it - I recommend making a PR that fails demonstrating the issue in future

hrshdhgd:
Unpinning the `linkml_runtime` version will reproduce the issue. I can start a new PR

PR #164, opened 2021-10-04
Flake8 Round 3: You didn't think we could do it again?

cthoyt:
This PR finishes the last 3 flake8 ignores. If it's not looking so complicated, it might add some of the other plugins

81mPquqrVLL _AC_SL1000_-1


matentzn:
You are _crazy_.. :D Thank you :)

cthoyt:
@matentzn not sure why CI didn't run on my last commit. Locally, all commands in `tox` are passing, but I'd ideally like to see it on CI as well before merging.

PR #167, opened 2021-10-08
Add py310 testing to CI

### LINKED ISSUE #160: Support for py36?
It's currently unclear whether py36 is really supported or not. The setup configuration says so, but the testing does not cover it.

Py36 dies in 3 months, so maybe it's best we just support py37+. See: https://endoflife.date/python

cthoyt:
Closes #160

Testing makes it clear that the dependencies of this package do not run on py36, so it's kicked out.

Py310 support blocked by:
- https://github.com/linkml/linkml/pull/401
- https://github.com/linkml/linkml-model/issues/49
- and probably more, since linkml is split among so many repos and they're _really_ hard to understand


matentzn:
@cthoyt we are getting ready to merge the remaining PRs.. do you want to try and finalise this one?

cthoyt:
I believe this PR is blocked by upstream repositories not yet having Python 3.10 support. CI is running now and will report if that's still the cas

.github/workflows/qc.yml
```diff
@@ -13,7 +13,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.7, 3.8, 3.9]
+        python-version: [3.7, 3.8, 3.9, "3.10"]
```
matentzn, line 16: Why is 3.10 in quotes?

matentzn:
Thank you @cthoyt!

.github/workflows/qc.yml
cthoyt: 3.10 not in quotes is the same as 3.1

matentzn: Very interesting. Thanks

.github/workflows/qc.yml
```diff
@@ -13,7 +13,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.7, 3.8, 3.9]
+        python-version: ["3.7", "3.10"]
```
matentzn, line 16: Just for me to understand, why did you remove 3.8, 3.9?

cthoyt: If it works on 3.7 and 3.10 it's pretty good chance it works in the middle. I also needed to make a change to get it to run again 😆 . We're still not ready for 3.10 apparently

hrshdhgd:
Solved by #261

PR #171, opened 2021-10-08
Add windows testing and update path construction

cthoyt:
A lot of file paths inside the tests were being constructed with string concatenation. This PR introduces `pathlib` (i.e., the high level interface) in many places and switches over to `os` (i.e., the low-level interface) for the rest. This also makes it possible to pass tests on Windows, so I added that to the GitHub actions as well.


matentzn:
@cthoyt was this ready? we are getting ready to merge the remaining PRs next week to make a new release..

hrshdhgd (dismissed):
Looks good to me.

hrshdhgd (approved):
Resolved conflicts and ran flake8 + linter to pass GH Actions.

matentzn (approved):
I have never seen this kind of syntax (using a `/` straight in the path rather than os.path.join()), but I will trust your judgement @cthoyt.

Thank you for this!

PR #172, opened 2021-10-08
Updating to new SSSOM model

matentzn:
Will probably break quite a few things.

@hrshdhgd I have updated SSSOM here, not tested anything, not looked at anything.. Can you fix it before #163 and then we can test that properly :P

Thanks.


matentzn:
Yeah this will be seriously hard.. Its a very much newer linkml version that generated the python file.. It does not look too bad:

```
sssom/util.py:31: error: Module "sssom.sssom_datamodel" has no attribute "Entity"
sssom/util.py:31: error: Module "sssom.sssom_datamodel" has no attribute "slots"
sssom/__init__.py:1: error: Module "sssom.sssom_datamodel" has no attribute "slots"
sssom/writers.py:18: error: Module "sssom.sssom_datamodel" has no attribute "slots"
sssom/rdf_util.py:9: error: Module "sssom.sssom_datamodel" has no attribute "EntityId"
```

But its actually quite touch because the concept of "Entity" was removed from the metadamodel.

If this is too hard, let me know, I will do it then when I get to it @hrshdhgd :)

matentzn:
I have updated sssom.yaml in https://github.com/mapping-commons/sssom

so you should pull changes to the repo and ditch any changes you have locally in that repo. This also adds the predicate_modifier column.

hrshdhgd:
The mappings type being dict has been fixed where now it is a list. See [this on linkml ](https://github.com/linkml/linkml/issues/477)and [this in our repo](https://github.com/mapping-commons/sssom-py/issues/177).

sssom/rdf_util.py
```diff
@@ -39,7 +40,7 @@ def expand_curie(curie: str) -> URIRef:
                 src, tgt = m.subject_id, m.object_id
             else:
                 src, tgt = m.object_id, m.subject_id
-            if not isinstance(src, EntityId) or not isinstance(tgt, EntityId):
+            if not isinstance(src, URIorCURIE) or not isinstance(tgt, URIorCURIE):
```
matentzn, line 43: Can we double check that this is correct? I thought we got rid of the URIorCURIE construct at some point. Is there a test that at least touches this method?

sssom/parsers.py
```diff
@@ -188,24 +197,37 @@ def from_sssom_dataframe(
...
-            #    v = Entity(id=v)
-            if hasattr(Mapping, k):
-                mdict[k] = v
-                ok = True
-            if hasattr(MappingSet, k):
-                ms[k] = v
-                ok = True
-            if not ok:
-                bad_attrs[k] += 1
+            if v and v == v:
```
matentzn, line 208: I have never see this expression. Isnt v==v redundant withjust  v?

matentzn: Or is this to protect agains NaNs?

sssom/parsers.py
```diff
@@ -260,11 +284,19 @@ def from_sssom_rdf(
...
                     if k:
+                        if (
+                            is_multivalued_slot(k)
+                            and v is not None
+                            and isinstance(v, str)
+                            and "|" in v
+                        ):
+                            # IF k is multivalued, then v = List[values]
+                            v = [s.strip() for s in v.split("|")]
                         mdict[k] = v
```
matentzn, line 300: Maybe refacror 292-299 and reuse in 214?

sssom/parsers.py
```diff
@@ -307,6 +339,12 @@ def from_sssom_json(
     mapping_set = cast(
         MappingSet, JSONLoader().load(source=jsondoc, target_class=MappingSet)
     )
+    # ** Will need to remove this line
+    # ** once linkml-runtime fixes JSONLoader.load()
```
matentzn, line 343: Add the linkml issue that relates to this.

matentzn: And a # TODO tag

sssom/parsers.py
```diff
@@ -382,8 +423,10 @@ def from_obographs(
     :return: An SSSOM data frame (MappingSetDataFrame)
     """
     _ensure_prefix_map(prefix_map)
-
-    ms = MappingSet()
+    if meta is not None:
+        ms = MappingSet(mapping_set_id=meta["mapping_set_id"], license=meta["license"])
+    else:
+        ms = MappingSet(mapping_set_id=DEFAULT_MAPPING_SET_ID, license=DEFAULT_LICENSE)
     mlist: List[Mapping] = []
```
matentzn, line 430: Refactor 426-429 (see L200)

sssom/parsers.py
```diff
@@ -596,6 +639,8 @@ def _cell_element_values(cell_node, prefix_map: PrefixMap) -> Optional[Mapping]:
             except NoCURIEException as e:
                 logging.warning(e)
 
+    mdict["match_type"] = "Unspecified"
```
matentzn, line 642: Put "Unspecified" into constants.py. Generally, we should avoid using raw strings as keys in dictionaries, better use CONSTANTS like MATCH_TYPE_UNSPECIFIED. No need to change the whole code now :P

sssom/parsers.py
```diff
@@ -612,26 +657,43 @@ def to_mapping_set_document(msdf: MappingSetDataFrame) -> MappingSetDocument:
...
     mlist: List[Mapping] = []
-    ms = MappingSet()
+    if msdf.metadata is not None:
+        ms = MappingSet(
+            mapping_set_id=msdf.metadata["mapping_set_id"],
+            license=msdf.metadata["license"],
+        )
+    else:
+        ms = MappingSet(mapping_set_id=DEFAULT_MAPPING_SET_ID, license=DEFAULT_LICENSE)
     bad_attrs = {}
```
matentzn, line 667: refactor along with L 200 and and L426

sssom/parsers.py
```diff
@@ -612,26 +657,43 @@ def to_mapping_set_document(msdf: MappingSetDataFrame) -> MappingSetDocument:
...
-                if hasattr(MappingSet, k):
-                    ms[k] = v
-                    ok = True
-                if not ok:
-                    if k not in bad_attrs:
-                        bad_attrs[k] = 1
-                    else:
-                        bad_attrs[k] += 1
+                # Check if v exists AND v is not nan
+                if v and v == v:
```
matentzn, line 673: Can we somehow refactor these lines with L200 ff? They seem to be doing mostly the same..

matentzn: L673-700, or even more.

matentzn (commented):
@hrshdhgd looks great. I made a few refactoring suggestions.. After that, this can be merged.

sssom/cliques.py
```diff
@@ -82,6 +83,18 @@ def split_into_cliques(msdf: MappingSetDataFrame) -> List[MappingSetDocument]:
...
     ]
 
+    # ** THIS BLOCK IS A BAND-AID SOLUTION **********************************************
+    #   DEPENDENT ON A LINKML BUG FIX REPORTED AT
+    #   https://github.com/linkml/linkml/issues/477
+
+    for item in documents:
+        if item.mapping_set.mappings == {}:
+            item.mapping_set.mappings = []
+        logging.warning(
```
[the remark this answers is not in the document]
hrshdhgd: This is an outdated version of the file. This was the band-aid solution for the dict => list. This code is deleted.

sssom/parsers.py
hrshdhgd: protect against NaNs.

sssom/parsers.py
hrshdhgd: good catch! It has been resolved by `inlined_as_list: true` for mappings . This is not needed anymore. Deleting.

sssom/parsers.py
hrshdhgd: Done

sssom/parsers.py
hrshdhgd: Done

sssom/parsers.py
hrshdhgd: Done

sssom/parsers.py
hrshdhgd: Done

sssom/rdf_util.py
hrshdhgd: `test_rewire.py` hits this code and all values of `tat` and `src` are of `<class 'sssom.sssom_datamodel.EntityReference'>`. I take out this check.

sssom/parsers.py
hrshdhgd: Done

hrshdhgd:
I have addressed all of your refactored suggestions. Thank you Nico!

schema/sssom.external.context.jsonld
```diff
@@ -14,6 +14,7 @@
         "AGRO": "http://purl.obolibrary.org/obo/AGRO_",
         "AISM": "http://purl.obolibrary.org/obo/AISM_",
         "AMPHX": "http://purl.obolibrary.org/obo/AMPHX_",
+        "APO": "http://purl.obolibrary.org/obo/YPO_",
```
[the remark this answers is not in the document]
matentzn: Related: https://github.com/OBOFoundry/OBOFoundry.github.io/issues/1676

PR #182, opened 2021-12-06
Path for sssom.yaml discoverable by other functions through constants.py

hrshdhgd:
This fixes the issue where the `sssom.yaml` located at a relative location is discovered by functions.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,7 +23,7 @@ schema/sssom.external.context.jsonld:
 	wget $(DEFAULT_PREFIX_MAP) -O $@
 schema/%.context.jsonld: .FORCE
 	wget $(SSSOM_JSONLD_CONTEXT) -O $@
-schema/%.yaml: .FORCE
+schema/%.yaml sssom/%.yaml: .FORCE
 	wget $(SSSOM_YAML) -O $@
 
 
@@ -52,7 +52,6 @@ deploy-dm: sssom/external_context.py sssom/internal_context.py
 	cp schema/sssom_datamodel.py sssom/
 	cp schema/sssom.context.jsonld sssom/
 	cp schema/sssom.external.context.jsonld sssom/
-	cp schema/sssom.yaml sssom/
 	
 
 install:
--- a/Makefile
+++ b/Makefile
@@ -26,6 +26,7 @@ schema/%.context.jsonld: .FORCE
 schema/%.yaml: .FORCE
 	wget $(SSSOM_YAML) -O $@
 
+
 test:
 	tox
 	sh tests/tests.sh
@@ -51,6 +52,7 @@ deploy-dm: sssom/external_context.py sssom/internal_context.py
 	cp schema/sssom_datamodel.py sssom/
 	cp schema/sssom.context.jsonld sssom/
 	cp schema/sssom.external.context.jsonld sssom/
+	cp schema/sssom.yaml sssom/
 	
 
 install:
--- a/Makefile
+++ b/Makefile
@@ -26,7 +26,6 @@ schema/%.context.jsonld: .FORCE
 schema/%.yaml sssom/%.yaml: .FORCE
 	wget $(SSSOM_YAML) -O $@
 
-
 test:
 	tox
 	sh tests/tests.sh
--- a/sssom/__init__.py
+++ b/sssom/__init__.py
@@ -1,5 +1,4 @@
 """Initializing imports and constants."""
-import os
 
 from .sssom_datamodel import slots  # noqa:401
 from .sssom_datamodel import Mapping, MappingSet  # noqa:401
@@ -11,6 +10,3 @@ from .util import (  # noqa:401
     group_mappings,
     parse,
 )
-
-SCHEMA_PARENT_PATH = os.path.abspath(os.path.join(os.path.dirname(__file__), os.pardir))
-SCHEMA_YAML = os.path.join(SCHEMA_PARENT_PATH, "schema/sssom.yaml")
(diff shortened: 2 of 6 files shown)


matentzn:
@cthoyt is this ok? I always get a little worried when using __FILE__ in python scripts..

cthoyt:
Yeah! This fix is great!

cthoyt (approved):
Only suggestion is I’d save the file path as a constant for reuse

matentzn:
Can you do this @hrshdhgd and do a pypi release?

sssom/writers.py
```diff
@@ -14,6 +13,7 @@
 from rdflib import Graph, URIRef
 from rdflib.namespace import OWL, RDF
 
+from . import SCHEMA_YAML
```
cthoyt, line 16: this might cause circular imports. better to make a constants file.

sssom/__init__.py
```diff
@@ -8,3 +11,6 @@
     group_mappings,
     parse,
 )
+
+SCHEMA_PARENT_PATH = os.path.abspath(os.path.join(os.path.dirname(__file__), os.pardir))
```
cthoyt, line 15: if this is referring to a file outside of the directory structure of the package, it simply won't work for anyone who pip installs it

cthoyt (changes requested):
Move the SSSOM configuration file inside python package structure otherwise this update is causes a new bug that results in the same undesired behavior

sssom/__init__.py
cthoyt: the other reason why it's nice to make constants is because it makes it easier to find bugs like this ^

sssom/writers.py
matentzn: there is a constants.py.. you can use that

sssom/__init__.py
matentzn: BUt if it is in the same directory as that __init__.py iys fine?

sssom/writers.py
hrshdhgd: That is in `tests`

hrshdhgd: I created a new one in `sssom`

cthoyt:
I added a test in https://github.com/mapping-commons/sssom-py/pull/182/commits/0c59fd00e7930eb6ac5293ebea58a55678c3fcf9 to demonstrate if the resource file is actually available during usage of the package

Makefile
```diff
@@ -51,6 +52,7 @@ deploy-dm: sssom/external_context.py sssom/internal_context.py
 	cp schema/sssom_datamodel.py sssom/
 	cp schema/sssom.context.jsonld sssom/
 	cp schema/sssom.external.context.jsonld sssom/
+	cp schema/sssom.yaml sssom/
```
cthoyt, line 55: why are there so many copies of this file now? just keep a single source of truth inside the python package

matentzn: schema is not in the python package. It should probably be removed from version control (this is where the current version from sssom is mirrored to). Not to worry in this PR though.

sssom/constants.py
```diff
@@ -0,0 +1,7 @@
+"""Constants."""
+
+import os
+import pathlib
+
+CWD = pathlib.Path(__file__).parent.resolve()
```
matentzn, line 6: This is the only line I cant judge. If you are positive this works with pip install, I am happy!

matentzn (dismissed):

matentzn:
I would sleep better if @cthoyt would give this another look.

Makefile
```diff
@@ -26,6 +26,7 @@ schema/%.context.jsonld: .FORCE
 schema/%.yaml: .FORCE
 	wget $(SSSOM_YAML) -O $@
 
+
```
cthoyt, line 29: why the spurious whitespace change?

Makefile
cthoyt: Since this PR is exacerbating the issue that would need fixing later, maybe it should be addressed first. I proposed an alternative solution that would enable a more clean solution that would supersede this PR in https://github.com/linkml/linkml-runtime/issues/86

cthoyt: It would also be possible to make the solution in SSSOM-py to getting/caching a specific version of the SSSOM yaml definition as a proof of concept, so it wouldn't be necessary for lots of discussion and a new release of linkml as well

sssom/constants.py
cthoyt: yes this does what you expect! however, `CWD` is a very misleading name, since this is not at all related to the notion of `os.getcwd()`. I'd suggest renaming this to `HERE`

```suggestion
HERE = pathlib.Path(__file__).parent.resolve()
```

sssom/writers.py
```diff
@@ -253,9 +253,10 @@ def to_rdf_graph(msdf: MappingSetDataFrame) -> Graph:
     # graph = graph.parse("sssom.ttl", format="ttl")
 
     # os.remove("sssom.ttl")  # remove the intermediate file.
+
```
cthoyt, line 256: why spurious whitespace change?

cthoyt:
This PR also needs a better title before merge

sssom/constants.py
hrshdhgd: Done

matentzn:
Cool, thanks @cthoyt

@hrshdhgd if tests pass you can merge!

cthoyt (approved):

hrshdhgd:
Thank you both @cthoyt and @matentzn !

Issue #183, opened 2021-12-07
sssom.yaml not found 


hrshdhgd:
When released to pypi and installed locally:

on running: `sssom convert test/tests/data/basic.tsv --output-format rdf`

I get the following error:

`FileNotFoundError: [Errno 2] No such file or directory: '/.../envs/test-sssom/lib/python3.9/site-packages/sssom/sssom.yaml'`

hrshdhgd:
When the tar file is downloaded separately, `sssom.yaml` does exist in the package folder

matentzn:
This issue makes me go ballistic man, and I am so Zen, I could autmeditate the Dalai Lama..

hrshdhgd:
I think I know the reason ... just need to prove it. `tox` has something to do with it.

hrshdhgd:
I think [this may be responsible](https://github.com/mapping-commons/sssom-py/blob/a2eb47254bc66d86e0ac5c05bdb8f5f54f78ed94/setup.cfg#L131) to support my comment above.

cthoyt:
> I think [this may be responsible](https://github.com/mapping-commons/sssom-py/blob/a2eb47254bc66d86e0ac5c05bdb8f5f54f78ed94/setup.cfg#L131) to support my comment above.
> 
> ```
> [coverage:paths]
> source =
>     sssom
>     .tox/*/lib/python*/site-packages/sssom
> ```

Nope, tox doesn't have anything to do with this. Check the way that the packaging is being done, specifically this https://github.com/mapping-commons/sssom-py/blob/a2eb47254bc66d86e0ac5c05bdb8f5f54f78ed94/setup.cfg#L85-L88

cthoyt:
This would be an excellent time to switch to src/ layout and the appropriate configuration for that which fixes a lot of these issues without any extra thought added

hrshdhgd:
Figured it out! [This](https://python-packaging.readthedocs.io/en/latest/non-code-files.html) was the missing link!

"In order for these files to be copied at install time to the package’s folder inside site-packages, you’ll need to supply include_package_data=True to the setup() function."

hrshdhgd:
#185 fixes this issue. I have released this to pypi and tested it. It works!

matentzn:
Awesome!

Issue #181, opened 2021-12-06
sssom.yaml cant be found in package


matentzn:
If you install sssom using `pip install --upgrade --force-reinstall --no-deps -e .` python is unable to find `sssom.yaml`. 

Reproduce by running sssom in a directory outside the source code directory (like in your `~/Documents/`).

```
Traceback (most recent call last):
  File "/Users/matentzn/opt/anaconda3/bin/sssom", line 8, in <module>
    sys.exit(main())
  File "/Users/matentzn/opt/anaconda3/lib/python3.8/site-packages/click/core.py", line 829, in __call__
    return self.main(*args, **kwargs)
  File "/Users/matentzn/opt/anaconda3/lib/python3.8/site-packages/click/core.py", line 782, in main
    rv = self.invoke(ctx)
  File "/Users/matentzn/opt/anaconda3/lib/python3.8/site-packages/click/core.py", line 1259, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/Users/matentzn/opt/anaconda3/lib/python3.8/site-packages/click/core.py", line 1066, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/Users/matentzn/opt/anaconda3/lib/python3.8/site-packages/click/core.py", line 610, in invoke
    return callback(*args, **kwargs)
  File "/Users/matentzn/opt/anaconda3/lib/python3.8/site-packages/sssom/cli.py", line 116, in convert
    convert_file(input_path=input, output=output, output_format=output_format)
  File "/Users/matentzn/opt/anaconda3/lib/python3.8/site-packages/sssom/io.py", line 35, in convert_file
    write_func(doc, output, serialisation=fileformat)  # type:ignore
  File "/Users/matentzn/opt/anaconda3/lib/python3.8/site-packages/sssom/writers.py", line 113, in write_owl
    graph = to_owl_graph(msdf)
  File "/Users/matentzn/opt/anaconda3/lib/python3.8/site-packages/sssom/writers.py", line 141, in to_owl_graph
    graph = to_rdf_graph(msdf=msdf)
  File "/Users/matentzn/opt/anaconda3/lib/python3.8/site-packages/sssom/writers.py", line 258, in to_rdf_graph
    schemaview=SchemaView(os.path.join(os.getcwd(), "schema/sssom.yaml")),
[...]

PR #185, opened 2021-12-08
Attempt 2 to solve the issue of sssom.yaml not showing in site-packages after pip install

hrshdhgd:
Relevant to issue #183 and follow-up to #184

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -5,6 +5,4 @@
 import setuptools
 
 if __name__ == "__main__":
-    setuptools.setup(
-        include_package_data=True
-    )
+    setuptools.setup(include_package_data=True)
--- a/setup.py
+++ b/setup.py
@@ -5,4 +5,6 @@
 import setuptools
 
 if __name__ == "__main__":
-    setuptools.setup()
+    setuptools.setup(
+        include_package_data=True
+    )
--- a/setup.py
+++ b/setup.py
@@ -5,4 +5,4 @@
 import setuptools
 
 if __name__ == "__main__":
-    setuptools.setup(include_package_data=True)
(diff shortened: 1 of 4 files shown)


hrshdhgd:
fixed #181 fixed #183

setup.py
```diff
@@ -5,4 +5,4 @@
 import setuptools
 
 if __name__ == "__main__":
-    setuptools.setup()
+    setuptools.setup(include_package_data=True)
```
cthoyt, line 8: put this in the `setup.cfg` under the `[options]` header. See https://cthoyt.com/2020/06/03/how-to-code-with-me-organization.html for more information on how to do packaging

tests/test_cli.py
```diff
@@ -73,12 +73,13 @@ def run_successful(self, result: Result, test_case: SSSOMTestCase) -> None:
 
     def run_convert(self, runner: CliRunner, test_case: SSSOMTestCase) -> Result:
         """Run the convert test."""
+        fmt = "owl"
```
cthoyt, line 76: relevant?

hrshdhgd: no ... I usually don't add irrelevant changes but this was minor, so thought of just pushing it.

cthoyt: kk :shipit:

setup.py
hrshdhgd: Done and tested. Works as expected!

cthoyt (approved):
I'm a little concerned that there's no specific git commit hash that corresponds to the releases, but that's an issue to solve elsewhere

Issue #188, opened 2021-12-21
Unpin linkml-runtime version


hrshdhgd:
The new release of `linkml-runtime` which is in turn dependent on `rdflib_shim` was causing errors in tests. [Issue created here](https://github.com/linkml/linkml/issues/529#issue-1085308851). Eventually we would like to use the latest versions as newer stable versions are released. The P.R. reflecting the pinned version is [here](https://github.com/mapping-commons/sssom-py/pull/163#issuecomment-998673725).

cmungall:
I think this is a fantom issue?

This PR works:

 *  #190

PR #191, opened 2021-12-21
unpinned linkml-runtime version

hrshdhgd:
Undoing pining of linkml-runtime as a solution used tin PR #163 . Addresses issue #188


hrshdhgd:
Seems like it works now.

PR #190, opened 2021-12-21
Testing, see https://github.com/linkml/linkml/issues/529

cmungall:

PR #193, opened 2022-01-06
unpinned rdflib

hrshdhgd:
Initiated by the conversation [here](https://github.com/mapping-commons/sssom-py/pull/190#issuecomment-1006758131)

- Unpinned `rdflib` from v5.0.0 to 6.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -58,7 +58,7 @@ install_requires =
     pandas
     pandasql
     pyyaml
-    rdflib==5.0.0
+    rdflib
     recommonmark>=0.7
     scikit_learn
     scipy


cmungall (approved):
you can go ahead and merge this

maybe also do a test where you pin to 6

cmungall:
aside:

```
    jsonasobj==1.3.1
    jsonasobj2>=1.0.4
```

why do we have this?

jsonasobj2 should be a transitive dependency, it should not be listed here at all, nevermind with such strict pinning. and AFAIK we have never needed jsonasobj

hrshdhgd:
I remember jsonasobj not being a requirement but was needed by linkml-runtime. It wasn't specified as a linkml-runtime dependency (which is now fixed) and so we had it in out requirements. I have removed the dependency and the tests pass. Opened new PR. #194

hrshdhgd:
Also pinned rdflib to v6. It works! Merging....

PR #204, opened 2022-02-14
make files - refresh sssom

matentzn:


schema/sssom.schema.json
```diff
@@ -320,7 +324,7 @@
             "subject_match_field": {
                "description": "A tuple of fields (term annotations on the subject) that was used for the match.",
                "items": {
-                  "$ref": "#/$defs/MatchFieldEnum"
+                  "type": "string"
```
hrshdhgd, line 327: does this mean, `MatchFieldEnum` is not an object anymore?

matentzn:
Yeah. We changed match field to entity feel free to merge

PR #213, opened 2022-03-04
Curie detection patch

hrshdhgd:
- Added test to anticipate wrong CURIE and `assertRasies(ValueError)`
- Raises error when CURIE is not legit.


tests/test_parsers.py
```diff
@@ -95,6 +99,16 @@ def test_parse_obographs(self):
...
             f"{self.obographs_file} has the wrong number of mappings.",
         )
+    
+    def test_broken_obographs(self):
+        """Test parsing OBO Graph JSON."""
+        
+        with self.assertRaises(ValueError):
+            msdf = from_obographs(
+            jsondoc=self.broken_obographs,
+            prefix_map=self.prefix_map,
```
matentzn, line 109: What I meant was injecting the prefix just before this row; the place you did change it up (external context) is updated automatically, so your changes will be overwritten!

matentzn (changes requested):

tests/test_parsers.py
hrshdhgd: something like this: `self.prefix_map["OMIM"] = "http://omim.org/entry/"`

tox.ini
```diff
@@ -37,6 +37,7 @@ deps =
     flake8
     flake8-black
     flake8-colors
+    bandit==1.7.2
```
hrshdhgd, line 40: Reason: https://github.com/PyCQA/bandit/issues/841

tests/test_parsers.py
matentzn: prefix_map = self.prefix_map
prefix_map["OMIM"] = "http://omim.org/entry/"

I think is better so this does not confuse any of the other checks (assuming that `prefix_map = self.prefix_map` creates a copy).

tox.ini
matentzn: Oh man :P Thanks

tests/test_parsers.py
hrshdhgd: Done.

matentzn (changes requested):

matentzn (changes requested):
If not curie, no hard fail, just logging.warn()

sssom/util.py
```diff
@@ -779,6 +779,10 @@ def to_mapping_set_dataframe(doc: MappingSetDocument) -> MappingSetDataFrame:
             m = get_dict_from_mapping(mapping)
             data.append(m)
     df = pd.DataFrame(data=data)
+    # The following 3 lines are to remove columns
+    # where all values are blank.
+    df = df.replace("", np.nan)
+    df = df.dropna(axis=1, how="all")
```
matentzn, line 785: I see you opted for this solution now. I think it is ok, let's roll with this for now. I can see some users complaining though when empty columns are indiscriminately dropped. Solving this cleanly may not be worth it right now, but just for documentation purposes, I will open a ticket.

matentzn (approved):
lets merge this together on monday.

sssom/util.py
```diff
@@ -965,3 +1000,82 @@ def is_multivalued_slot(slot: str) -> bool:
...
+        k
+        for k, v in prefix_synonyms.items()
+        if (k in prefix_map.keys() or v in prefix_map.keys())
+        and v not in expansion_replace.keys()
+    ]
+
+    if len(prefix_replace) > 0:
+        for pr in prefix_replace:
+            correct_prefix = prefix_synonyms[pr]
+            correct_expansion = prefix_expansion[correct_prefix]
```
matentzn, line 1053: What if the correct prefix is not in prefix expansions?

sssom/util.py
```diff
@@ -965,3 +1000,82 @@ def is_multivalued_slot(slot: str) -> bool:
...
+            if pr in prefix_map.keys():
+                prefix_map.pop(pr, None)
+                data_switch_dict[pr] = correct_prefix
+
+                logging.warning(f"Replacing prefix {pr} with {correct_prefix}")
+
+    # Data editing
+    if len(data_switch_dict) > 0:
+        # Read schema file
+        with open(SCHEMA_YAML) as file:
```
matentzn, line 1065: Isnt it better to use schemaview? I feel this here would break every time the yaml file changes a bit..

hrshdhgd: I can change it to use `SchemaView` but am not sure how this will break. I wasn't aware of `SchemaView` at the time of writing this.

sssom/util.py
```diff
@@ -965,3 +1000,82 @@ def is_multivalued_slot(slot: str) -> bool:
...
+
+    # The prefix exists but the expansion needs to be updated.
+    expansion_replace = {
+        k: v
+        for k, v in prefix_expansion.items()
+        if k in prefix_map.keys() and v != prefix_map[k]
+    }
+
+    # Updates expansions in prefix_map
+    prefix_map.update(expansion_replace)
```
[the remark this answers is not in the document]
hrshdhgd: I thought a little more about this and if your example reflects the YAML file, the logic will still have `ICD10CM: DEFEXPENSION` as the reconciliation. That code is lower than this line where you have the next comment.

sssom/util.py
hrshdhgd: I do not follow ... I'll need an example.

sssom/prefix_reconciliation.yaml
```diff
@@ -0,0 +1,5 @@
+custom_prefix_map:
```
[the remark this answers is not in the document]
hrshdhgd: we do have `reconciliation` as a parameter to `merge` but that does something else. See [here](https://github.com/mapping-commons/sssom-py/blob/bf9c32b7fa346361d7e4b5993b00204fa6060d7f/sssom/util.py#L508). That ha nothing to do with `prefix_reconciliation`. So you're saying the `prefix_reconcilation` logic should not reside in `sssom-py` at all ? I thought we had decided to keep it here. I'm confused.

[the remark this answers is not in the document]
matentzn: no the prefix reconciliation logic should definitely reside in soom, but not the prefixes to be reconciled (the yaml file) this should be passed in as a parameter like `--reconcile-prefixes`, not baked into sssom-py. This will change from case to case!

sssom/util.py
matentzn:
```
prefix_synonyms:
     ICD10CM: ICD10WHO
     uberon: UBERON
 prefix_expansion_reconciliation:
     ICD10WHO: DEFEXPENSION
```

?

sssom/prefix_reconciliation.yaml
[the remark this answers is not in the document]
hrshdhgd: got it!

sssom/util.py
hrshdhgd: Yup, that should replace `uberon` to `UBERON`. The `prefix_replace` list should have `uberon` which will then be replaced by `UBERON` and then later translate the change to the dataframe.

sssom/prefix_reconciliation.yaml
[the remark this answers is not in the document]
hrshdhgd: I've implemented this only for the `merge` CLI as of now. Is that what' expected?

sssom/cli.py
```diff
@@ -434,8 +442,19 @@ def merge(inputs: str, output: TextIO, reconcile: bool = True):
     """  # noqa: DAR101
     msdfs = [read_sssom_table(i) for i in inputs]
     merged_msdf = merge_msdf(*msdfs, reconcile=reconcile)
-    # Export MappingSetDataFrame into a TSV
-    write_table(merged_msdf, output)
+    # Reconcile prefixes if needed.
+    if reconcile_prefixes:
```
matentzn, line 446: What happens if msdf1 says:

```
ICD10: ABC
```

and msdf2 says:

```
ICD10: DEF
```

Wouldnt that cause a clash just before resolution?

Maybe we should create a new client method, `reconcile_prefixes()` that takes as an input a sssom table and applies the reconciled prefixes. Sorry about that, I didn't really think this through.. That way the client is responsible for reconciling prefixes before running the merge.. So you don't need the option in `merge` at all, just add to a new, stand alone method that reads a sssom table, reconciles, and writes again.

sssom/util.py
```diff
@@ -779,6 +805,11 @@ def to_mapping_set_dataframe(doc: MappingSetDocument) -> MappingSetDataFrame:
             m = get_dict_from_mapping(mapping)
             data.append(m)
     df = pd.DataFrame(data=data)
+    # The following 3 lines are to remove columns
+    # where all values are blank.
+    df = df.replace("", np.nan)
+    df = df.dropna(axis=1, how="all")
+    df.loc[:, df.columns != CONFIDENCE].replace(np.nan, "", inplace=True)
```
matentzn, line 812: Like in the other PR, shouldnt this be applied to all `double` columns?

hrshdhgd: It would have changed on merging with the other branch but I've changed it here too.

sssom/cli.py
hrshdhgd: If there is a `reconcile_prefix.yaml` file, that should take care of such clashes. But in the even of the absence, I think the prefix from `msdf1` supersedes the one in `msdf2`.

matentzn (approved):
Great job! It looks all good to me now, if we encounter problems, we can create small fixes.

PR #219, opened 2022-04-04
Get dict of schema

hrshdhgd:
Fixed identifying slots initialized as `empty_list()` from linkml-runtime.


sssom/util.py
```diff
@@ -208,7 +210,7 @@ def sort_sssom(df: pd.DataFrame) -> pd.DataFrame:
     :param df: SSSOM DataFrame to be sorted.
     :return: Sorted SSSOM DataFrame
     """
-    df.sort_values(by=sorted(df.columns), ascending=False, inplace=True)
+    df.astype(str).sort_values(by=sorted(df.columns), ascending=False, inplace=True)
```
matentzn, line 213: This should not be necessary, need to chase down why some fields are lists.

sssom/util.py
```diff
@@ -194,6 +194,8 @@ def parse(filename: str) -> pd.DataFrame:
 
 def collapse(df: pd.DataFrame) -> pd.DataFrame:
     """Collapse rows with same S/P/O and combines confidence."""
+    df[CONFIDENCE] = df[CONFIDENCE].replace("", np.nan)
+    df[CONFIDENCE] = df[CONFIDENCE].astype(float)
```
matentzn, line 198: Something like this should live where in to_mapping_set_data_frame, basically wherever a LinkML model gets converted to pandas (should be one place only)

sssom/util.py
```diff
@@ -231,6 +233,8 @@ def filter_redundant_rows(
     else:
         key = [SUBJECT_ID, OBJECT_ID, PREDICATE_ID]
     dfmax: pd.DataFrame
+    df[CONFIDENCE] = df[CONFIDENCE].replace("", np.nan)
```
matentzn, line 236: I feel this should be more principled.. We already say in LinkML that something should be interpreted as a double, so we need to retain that typing information. We don't want to add lines like this for every single numeric veriable.

sssom/util.py
```diff
@@ -257,7 +261,7 @@ def filter_redundant_rows(
         ]
     # We are preserving confidence = NaN rows without making assumptions.
     # This means that there are potential duplicate mappings
-    return_df = df.append(nan_df).drop_duplicates()
+    return_df = df.astype(str).append(nan_df).drop_duplicates()
```
matentzn, line 264: Should not be necessary!

sssom/util.py
```diff
@@ -497,7 +501,7 @@ def merge_msdf(
     # merge df [# 'outer' join in pandas == FULL JOIN in SQL]
     df_merged = reduce(
         lambda left, right: left.merge(right, how="outer", on=list(left.columns)),
-        [msdf.df for msdf in msdf_with_meta if msdf.df is not None],
+        [msdf.df.astype(str) for msdf in msdf_with_meta if msdf.df is not None],
```
matentzn, line 504: Should not be necessary, see above

sssom/util.py
```diff
@@ -571,8 +575,8 @@ def deal_with_negation(df: pd.DataFrame) -> pd.DataFrame:
         CONFIDENCE,
         MATCH_TYPE,
     ]
-    negation_subset = normalized_negation_df[columns_of_interest]
-    positive_subset = positive_df[columns_of_interest]
+    negation_subset = normalized_negation_df[columns_of_interest].astype(str)
```
matentzn, line 578: Should not be necessary, see above

sssom/util.py
```diff
@@ -571,8 +575,8 @@ def deal_with_negation(df: pd.DataFrame) -> pd.DataFrame:
...
+    positive_subset = positive_df[columns_of_interest].astype(str)
```
matentzn, line 579: Should not be necessary, see above

sssom/util.py
```diff
@@ -631,9 +635,11 @@ def deal_with_negation(df: pd.DataFrame) -> pd.DataFrame:
             PREDICATE_MODIFIER
         ].fillna("")
 
-    reconciled_df = df.merge(
-        reconciled_df_subset, how="right", on=list(reconciled_df_subset.columns)
-    ).fillna("")
+    reconciled_df = (
+        df.astype(str)
```
matentzn, line 639: Should not be necessary, see above

sssom/util.py

sssom/util.py

sssom/util.py
hrshdhgd: removed

sssom/util.py
```diff
@@ -495,10 +494,13 @@ def merge_msdf(
...
-    df_merged = reduce(
-        lambda left, right: left.merge(right, how="outer", on=list(left.columns)),
-        [msdf.df for msdf in msdf_with_meta if msdf.df is not None],
-    )
+    # df_merged = reduce(
+    #     lambda left, right: left.merge(right, how="outer", on=list(left.columns)),
+    #     [msdf.df for msdf in msdf_with_meta if msdf.df is not None],
+    # )
+    df_list = [msdf.df for msdf in msdf_with_meta if msdf.df is not None]
+    df_merged = pd.concat(df_list, axis=0, ignore_index=True)
```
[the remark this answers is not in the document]
hrshdhgd: The main reason I changed it was because it was giving an error stating that you cannot merge a float and object. I think I fixed that issue while working in other places. Reverting this back.

sssom/util.py
```diff
@@ -781,6 +783,9 @@ def to_mapping_set_dataframe(doc: MappingSetDocument) -> MappingSetDataFrame:
     df = pd.DataFrame(data=data)
     meta = extract_global_metadata(doc)
     meta.pop(PREFIX_MAP_KEY, None)
+    df.replace("", np.nan, inplace=True)
+    df = df.dropna(axis=1, how="all")  # remove columns with all row = 'None'-s.
+    df.loc[:, df.columns != CONFIDENCE].replace(np.nan, "", inplace=True)
```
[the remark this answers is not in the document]
hrshdhgd: I agree! Now that I know about `Schemaview` , I've changed the code for this to be dynamic.

sssom/util.py
```diff
@@ -792,18 +802,43 @@ def get_dict_from_mapping(map_obj: Union[Any, Dict[Any, Any], SSSOM_Mapping]) ->
     :param map_obj: Mapping object
     :return: Dictionary
     """
+    # TODO: Change logic
```
matentzn, line 805: Best remove that comment :)

sssom/util.py
```diff
@@ -792,18 +802,43 @@ def get_dict_from_mapping(map_obj: Union[Any, Dict[Any, Any], SSSOM_Mapping]) ->
...
+                    SCHEMA_DICT["slots"][property]["range"]
+                    in SCHEMA_DICT["enums"].keys()
+                ):
+                    map_dict[property] = map_obj[property].code.text
+                else:
+                    map_dict[property] = map_obj[property]
         else:
-            map_dict[property] = map_obj[property]
+            # IF map_obj[property] is None:
+            map_dict[property] = ""
```
matentzn, line 840: Is this ok, even if its `double`? Probably, just double checking.

hrshdhgd: Good catch! fixed it!

matentzn (approved):
The only thing that is not clear to me (feel free to fix in subsequent PR), is what happens if you serialised a table with a column with double values. will it say NaN everywhere there is no value?

hrshdhgd:
> The only thing that is not clear to me (feel free to fix in subsequent PR), is what happens if you serialised a table with a column with double values. will it say NaN everywhere there is no value?

Yes.

PR #222, opened 2022-04-11
Add metadata to the results to the `dosql` function

hrshdhgd:


matentzn (approved):
Such a weird method, sqldf(), but ok. Probably it nows how to use the last df in the script. Thanks!

PR #223, opened 2022-04-12
Added sort function and tested! fixes #51

### LINKED ISSUE #51: New command: sort

hrshdhgd:


sssom/cli.py
```diff
@@ -501,5 +502,20 @@ def reconcile_prefixes(input: str, reconcile_prefix_file: Path, output: TextIO):
     write_table(recon_msdf, output)
 
 
+@main.command()
+@input_argument
+@output_option
+def sort_columns(input: str, output: TextIO):
```
matentzn, line 508: Can we make a generic `sort` method that takes as an input two flags `--sort-columns` and `--sort-rows` (both true by default)?

sssom/util.py
```diff
@@ -657,9 +658,14 @@ def deal_with_negation(df: pd.DataFrame) -> pd.DataFrame:
...
 
+    # .fillna(df) towards the end fills an empty value
+    # with a corresponding value from df.
+    # This needs to happen because the columns in df
+    # not in reconciled_df_subset will be NaN otherwise
+    # which is incorrect.
     reconciled_df = df.merge(
         reconciled_df_subset, how="right", on=list(reconciled_df_subset.columns)
-    ).fillna("")
+    ).fillna(df)
```
matentzn, line 668: What does thta syntax mean (asking out of curiosity)

matentzn (changes requested):
Just one comment on the CLI, the rest looks great.

sssom/util.py
hrshdhgd: when we do a `right` join where the columns in the left DataFrame do not exist in the right DataFrame, the corresponding columns of the resultant DataFrame are `NaN` in spite of the left DataFrame for that row having data. The `df` basically fills the resultant DataFrame with appropriate data from the left one.

sssom/cli.py
hrshdhgd: Done!

hrshdhgd: I cannot redefine `sort` since it throws errors in `flake8` and `mypy` : 
```
Name "sort" already defined (possibly by an import)
F811 redefinition of unused 'sort' from line 32
```

matentzn (approved):

PR #233, opened 2022-04-26
Empty df

hrshdhgd:
Addressing empty pandas DataFrames

### DIFF
--- a/sssom/util.py
+++ b/sssom/util.py
@@ -944,9 +944,10 @@ def curie_from_uri(uri: str, prefix_map: Mapping[str, str]) -> str:
 def get_prefixes_used_in_table(df: pd.DataFrame) -> List[str]:
     """Get a list of prefixes used in CURIEs in key feature columns in a dataframe."""
     prefixes = []
-    for col in KEY_FEATURES:
-        for v in df[col].values:
-            prefixes.append(get_prefix_from_curie(v))
+    if not df.empty:
+        for col in KEY_FEATURES:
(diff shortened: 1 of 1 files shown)


sssom/util.py
```diff
@@ -944,9 +944,10 @@ def curie_from_uri(uri: str, prefix_map: Mapping[str, str]) -> str:
 def get_prefixes_used_in_table(df: pd.DataFrame) -> List[str]:
     """Get a list of prefixes used in CURIEs in key feature columns in a dataframe."""
     prefixes = []
-    for col in KEY_FEATURES:
-        for v in df[col].values:
-            prefixes.append(get_prefix_from_curie(v))
+    if not df.empty:
```
matentzn, line 947: Why not use same len() logic as in the other case?

sssom/util.py
```diff
@@ -944,9 +944,10 @@ def curie_from_uri(uri: str, prefix_map: Mapping[str, str]) -> str:
...
+        for col in KEY_FEATURES:
+            for v in df[col].values:
+                prefixes.append(get_prefix_from_curie(v))
     return list(set(prefixes))
```
matentzn, line 950: Not now, but we need to make sure that any sssom file is valid, so it should have at least all required columns. Don't worry in this PR

sssom/util.py
hrshdhgd: because we don't need to check columns and rows individually for this particular case.

matentzn (approved):

PR #234, opened 2022-04-27
Solves multiple issues

hrshdhgd:
- [x] merge when columns are not the same
- [x] make `_init_mapping_set()` robust
~~- [ ] `sssom parse` should convert `A rdfs:subClassOf B` into~~
  ~~- `A skos:broadMatch B` and~~
  ~~- `B skos:narrowMatch A`~~

### DIFF
--- a/sssom/parsers.py
+++ b/sssom/parsers.py
@@ -174,18 +174,23 @@ def _address_multivalued_slot(k: str, v: str) -> Union[str, List[str]]:
 
 
 def _init_mapping_set(meta: Optional[MetadataType]) -> MappingSet:
-    if all(["mapping_set_id", "license"]) in meta.keys():
-        return MappingSet(
-            mapping_set_id=meta["mapping_set_id"], license=meta["license"]
-        )
-    elif "mapping_set_id" in meta.keys():
-        return MappingSet(
-            mapping_set_id=meta["mapping_set_id"], license=DEFAULT_LICENSE
-        )
-    elif "license" in meta.keys():
-        return MappingSet(
-            mapping_set_id=DEFAULT_MAPPING_SET_ID, license=meta["license"]
-        )
+    if meta is not None:
+        if all(["mapping_set_id", "license"]) in meta.keys():
+            return MappingSet(
+                mapping_set_id=meta["mapping_set_id"], license=meta["license"]
+            )
+        elif "mapping_set_id" in meta.keys():
+            return MappingSet(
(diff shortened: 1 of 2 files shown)


hrshdhgd:
@matentzn , Does the second bullet make sense?

matentzn:
I am not so sure anymore, maybe just extract rdfs:subClassOf as the predicate instead of using skos:match. Will think about it tomorrow.

sssom/parsers.py
```diff
@@ -175,9 +175,22 @@ def _address_multivalued_slot(k: str, v: str) -> Union[str, List[str]]:
 
 def _init_mapping_set(meta: Optional[MetadataType]) -> MappingSet:
     if meta is not None:
-        return MappingSet(
-            mapping_set_id=meta["mapping_set_id"], license=meta["license"]
-        )
+        if all(["mapping_set_id", "license"]) in meta.keys():
```
matentzn, line 178: Just curious, do you like this way to if..else more than:

```
license=DEFAULT_LICENSE
mapping_set_id=DEFAULT_MAPPING_SET_ID
if meta is not None:
        if "mapping_set_id" in meta.keys():
                mapping_set_id=meta["mapping_set_id"]
        if "license" in meta.keys():
                license = meta["license"]
return MappingSet(
                mapping_set_id=mapping_set_id, license=license
            )
```

sssom/util.py
```diff
@@ -521,10 +521,15 @@ def merge_msdf(
...
 
     # merge df [# 'outer' join in pandas == FULL JOIN in SQL]
+    # df_merged = reduce(
+    #     lambda left, right: left.merge(right, how="outer", on=list(left.columns)),
+    #     [msdf.df for msdf in msdf_with_meta if msdf.df is not None],
+    # )
+    # Concat is an alternative to merge when columns are not the same.
     df_merged = reduce(
-        lambda left, right: left.merge(right, how="outer", on=list(left.columns)),
+        lambda left, right: pd.concat([left, right], axis=0, ignore_index=True),
```
matentzn, line 530: I trust you here, dont understand the implications :)

hrshdhgd: I've tested it with our unit test file, did a `df1.equals(df2)` which equates to `True`

sssom/parsers.py
hrshdhgd: This is better! I'll change it.

matentzn (approved):

PR #236, opened 2022-04-29
Adding mapping predicates to parse()

matentzn:
- Adding optional mapping_predicate options to parse methods
- renaming to "read" methods to "parse" - there is a conceptual distinctions here as read methods pertain to natively supported formats, and parse methods are to external formats, like obographs or alignment api.

Addresses #237


sssom/cli.py
```diff
@@ -89,6 +89,28 @@
...
+predicate_filter_option = click.option(
+    "-F",
+    "--mapping-predicate-filter",
+    multiple=True,
+    help="A list of predicates or a file path containing the list of predicates to be considered.",
+)
+
+
+def _get_list_of_predicates(predicate_filter: tuple) -> list:
+
```
matentzn, line 101: What a out the case ["a:123", "list.txt", "b:234"]?

sssom/cli.py
```diff
@@ -89,6 +89,28 @@
...
+def _get_list_of_predicates(predicate_filter: tuple) -> list:
+    
+    if all(":" in pred for pred in list(predicate_filter)):
+        mapping_predicates = list(predicate_filter)
+        # TODO: Cross-check if the predicate is a valid one 
+        # For e.g. check for typos etc.
+    elif os.path.isfile(predicate_filter[0]):
+        with open(predicate_filter[0], "r") as f:
+            mapping_predicates = f.read().splitlines()
+    else:
```
matentzn, line 109: Should be possible to be empty

sssom/cli.py
```diff
@@ -89,6 +89,28 @@
     help="Fields.",
 )
 
+predicate_filter_option = click.option(
+    "-F",
+    "--mapping-predicate-filter",
+    multiple=True,
```
matentzn, line 95: What does multiple do? Will it parse a space separated string into a list?

hrshdhgd: Yes

sssom/cli.py
hrshdhgd: Do we want this or enforce either a list OR individual CURIES?

hrshdhgd: Changed the code to accept the above.

sssom/cli.py
hrshdhgd: added an if statement below before calling the function.

sssom/cli.py
```diff
@@ -89,6 +91,32 @@
...
+def _get_list_of_predicate_iri(predicate_filter: tuple) -> list:
+    pred_filter_list = list(predicate_filter)
+    preds = [p for p in pred_filter_list if ":" in p]
+    preds_iri = [get_iri(p) for p in preds]
+
+    if len(preds) != len(pred_filter_list) and len(preds) > 0:
+        # The user passed file paths too.
+        pred_fps = [p for p in pred_filter_list if p not in preds]
+        if all(os.path.isfile(p) for p in pred_fps):
+            pred_list = list(
```
matentzn, line 111: we should get a warning if a user supplied option is neither a valid curie, nor a valid file path

sssom/cli.py
```diff
@@ -89,6 +91,32 @@
...
 
+predicate_filter_option = click.option(
+    "-F",
+    "--mapping-predicate-filter",
+    multiple=True,
+    help="A list of predicates or a file path containing the list of predicates to be considered.",
+)
+
+
+def _get_list_of_predicate_iri(predicate_filter: tuple) -> list:
```
matentzn, line 102: we should have no methods in cli that are not 100% cli() related. This method here is a bit on the fence. don't we do all the argument parsing usually in io.py? Not sure whats best here.

sssom/cli.py
```diff
@@ -153,16 +182,30 @@ def parse(
...
+        parse_file(
+            input_path=input,
+            output=output,
+            input_format=input_format,
+            metadata_path=metadata,
+            prefix_map_mode=prefix_map_mode,
+            clean_prefixes=clean_prefixes,
+            mapping_predicates=mapping_predicates,
+        )
+    else:
```
matentzn, line 200: I did this as well in parsers.py but I really hate it... Is there no better way? Using args *? Maybe there is a better design?

matentzn: In this case here, in any case, let the if-else be in parse_file()? If mapping_set is passed as`None`, it should work fine.

sssom/parsers.py
```diff
@@ -53,10 +59,15 @@ def read_sssom_table(
     file_path: Union[str, Path],
     prefix_map: Optional[PrefixMap] = None,
     meta: Optional[MetadataType] = None,
+    mapping_predicates: Optional[List[str]] = None,
```
matentzn, line 62: i basically thought to not have the `mapping_predicates` option for all the `read_` methods - only for the `parse_` methods in `parsers.py`...

sssom/cli.py
matentzn: Maybe something like this, but even here I am unsure: https://www.geeksforgeeks.org/python-star-or-asterisk-operator/#:~:text=Here%20single%20asterisk(%20*%20)%20is,and%20variable%2Dlength%20argument%20list.

sssom/cli.py
```diff
@@ -21,6 +22,7 @@
 import click
 import pandas as pd
 import yaml
+from bioregistry import get_iri
```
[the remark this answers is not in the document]
hrshdhgd: I doubt it.

sssom/cli.py
```diff
@@ -89,6 +91,32 @@
...
+    "-F",
+    "--mapping-predicate-filter",
+    multiple=True,
+    help="A list of predicates or a file path containing the list of predicates to be considered.",
+)
+
+
+def _get_list_of_predicate_iri(predicate_filter: tuple) -> list:
+    pred_filter_list = list(predicate_filter)
+    preds = [p for p in pred_filter_list if ":" in p]
```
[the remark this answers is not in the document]
hrshdhgd: Done!

sssom/cli.py
hrshdhgd: Done

sssom/cli.py
hrshdhgd: I think it'll work with the if-else statement. There's no need for the `*`. I've changed it.

sssom/cli.py
hrshdhgd: moved to io.py

sssom/io.py
```diff
@@ -120,3 +138,27 @@ def get_metadata_and_prefix_map(
...
+
+def get_list_of_predicate_iri(predicate_filter: tuple) -> list:
+    """Return a list of IRIs for predicate CURIEs passed.
+
+    :param predicate_filter: CURIE OR list of CURIEs OR file path containing the same.
+    :return: A list of IRIs.
+    """
+    pred_filter_list = list(predicate_filter)
+    preds = [p for p in pred_filter_list if is_curie(p)]
+    preds_iri = [get_iri(p) for p in preds]
```
matentzn, line 151: What happens if get_iri() fails? I still think we need a way to configure get_iri() with a specific prefixmap..

matentzn: Looking at https://github.com/biopragmatics/bioregistry/blob/main/src/bioregistry/resolve_identifier.py#L249

Cany you simply pass our prefix_map in? Also set use_bioregistry_io to false, just in case.

hrshdhgd: Done! I had to move the implementation of `get_list_of_predicate_iri()` in `io.py` instead of `cli.py` since that's where I could get the prefix_map.

hrshdhgd: For some reason I cannot ask you to review it... so weird.

matentzn: Its because I created the branch

PR #238, opened 2022-05-02
Fix bug with default metadata

matentzn:
Fixing bug where `get_default_metadata()` was processing a completely wrong return type! That is why --prefix-map-mode didn't work anywhere!


hrshdhgd (approved):
Aah! Nice catch!!!

PR #240, opened 2022-05-03
Fix all warnings

### LINKED ISSUE #227: `sssom parse`: `license` field not found, even though it's in metadata `.yml`
## Description
I have a `license` field in my `.yml` file, and I'm passing it to `sssom parse`, but I get a warning message saying that it can't find `license`.

Command I'm running
`sssom parse omim.json -I obographs-json -m data/metadata.sssom.yml -o omim.sssom.tsv`

`metadata.sssom.yml`:
```yml
mapping_set_id: http://example.org
license: http://w3id.org/sssom/license/unspecified
creator_id: https://orcid.org/0000-0002-2906-7319
curie_map:
  HGNC: "https://identifiers.org/hgnc:"
  HGNC_symbol: https://www.genenames.org/data/gene-symbol-report/#!/hgnc_id/
  OMIM: https://omim.org/entry/
  UMLS: http://linkedlifedata.com/resource/umls/id/
  ORPHA: http://www.orpha.net/ORDO/Orphanet_

```

## Error/warnings
`WARNING:root:No License provided, using http://w3id.org/sssom/license/unspecified`

## What I've tried
I looked at `sssom/context.py` to see if I could find anything wrong, but I don't see an obvious issue.

```py
def set_default_license(meta: Metadata) -> Metadata:
    if ("license" not in meta.metadata) or (meta.metadata["license"] is None):
        meta.metadata["license"] = DEFAULT_LICENSE
        logging.warning(f"No License provided, using {DEFAULT_LICENSE}")
    return meta
```

I'm assuming that the contents of `meta.metadata` should be equivalent to the root of my metadata `.yml` file. If that's the case, I don't know why there's an issue here. Perhaps when it reads the `.yml` file, something happens where the value of the `license` does not get set correctly in `meta.metadata`?

I can probably figure this out on my own using the debugger, but just wanted to report it first.

### LINKED ISSUE #228: Make sure that `parse` can work without needing to pass a metadata `.yml`
See: https://github.com/mapping-commons/sssom-py/issues/227#issuecomment-1101205374

The following should work w/out needing a metadata `.yml`:
`sssom parse omim.json -I obographs-json -o omim.sssom.tsv`

It should warn and populate with defaults if any of the following are missing:
- license
- (the other ones in the yaml)

And should have the prefixes for HGNC, Orphanet, etc loaded.

### LINKED ISSUE #239: Pandas df warning
```
A value is trying to be set on a copy of a slice from a DataFrame

See the caveats in the documentation: https://pandas.pydata.org/pandas-docs/stable/user_guide/indexing.html#returning-a-view-versus-a-copy
  return super().replace(
```

This is caused by this section in `to_mapping_set_dataframe(doc: MappingSetDocument) -> MappingSetDataFrame` (util.py):
:

```
df = df.dropna(axis=1, how="all").copy()  # remove columns with all row = 'None'-s.
df.loc[:, ~df.columns.isin(slots_with_double_as_range)].replace(
        np.nan, "", inplace=True
    )
```

hrshdhgd:
- [x]  Fixes #239 
~~- [ ] Fixes #228~~
~~- [ ] Fixes #227~~ 
Both above fixed by #238
 - [x] Fix warning: `FutureWarning: The frame.append method is deprecated and will be removed from pandas in a future version. Use pandas.concat instead.`

### DIFF
--- a/sssom/parsers.py
+++ b/sssom/parsers.py
@@ -337,7 +337,7 @@ def from_sssom_dataframe(
     """
     prefix_map = _ensure_prefix_map(prefix_map)
 
-    # Need to revisit this solution. 
+    # Need to revisit this solution.
     # This is to address: A value is trying to be set on a copy of a slice from a DataFrame
     if "confidence" in df.columns:
         df2 = df.copy()
--- a/sssom/parsers.py
+++ b/sssom/parsers.py
@@ -337,8 +337,12 @@ def from_sssom_dataframe(
     """
     prefix_map = _ensure_prefix_map(prefix_map)
 
+    # Need to revisit this solution. 
+    # This is to address: A value is trying to be set on a copy of a slice from a DataFrame
     if "confidence" in df.columns:
-        df["confidence"].replace(r"^\s*$", np.NaN, regex=True, inplace=True)
+        df2 = df.copy()
+        df2["confidence"].replace(r"^\s*$", np.NaN, regex=True, inplace=True)
+        df = df2
 
     mlist: List[Mapping] = []
     ms = _init_mapping_set(meta)
--- a/sssom/util.py
+++ b/sssom/util.py
@@ -827,10 +827,9 @@ def to_mapping_set_dataframe(doc: MappingSetDocument) -> MappingSetDataFrame:
     # The following 3 lines are to remove columns
     # where all values are blank.
     df.replace("", np.nan, inplace=True)
-    df = df.dropna(axis=1, how="all")  # remove columns with all row = 'None'-s.
-    df.loc[:, ~df.columns.isin(slots_with_double_as_range)].replace(
-        np.nan, "", inplace=True
-    )
+    df.dropna(axis=1, how="all", inplace=True)  # remove columns with all row = 'None'-s.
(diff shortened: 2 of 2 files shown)


sssom/parsers.py
```diff
@@ -337,8 +337,12 @@ def from_sssom_dataframe(
     """
     prefix_map = _ensure_prefix_map(prefix_map)
 
+    # Need to revisit this solution.
+    # This is to address: A value is trying to be set on a copy of a slice from a DataFrame
     if "confidence" in df.columns:
-        df["confidence"].replace(r"^\s*$", np.NaN, regex=True, inplace=True)
+        df2 = df.copy()
```
matentzn, line 343: Why does this solve the issue? is "df" already a slice (subset) of some original data frame?

matentzn (approved):
I cant see anything wrong here, just my typical concerns of pandas data frame code being excessively verbose :) Thanks @hrshdhgd

sssom/parsers.py
hrshdhgd: When I do the `replace()`  `inplace`, it performs the operation on a slice and hence a `.copy()` ensures that's not the case.

PR #246, opened 2022-05-08
Refactoring prefixmap mode code

matentzn:
- Moving prefix map mode to constants
- Fixing some (unrelated) issue with default values in prefix map generation
- Add error checking for prefix map mode


hrshdhgd (commented):
Both PRs look good

hrshdhgd (approved):

Issue #217, opened 2022-03-18
Improve sssom-py documentations


cmungall:
- https://mapping-commons.github.io/sssom-py/installation.html
   * [ ] end-users should not have to install from git. use pypi
   * [ ] change cmungall to mapping-commons
   * [ ] consider consolidating docs aimed at developers in one place. keep the main docs aimed at the end-user
- ordering of pages
   * consult some of our other documentation sites for a better flow here (boomer, linkml, robot, ...)
   * [ ] the first page should be a general intro page, not a confusing index labeled SSSOM-PY in shouty caps :-)
   * [ ] put a link to (a) the general SSSOM docs (b) the paper in the intro
   * [ ] don't label pages generically like "Documentation" (https://mapping-commons.github.io/sssom-py/documentation.html). It's all documentation! This should be "python API documentation" or similar
   * [ ] https://mapping-commons.github.io/sssom-py/examples.html -- this needs filled out
- https://mapping-commons.github.io/sssom-py/cli_usage.html
   * [ ] remove the pointless extra header level
   * [ ] the CLI guide is fantastic, this is what most people use, put this BEFORE python API docs, which most won't use
   * [ ] is it `sssom` or `sssom-py`? the guide is inconsistent
   * [ ] link to boomer docs here https://mapping-commons.github.io/sssom-py/cli_usage.html?highlight=merge#sssom-py-ptable
   * [ ] is the warning here true? https://mapping-commons.github.io/sssom-py/cli_usage.html?highlight=merge#sssom-py-convert if so make issues
   * [ ] in general the docstrings for each of these commands could do with giving a bit more context
   * [ ] what is the difference between parse and convert?
   * [ ] https://mapping-commons.github.io/sssom-py/cli_usage.html?highlight=merge#sssom-py-sparql - give an example
- https://mapping-commons.github.io/sssom-py/sphinx_documentation.html
   * [ ] this isn't of interest to a general user. consider folding into a more general developer docs page where we also talk about style guide black etc and CONTRIBUTING.md
   * [ ] fix formatting errors e.g. `::`
- Other
   * [ ] remove the "Edit on github" button, it 404s
   * [ ] is it really necessary to have the tests here https://mapping-commons.github.io/sssom-py/documentation.html

general comments: put yourself in the mindset of a naive user and try and write docs that help them. think of times when you have had to install and run a package in a project you are not associated with, what helped you, what didn't

matentzn:
[...]

PR #249, opened 2022-05-12
Improve Documentation

hrshdhgd:
Fixes issue #217 

- [x]  end-users should not have to install from git. use pypi
- [x] change cmungall to mapping-commons
- [x] consider consolidating docs aimed at developers in one place. keep the main docs aimed at the end-user

### DIFF
--- a/docs/_sources/index.rst.txt
+++ b/docs/_sources/index.rst.txt
@@ -15,7 +15,7 @@ https://github.com/OBOFoundry/SSSOM
 sssom-py is a Python library and command line toolkit for working with
 SSSOM:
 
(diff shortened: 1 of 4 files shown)


hrshdhgd:
@matentzn , I have addressed a couple of bullets mentioned by Chris. I also changed the architecture for docs to auto-deploy to the `gh-pages` branch. I'll need some help from you to cover the remaining bullets and then this PR will be complete. For debugging purposes, the workflow updates docs on commit to this PR. I will eventually change it to do the same on merging the PR.

matentzn (approved):
I think we can merge/try/iterate. Looks nice!

Issue #21, opened 2021-03-10
New command: validate


matentzn:
Simple variation of the parser that returns true or false and warnings.

cthoyt:
Was just thinking about how this would be nice. Want me to start a simple PR?

matentzn:
Have at it! I could really use your help also with the architecture of this code base.. If you have any suggestions, let me know. Its very hacky at the moment.

hrshdhgd:
Waiting to confirm if the above commit satisfies this ticket before closing it.

cmungall:
@hrshdhgd - are these commits part of a PR?

hrshdhgd:
yes. `cli_edit`. Still working on it.

cmungall:
always make a PR - you can mark it as draft if not ready

hrshdhgd:
I meant #50 is the draft

matentzn:
I think for the validate command, it would be good if:

1. It would check @cthoyt and my 5 star  proposal https://github.com/mapping-commons/SSSOM/pull/69 and implement 3 and 4 star computationally
2. Make sure that all fields that can take multiple values are implemented correctly (#124)

matentzn:
Doing this well is now getting to the top of the list.

matentzn:
Prerequisite for this issue here is https://github.com/mapping-commons/sssom-py/issues/124

hrshdhgd:
Fixed by #270

PR #251, opened 2022-05-16
Create omop2obo-fake.sssom.tsv

matentzn:
@hrshdhgd 

This mapping file should fail because it is missing prefixes, but it does not. It produces some default output and I don't understand at a cursory glance how. Can we make sure that a case like this, where prefixes are used but not declared, fails?

This should be done generically, maybe as part of #21


matentzn:
No longer needed, we now have detailed tests for missing prefixes.

PR #255, opened 2022-05-17
Added `--version` to display version in CLI

### LINKED ISSUE #243: Add something like sssom --version
Would be convenient especially for me to understand which version I am using atm, as I am always dealing with a wild mix of pip install github, `.`, pypi..

hrshdhgd:
Fixes #243

### DIFF
--- a/sssom/cli.py
+++ b/sssom/cli.py
@@ -48,6 +48,7 @@ from .util import (
     to_mapping_set_dataframe,
 )
 from .writers import write_table
+import pkg_resources
 
 # Click input options common across commands
 input_argument = click.argument("input", required=True, type=click.Path())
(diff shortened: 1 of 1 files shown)


matentzn (commented):
Have you tried installing in clean environment and running? I remember some issues with pkg_resources and setup tools in the past..

hrshdhgd:
You are correct! Looking for an alternative ay to do this.

hrshdhgd:
Using `import importlib.metadata`.

hrshdhgd:
`importlib` varies for Python < 3.8 and >= 3.8 . Crappy solution but is a solution. Do we still need to support 3.7?

hrshdhgd:
Tested in 3.7 nd 3.9 by installing from `git` and both work!

matentzn (approved):
Excellent!

PR #261, opened 2022-05-31
upgraded to 310

hrshdhgd:
Basically mirrors what Charle did in #167 with an addition of editing the `toml` file which I think isn't in his PR and hence it didn't work (my theory).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.black]
 # line-length = 100
-target-version = ['py36', 'py37', 'py38', 'py39', 'py310']
+target-version = ['py37', 'py38', 'py39', 'py310']
 
 [tool.isort]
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.black]
 # line-length = 100
-target-version = ['py36', 'py37', 'py38', 'py39']
+target-version = ['py36', 'py37', 'py38', 'py39', 'py310']
```
matentzn, line 3: shouldnt we remove py36 if we dont test against it anymore

matentzn: from here an setup. cfg?

hrshdhgd: Done!

matentzn (approved):

PR #273, opened 2022-06-24
Add inverse_factor handling to ptable

matentzn:
I just saw that the ptable method itself included the parameter. So why not add it.. What do you think @hrshdhgd maybe I am not seeing something.


hrshdhgd:
It just blindly assigns a hard-coded inverse factor. Doesn't `boomer` calculate it off the confidence? and then do something more with it? I think that's why it was just a placeholder in the code with the code `f"inverse_factor ({inverse_factor}) ignored by this method, not implemented yet."`

matentzn:
Look at the potable method though - it uses the parameter clearly.

hrshdhgd (approved):

matentzn:
Probably not all that needed to be done :) Just seemed logical to me to add this connection - since the parameters already exist as defaults.

cmungall (approved):

PR #274, opened 2022-06-28
Fix yaml list

matentzn:


hrshdhgd:
No, you just undid what I added. With the code you have, when we `write` to JSON, the multivalued slots get written as a list of list which we do not want. Hence the `','.join()` which makes the output more in the lines of what we expect it to be.

PR #277, opened 2022-06-29
GH Action workflow for PyPI release

### LINKED ISSUE #186: Start using GitHub releases and autopushing to PyPi
For code versioning, I think we should start using GitHub releases.

We probably want some GitHub action as well that pushes a new module to pypi with each release, and a dev- version with each merge to master..

hrshdhgd:
Fixes #186


hrshdhgd:
It worked!

PR #278, opened 2022-06-30
Add `--embedded-mode` param to `parse`

### LINKED ISSUE #242: Add binary option `--embedded-mode` to `convert`
sssom convert should get a new binary option (flag), `--embedded-mode`, which is `True` by default. If false, the sssom file should be saved as 

- filename (whatever we have in -o) (table data)
- filename.yml (metadata header)

As a special case: if the filename ends with the canonical `.sssom.tsv`, the yaml file should be stored as `.sssom.yml`.

hrshdhgd:
Fixes #242

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -3,7 +3,7 @@
 ##########################
 [metadata]
 name = sssom
-version = 0.3.12.dev
+version = 0.3.12
 description = Operations on SSSOM mapping tables
 long_description = file: README.md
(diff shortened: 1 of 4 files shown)


matentzn (approved):
It looks great to me! I assume you tried it with "embedded" false locally, if so, feel free to merge. Thank you!

hrshdhgd:
Yes, tried it and got expected results.

PR #283, opened 2022-07-07
 Validate looks for prefix map mismatches and throws `ValueError`

### LINKED ISSUE #275: Validate should fail if built in prefixes are redefined
No built in prefix should ever be re-defined, see

https://github.com/mapping-commons/sssom-py/blob/master/sssom/context.py#L76

### LINKED ISSUE #276: Add validation step before writing rdf file
See `writers.py`. commit: 46148e6

For now, lets just run PrefixValidation.

This makes sense to avoid <orcid:123> type nodes in the RDF graph

hrshdhgd:
- [x] Fixes #275 
 - [x] Fixes #276 
 - [x] Fix test files.


hrshdhgd:
@matentzn , how do we implement `add_built_in_prefixes_to_prefix_map` in validation?

sssom/validators.py
```diff
@@ -70,6 +71,11 @@ def check_all_prefixes_in_curie_map(msdf: MappingSetDataFrame) -> None:
     :raises ValidationError: If all prefixes not in curie_map.
     """
     prefixes = get_all_prefixes(msdf)
+    prefixes_including_builtins = add_built_in_prefixes_to_prefix_map(msdf.prefix_map)
+    msdf.prefix_map = prefixes_including_builtins
+    # TODO: If prefixes_including_builtins NOT EQUAL to msdf.prefix_map
+    # 1. Raise error OR
+    # 2. Warning that staes we added built-ins.
```
matentzn, line 78: I think its fine. I anything you can add a logging.info() - not adding built-ins to the prefix map should not be considered a problem!

matentzn (approved):
Excellent, looks great!

sssom/validators.py
```diff
@@ -20,15 +22,14 @@ def validate(
...
-    :return: Validation error or None.
     """
     validation_methods = {
         SchemaValidationType.JsonSchema: validate_json_schema,
         SchemaValidationType.Shacl: validate_shacl,
         SchemaValidationType.PrefixMapCompleteness: check_all_prefixes_in_curie_map,
     }
     for vt in validation_types:
-        return validation_methods[vt](msdf)
+        validation_methods[vt](msdf)
```
matentzn, line 32: oh.. good point

sssom/validators.py
```diff
@@ -70,6 +71,16 @@ def check_all_prefixes_in_curie_map(msdf: MappingSetDataFrame) -> None:
     :raises ValidationError: If all prefixes not in curie_map.
     """
     prefixes = get_all_prefixes(msdf)
+    prefixes_including_builtins = add_built_in_prefixes_to_prefix_map(msdf.prefix_map)
+    uncommon_maps = {
```
matentzn, line 75: rename to "not_built_in" for clarity?

sssom/validators.py
```diff
@@ -70,6 +71,16 @@ def check_all_prefixes_in_curie_map(msdf: MappingSetDataFrame) -> None:
...
+        k: v
+        for k, v in msdf.prefix_map.items()
+        if k not in prefixes_including_builtins.keys()
```
matentzn, line 78: Is the purpose of this to print out the built_in_prefixes that were previously not in the prefix map, and now were added? If so, shouldn't you loop through prefixes_including_builtins and ehck if `k not in msdf.prefix_map` instead?

tests/data/basic3.tsv
```diff
@@ -1,25 +1,27 @@
...
 #curie_map:
 #  a: "http://example.org/a/"
 #  b: "http://example.org/b/"
 #  c: "http://example.org/c/"
 #  d: "http://example.org/d/"
-#  rdfs: "http://example.org/rdfs/"
-#  owl: "http://example.org/owl/"
+#  rdfs: "http://www.w3.org/2000/01/rdf-schema#"
+#  owl: "http://www.w3.org/2002/07/owl#"
+#  orcid: "https://orcid.org/my-orcid?orcid="
```
matentzn, line 14: That is not the correct orcid prefix :P I guess it does not matter, of course, but just FYI.: `https://orcid.org/0000-0002-7356-1779`

sssom/validators.py
hrshdhgd: The purpose here is to check that all prefixes in the `msdf.df` are defined in the `curie_map`, that's all.

matentzn (approved):

PR #284, opened 2022-07-08
Add new command filter

### LINKED ISSUE #30: New command: filter
We need a filter command that allows us to subset a mapping table for more common filter operations easily. The alternative is a powerful filter language that will end up reimplementing SQL, which we already have in DOSQL. So for the moment, these are my most important reasons to filter:

- I want to get rid of unused prefixes in the prefix map
- I want to get mappings only between mondo and do terms
- I only want exact and broad matches, but not oboInOwl:hasDbXref

There are many more things I want, but I can do the rest with dosql

I am thinking of something like this:

```
sssom filter input.sssom.tsv --prefixes "MP HP MONDO" --predicates "skos:exactMatch skos:broadMatch"
```

I am also ok with 

```
sssom filter input.sssom.tsv --prefix MP --prefix HP --prefix MONDO --predicate_id skos:exactMatch --predicate_id skos:broadMatch
```

If that feels cleaner, but the selection logic in the prefix case should be conjunctive, i.e. the Mapping should not have any other prefix in the subject or object ids other than the ones provided. 

There is some case to be made for this design as well:

```
sssom filter input.sssom.tsv --subject_id MP:* --subject_id MP:* --object_id MP:* --predicate_id skos:exactMatch --predicate_id skos:broadMatch
```

then the subject_id parameter (multivalued) should be conjunctive. Actually, I kinda like this  last  proposal now. It should follow the same dynamic parameter logic as #280?

hrshdhgd:
- [x] Edit test files to have diverse predicates
 - [x] Change `assert` values that are impacted due to the above change
 - [x] Add new CLI command `filter` that filters MSDF based on prefixes and predicates provided through CLI.
 - [x] Add `test_filter.py`
 - [x] Add function `filter_file` that the CLI calls
 - [x] Add feature to filter by sssom fields
 - [x] Add documentation

Fixes #30


hrshdhgd:
Oh... I just noticed, that you wanted to filter by sssom fields. I haven't implemented that so far. For now I have implemented the following:

- You pass input file (sssom.tsv) with params: `prefix`, `predicates` and `output`.
- The prefixes provided filter:
  - The `prefix_map`
  - Looks in `subject_id` and `object_id` for prefix existence
- The predicates provided filter:
  - Associated prefixes in the `prefix_map` 
  - Looks in the `predicate_id` column for predicate existence.

We could add a feature to filter by column names too. I will just need to run it through with you before coding anything.

matentzn:
This looks great, lets look at this during our call today!

sssom/io.py
```diff
@@ -228,3 +232,114 @@ def extract_iri(input, prefix_map) -> list:
...
+    query = "SELECT * FROM df WHERE ("
+    multiple_params = True if len(params) > 1 else False
+    for idx, (k, v) in enumerate(params.items(), start=1):
+        query += k + " LIKE '" + v[0] + "' "
+        if len(v) > 1:
+            for exp in v[1:]:
+                query += " OR "
+                query += k + " LIKE '" + exp + "') "
+        if multiple_params and idx != len(params):
+            query += " AND ("
```
matentzn, line 344: Does it make sense to throw an exception of no legal parameter is supplied? Just to be sure you don't try to query anything if there is not filter criteria?

tests/test_config.yaml
```diff
@@ -5,18 +5,18 @@ tests:
     ct_json_elements: 13
     ct_data_frame_rows: 141
     ct_graph_queries_owl:
-      query_count_equivalent_classes: 90
+      query_count_equivalent_classes: 84
```
matentzn, line 8: Why did this happen?

tests/test_filter.py
```diff
@@ -0,0 +1,40 @@
...
+        filtered_msdf = filter_file(
+            input=self.input, prefix=self.prefixes, predicate=self.predicates
+        )
+        validation_msdf = read_sssom_table(self.validation_file)
+
+        # Drop empty columns since read_sssom_table drops them by default.
+        filtered_df = filtered_msdf.df
+        filtered_df.replace("", np.NAN, inplace=True)
+        filtered_df.dropna(how="all", axis=1, inplace=True)
+        filtered_df.fillna("", inplace=True)
```
matentzn, line 35: It is not very pretty that all of the above is needed to create a valid instance of a data frame.. better to use a table where this is not necessary?

tests/test_merge.py
```diff
@@ -21,8 +21,8 @@ def setUp(self) -> None:
     def test_merge_multiple_inputs(self):
         """Test merging of multiple msdfs."""
         merged_msdf = merge_msdf(*self.msdfs)
-        self.assertEqual(123, len(merged_msdf.df))
+        self.assertEqual(116, len(merged_msdf.df))
```
matentzn, line 24: Again this, which change in the code made this necessary?

tests/test_config.yaml
hrshdhgd: All of these changes were made in the previous PR (Validation) since many of the test files changed [as seen here](https://github.com/mapping-commons/sssom-py/pull/283/files#diff-a1794df9a43ac98ef5982e745aaf02c8cc7d2ff2d91878e34ee621d12b13d439). But it seemed like that never travelled to the main branch. When I pulled from main, those changes never made it to this branch. I'm guessing since I had these branches started in parallel, something got missed somewhere.

tests/test_filter.py
hrshdhgd: I agree. Pandas' `assert_frame_equal` test is acting funny and finding differences is exact same data frames. I implemented a different test for now.

sssom/io.py
hrshdhgd: Absolutely! Good catch, I've updated it.

sssom/io.py
```diff
@@ -228,3 +233,131 @@ def extract_iri(input, prefix_map) -> list:
...
+        fn = inputs[n - 1]
+        msdf = parse_sssom_table(fn)
+        df = msdf.df
+        # df = parse(fn)
+        globals()[f"df{n}"] = df
+        tn = re.sub("[.].*", "", Path(fn).stem).lower()
+        globals()[tn] = df
+        n += 1
+
+    new_df = sqldf(query)
```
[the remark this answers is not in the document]
matentzn: Hm. I hope you dont mind if I dig one layer deeper, because I really do not understand it.. So the SQL query we pass to pandsql has some kind of FROM clause that knows all by itself to look at globals()[TABLENAME]? In your select clause I see `SELECT * FROM df` - how does know that it should look at input1, input2 etc? Have you tried to run this with two inputs? I am a bit inclined to change the method to only allow a single input..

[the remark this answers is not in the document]
hrshdhgd: No, `pandasql` is a separate library: https://pypi.org/project/pandasql/

sssom/parsers.py
```diff
@@ -23,874 +23,874 @@
 from sssom_schema import Mapping, MappingSet
 
 from sssom.constants import (
-    CONFIDENCE,
```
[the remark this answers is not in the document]
matentzn: I would prefer we not do that.. I feel if we want to do this, we have to do it in a separate PR. Like this I do not know what is going in in any of the open PRs..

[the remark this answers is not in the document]
hrshdhgd: I agree , sorry about that.

hrshdhgd:
https://github.com/mapping-commons/sssom-py/pull/284/commits/51a774a3a288a4c95776a4f71654892757f7bf2a commit is because tox was complaining about indentation. Sorry, I couldn't help it. fox's linting did this.

matentzn (approved):

matentzn:
Awesome, loving it :)

PR #292, opened 2022-07-18
Add new command: annotate

### LINKED ISSUE #280: New method sssom annotate
To integrate SSSOM more into automated pipelines, it would be super cool to be able to annotate a mapping sets from the command line, especially to support dynamic versioning.

```
sssom annotate --mapping_set_id http://w3id.org/my/mapping.sssom.tsv --mapping_set_version 2021-01-01
```

The idea would be to add a dynamic decorator to click that generates long option (no short ones) from the SSSOM spec dynamically, and then allows setting their values. See something like https://stackoverflow.com/questions/50499340/specify-options-and-arguments-dynamically.

We only care about MappingSet metadata, not mapping metadata!

hrshdhgd:
Fixes #280

### DIFF
--- a/sssom/io.py
+++ b/sssom/io.py
@@ -365,11 +365,15 @@ def filter_file(input: str, output: TextIO, **kwargs) -> MappingSetDataFrame:
     return run_sql_query(query=query, inputs=[input], output=output)
 
 
-def annotate_file(input: str, output: TextIO, **kwargs) -> MappingSetDataFrame:
+def annotate_file(
+    input: str, output: TextIO, replace_multivalued: bool = False, **kwargs
+) -> MappingSetDataFrame:
     """Annotate a file i.e. add custom metadata to the mapping set.
 
     :param input: SSSOM tsv file to be queried over.
     :param output: Output location.
+    :param replace_multivalued: Multivalued slots should be
+        replaced or not, defaults to False
     :param **kwargs: Options provided by user
         which are added to the metadata (e.g.: --mapping_set_id http://example.org/abcd)
     :return: Annotated MappingSetDataFrame object.
@@ -377,6 +381,6 @@ def annotate_file(input: str, output: TextIO, **kwargs) -> MappingSetDataFrame:
     params = {k: v for k, v in kwargs.items() if v}
     are_params_slots(params)
     input_msdf = parse_sssom_table(input)
-    msdf = augment_metadata(input_msdf, params)
+    msdf = augment_metadata(input_msdf, params, replace_multivalued)
     write_table(msdf, output)
     return msdf
--- a/sssom/io.py
+++ b/sssom/io.py
@@ -376,6 +376,11 @@ def annotate_file(input: str, output: TextIO, **kwargs) -> MappingSetDataFrame:
     :return: Annotated MappingSetDataFrame object.
     """
     params = {k: v for k, v in kwargs.items() if v}
+
+    empty_params = {k: v for k, v in kwargs.items() if v is None or v == ""}
+    if len(empty_params) > 0:
+        logging.info(f"Parameters: {empty_params.keys()} has(ve) no value.")
+
     legit_params = all(p in MAPPING_SET_SLOTS for p in params.keys())
     if not legit_params:
         invalids = [p for p in params if p not in MAPPING_SET_SLOTS]
--- a/sssom/io.py
+++ b/sssom/io.py
@@ -29,6 +29,7 @@ from .parsers import get_parsing_function, parse_sssom_table, split_dataframe
 from .typehints import Metadata
 from .util import (
     MappingSetDataFrame,
+    augment_metadata,
     is_curie,
     is_iri,
     raise_for_bad_path,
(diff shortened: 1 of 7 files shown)


sssom/io.py
```diff
@@ -361,3 +362,32 @@ def filter_file(input: str, output: TextIO, **kwargs) -> MappingSetDataFrame:
...
+        invalids = [p for p in params if p not in MAPPING_SET_SLOTS]
+        raise ValueError(
+            f"The params are invalid: {invalids}. Should be any of the following: {MAPPING_SET_SLOTS}"
+        )
+
+    input_msdf = parse_sssom_table(input)
+    if input_msdf.metadata:
+        for k, v in params.items():
+            if len(v) <= 1:
+                input_msdf.metadata[k] = v[0]
```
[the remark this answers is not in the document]
matentzn: i guess this case is filtered by the row 377..

sssom/io.py
```diff
@@ -361,3 +362,32 @@ def filter_file(input: str, output: TextIO, **kwargs) -> MappingSetDataFrame:
...
+    :param input: DataFrame to be queried over.
+    :param output: Output location.
+    :param **kwargs: Options provided by user
+        which are added to the metadata (e.g.: --mapping_set_id http://example.org/abcd)
+    :raises ValueError: If parameter provided is invalid.
+    :return: Annotated MappingSetDataFrame object.
+    """
+    params = {k: v for k, v in kwargs.items() if v}
+    legit_params = all(p in MAPPING_SET_SLOTS for p in params.keys())
+    if not legit_params:
```
matentzn, line 379: can we refactor this and all the rest to get an "augment_metadata()" method you can put into utils.py that takes as an input an msdf, and takes back an msdf where the metadata is augmented with the input, and has a sanitised parameter for the stuff that is supposed to be annotated (ideally just a "Metadata" object?) Something like

```
augment_metadata(msdf, meta: Metadata):
```
where meta is more or less your `params` object? this makes this method also usable by users of the python API.

matentzn (dismissed):

matentzn (commented):
Sorry meant to comment before, not Approve, just make sure you read the comments. Great job :)

sssom/io.py
```diff
@@ -361,3 +362,32 @@ def filter_file(input: str, output: TextIO, **kwargs) -> MappingSetDataFrame:
         if multiple_params and idx != len(params):
             query += " AND ("
     return run_sql_query(query=query, inputs=[input], output=output)
+
+
+def annotate_file(input: str, output: TextIO, **kwargs) -> MappingSetDataFrame:
+    """Annotate a file i.e. add custom metadata to the mapping set.
+
+    :param input: DataFrame to be queried over.
```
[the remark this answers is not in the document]
hrshdhgd: Done

sssom/io.py
```diff
@@ -361,3 +362,32 @@ def filter_file(input: str, output: TextIO, **kwargs) -> MappingSetDataFrame:
...
+    :param output: Output location.
+    :param **kwargs: Options provided by user
+        which are added to the metadata (e.g.: --mapping_set_id http://example.org/abcd)
+    :raises ValueError: If parameter provided is invalid.
+    :return: Annotated MappingSetDataFrame object.
+    """
+    params = {k: v for k, v in kwargs.items() if v}
```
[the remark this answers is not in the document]
hrshdhgd: The dictionary comprehension step may get a little wordy with an `if-else` statement rather than just an `if`. I can do it in another step.

sssom/io.py
hrshdhgd: Done!

sssom/util.py
```diff
@@ -1250,3 +1250,19 @@ def get_all_prefixes(msdf: MappingSetDataFrame) -> list:
...
+    :param msdf: MappingSetDataFrame (MSDF) object.
+    :param meta: Dictionary that needs to be added/updated to the metadata of the MSDF.
+    :return: MSDF with updated metadata.
+    """
+    if msdf.metadata:
+        for k, v in meta.items():
+            if len(v) == 1:
+                msdf.metadata[k] = v[0]
+            else:
+                msdf.metadata[k] = list(v)
```
matentzn, line 1267: The problem with this is that if have have to creators in the metadata and want to add a third one.. it wont work. Would it make sense to make the method strictly additive for multi valued metadata? Maybe not. Do you have an opinion on this? Maybe it is correct the way it is..

hrshdhgd: Good catch! We will have to think about it. When is it additive and when is it replacement? Should there be a flag?

hrshdhgd: So what I have implemented now is if a `slot` to be annotated is `multivalued` , add new values to the list. But what if the entire `slot` value needs to be changed? How does the user indicate that?

matentzn (approved):
This looks great! THANK you! The only tiny thing that is slightly unsatisfactory, is the fact that the sanitation of the kwars happens during `annotate_file`, but nothing keeps an API user to pass a broken "meta" dict into `augment_metadata` method.

sssom/util.py
matentzn: What about a boolean flag `--replace-multivalued` which defaults to false, and if it is true, it will remove any existing values from a multivalued value?

hrshdhgd: But what if the users want to retain some of the contents of the list and add new stuff? They'll have to type the ones they want again. If that's the case, the first solution would have worked and was a simple one too.

hrshdhgd: For now I implemented `replace_multivalued`. If we can think of a cleaner way to do this, we can always go back and improve the code.

matentzn: I think this is fine. good one :)

matentzn (approved):

PR #293, opened 2022-08-01
Filter prefixes

hrshdhgd:
- [x] Added a new function `filter_prefixes`: This filters the `msdf.df` to contain only those mappings with prefixes that are present in a list provided by the user.
 - [x] Add class method `remove_mappings()` to MSDF class. Addresses one checkbox in https://github.com/monarch-initiative/mondo-ingest/issues/49

### DIFF
--- a/sssom/cli.py
+++ b/sssom/cli.py
@@ -675,5 +675,26 @@ def annotate(input: str, output: TextIO, replace_multivalued: bool, **kwargs):
     )
 
 
+@main.command()
+@input_argument
+@click.option(
+    "--remove-map",
+    type=click.Path(),
+    help="Mapping file path that needs to be removed from input.",
+)
+@output_option
+def remove(input: str, output: TextIO, remove_map: str):
+    """Remove mappings from an input mapping.
+
+    :param input: Input SSSOM tsv file.
+    :param output: Output path.
+    :param remove_map: Mapping to be removed.
+    """
+    input_msdf = parse_sssom_table(input)
+    remove_msdf = parse_sssom_table(remove_map)
+    input_msdf.remove_mappings(remove_msdf)
+    write_table(input_msdf, output)
+
+
 if __name__ == "__main__":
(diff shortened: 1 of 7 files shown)


matentzn (dismissed):
I approve this, but if you really want to make me happy you can add a CLI method

```
remove(msdf1, msdf2)
  msdf1.remove_mappings(msdf2)
  write(msdf1)
```

hrshdhgd:
Your wish is my command! 
```
sssom remove abcd.sssom.tsv --remove-map def.sssom.tsv -o abde.sssom.tsv` 
```
ashould work.

sssom/cli.py
```diff
@@ -675,5 +675,26 @@ def annotate(input: str, output: TextIO, replace_multivalued: bool, **kwargs):
     )
 
 
+@main.command()
+@input_argument
+@click.option(
+    "--remove-map",
```
matentzn, line 681: This is great, we can later consider making this multivalued (not now). Thank you!

sssom/io.py
```diff
@@ -360,6 +360,8 @@ def filter_file(input: str, output: TextIO, **kwargs) -> MappingSetDataFrame:
             for exp in v[1:]:
                 query += " OR "
                 query += k + " LIKE '" + exp + "') "
+        else:
+            query += ") "
```
matentzn, line 364: this is new, I assume there was no test for the `else` case?

matentzn (approved):
Excellent, left a few comments but nothing that blocks merging. THANKS

sssom/io.py
hrshdhgd: This was missing and became apparent when only 1 value to a param was given for e.g. `--subject_id ABCD:1234` instead of multiple `--subject_id ABCD:1234 --subject_id DEFG:4567`. The missing `)` that's added in the code above threw an error. There was no test to check this.

PR #308, opened 2022-08-19
Fixes #305 and  #306

### LINKED ISSUE #305: clean_prefix_map cleans out prefixes used in metadata
MappingSetDataFrame.clean_prefix_map()  has two problems:

1. Does not take into account `metadata` (YAML header)
2. Does not include `semapv` in the prefix map, even though it is used for `mapping_justification`. I think the mistake is that we only look at KEY_FEATURES. instead, we should look at all columns which have type EntityReference

### LINKED ISSUE #306: Redundant empty prefix checked needlessly by validate
When validating the file below:

```
sssom validate mp_hp_owt_impc.sssom.tsv
```

I get:

```
check_all_prefixes_in_curie_map
    raise ValidationError(
jsonschema.exceptions.ValidationError: The prefixes in [''] are missing from 'curie_map'.
```

This can be fixed by adding another orcid into the reviewer column. Something is wrong with the way the prefixes are extracted from the table?

```
# creator_id: orcid:0000-0002-2232-0967
# curie_map:
#   HP: http://purl.obolibrary.org/obo/HP_
#   MP: http://purl.obolibrary.org/obo/MP_
#   orcid: https://orcid.org/
#   obo: http://purl.obolibrary.org/obo/
# license: https://creativecommons.org/publicdomain/zero/1.0/
# mapping_provider: https://www.mousephenotype.org
# mapping_set_description: 'The IMPC Mouse Morphology Mappings: Organ Weight Test'
# mapping_set_id: mp_hp_owt_impc
subject_id	subject_label	predicate_id	object_id	object_label	mapping_justification	reviewer_id
MP:0001259	abnormal body weight	skos:exactMatch	HP:0004323	Abnormality of body weight	semapv:LexicalMatching	orcid:0000-0002-7356-1779
MP:0002707	abnormal kidney weight	skos:closeMatch	HP:0000077	Abnormality of the kidney	semapv:LogicalReasoning	
```

hrshdhgd:
Issues: 
 - When a SSSOM tsv column that is supposed to have a CURIE is <empty> , `validate` throws an error saying the file is't valid. Empty values are allowed in a mapping file and this minor bug fix takes care of it. Fixes #306 
 - `Clean_prefix_map()` makes sure all dataframe prefixes are present in the `prefix_map`. This needs refactoring to `clean_prefix_map_and_metadata()` where it checks BOTH `prefix_map` and `metadata`. Fixes #305

### DIFF
--- a/sssom/io.py
+++ b/sssom/io.py
@@ -109,7 +109,7 @@ def parse_file(
     #     )
     if clean_prefixes:
         # We do this because we got a lot of prefixes from the default SSSOM prefixes!
-        doc.clean_prefix_map()
+        doc.clean_prefix_map_and_metadata()
     write_table(doc, output, embedded_mode)
 
 
--- a/sssom/io.py
+++ b/sssom/io.py
@@ -109,7 +109,7 @@ def parse_file(
     #     )
     if clean_prefixes:
         # We do this because we got a lot of prefixes from the default SSSOM prefixes!
-        doc.clean_prefix_map_and_metadata()
+        doc.clean_prefix_map()
     write_table(doc, output, embedded_mode)
 
 
--- a/sssom/util.py
+++ b/sssom/util.py
@@ -1097,6 +1097,7 @@ def get_prefixes_used_in_table(df: pd.DataFrame) -> List[str]:
                     prefixes.append(get_prefix_from_curie(v))
     return list(set(prefixes))
 
+
 def get_prefixes_used_in_metadata(meta: MetadataType) -> List[str]:
     """Get a list of prefixes used in CURIEs in the metadata."""
     prefixes = []
@@ -1104,8 +1105,7 @@ def get_prefixes_used_in_metadata(meta: MetadataType) -> List[str]:
         for v in meta.values:
             prefixes.append(get_prefix_from_curie(v))
     return prefixes
-        
-        
+
 
 def filter_out_prefixes(
     df: pd.DataFrame, filter_prefixes: List[str], features: list = KEY_FEATURES
--- a/sssom/util.py
+++ b/sssom/util.py
@@ -1379,6 +1379,7 @@ def get_all_prefixes(msdf: MappingSetDataFrame) -> list:
                         [
                             get_prefix_from_curie(s)
                             for s in list(set(msdf.df[slot].to_list()))  # type: ignore
+                            if get_prefix_from_curie(s) != ""
                         ]
                     )
                 )
--- a/sssom/util.py
+++ b/sssom/util.py
@@ -143,17 +143,16 @@ class MappingSetDataFrame:
             description += self.df.tail().to_string() + "\n"
         return description
 
-    def clean_prefix_map_and_metadata(self) -> None:
+    def clean_prefix_map(self) -> None:
         """Remove unused prefixes from the internal prefix map based on the internal dataframe."""
         prefixes_in_map = get_prefixes_used_in_table(self.df)
         new_prefixes: PrefixMap = dict()
-        new_meta: MetadataType = dict()
         missing_prefixes = []
         for prefix in prefixes_in_map:
             if prefix in self.prefix_map:
                 new_prefixes[prefix] = self.prefix_map[prefix]
             elif prefix in self.metadata:  # type: ignore
-                new_meta[prefix] = self.metadata[prefix]  # type: ignore
+                continue
             else:
                 logging.warning(
                     f"{prefix} is used in the dataframe but neither exists in prefix map nor metadata"
@@ -162,7 +161,6 @@ class MappingSetDataFrame:
         if missing_prefixes:
             self.df = filter_out_prefixes(self.df, missing_prefixes)
         self.prefix_map = new_prefixes
-        self.metadata = new_meta
 
     def remove_mappings(self, msdf: "MappingSetDataFrame"):
         """Remove mappings in right msdf from left msdf.
@@ -184,7 +182,7 @@ class MappingSetDataFrame:
         )
 
         self.df = self.df[self.df.columns.drop(list(self.df.filter(regex=r"_2")))]
-        self.clean_prefix_map_and_metadata()
+        self.clean_prefix_map()
 
 
 @dataclass
--- a/sssom/util.py
+++ b/sssom/util.py
@@ -146,7 +146,7 @@ class MappingSetDataFrame:
     def clean_prefix_map(self) -> None:
         """Remove unused prefixes from the internal prefix map based on the internal dataframe."""
         prefixes_in_map = get_prefixes_used_in_table(self.df)
-        prefixes_in_metadata = get_prefixes_used_in_metadata(self.metadata)
+        prefixes_in_metadata = get_prefixes_used_in_metadata(self.metadata)  # type: ignore
         new_prefixes: PrefixMap = dict()
         missing_prefixes = []
         for prefix in prefixes_in_map:
--- a/sssom/util.py
+++ b/sssom/util.py
@@ -73,7 +73,12 @@ from .constants import (
     SUBJECT_LABEL,
     SUBJECT_SOURCE,
 )
-from .context import SSSOM_URI_PREFIX, get_default_metadata, get_jsonld_context
+from .context import (
+    SSSOM_BUILT_IN_PREFIXES,
+    SSSOM_URI_PREFIX,
+    get_default_metadata,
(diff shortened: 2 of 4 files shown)


matentzn (approved):

sssom/validators.py
```diff
@@ -83,7 +83,7 @@ def check_all_prefixes_in_curie_map(msdf: MappingSetDataFrame) -> None:
 
     missing_prefixes = []
     for pref in prefixes:
-        if pref not in list(msdf.prefix_map.keys()):
```
matentzn, line 86: Out of curiosity (not a blocker for merging) but shouldn't we prefent an empty string being added to the `prefixes` variable in the first place?

hrshdhgd: You're spot on! Was missing an `if` in `util.py`. [Added it now](https://github.com/mapping-commons/sssom-py/pull/308/commits/6fd501fb8b5be197d27a9c467400a8275364bcf0)!

sssom/util.py
```diff
@@ -143,22 +143,26 @@ def __str__(self) -> str:  # noqa:D105
             description += self.df.tail().to_string() + "\n"
         return description
 
-    def clean_prefix_map(self) -> None:
+    def clean_prefix_map_and_metadata(self) -> None:
```
matentzn, line 146: Sorry I think I explained this wrong. The function should only clean the prefix_map, but _take into account any prefixes used in the metadata_. So I expected something like:

```
prefixes_in_map = get_prefixes_used_in_table(self.df)
prefixes_in_metadata = get_prefixes_used_in_metadata(self.metadata)
prefixes_in_msdf = prefixes_in_map
prefixes_in_msdf.extent(prefixes_in_metadata)
```

Or some such? So: make sure you don't remove prefixes from the prefix map that are needed for the metadata!

sssom/io.py
```diff
@@ -109,7 +109,7 @@ def parse_file(
     #     )
     if clean_prefixes:
         # We do this because we got a lot of prefixes from the default SSSOM prefixes!
-        doc.clean_prefix_map()
+        doc.clean_prefix_map_and_metadata()
```
matentzn, line 112: Maybe it wont be necessary to rename, read next comment

matentzn (changes requested):

sssom/util.py
hrshdhgd: Does the new solution make sense?

matentzn: This is what I am thinking with:

```
prefixes_in_metadata = get_prefixes_used_in_metadata(self.metadata)
```

This is example metadata:

```
#mapping_set_id: MGI_KidsFirst_MP_HPO
#mapping_set_title: Mapping of MP terms to HPO terms from KidsFirst list
#mapping_set_description: "HPO terms were provided by the KidsFirst team from the set of terms used for patients in their cohorts. Initially the mapping date was not tracked so these were given the date of when we recieved the file (5/27/21). The set of HPO terms was updated on 5/17/212. Use of skos mappings terms: exact match: MP term is the same as the HPO term, ignoring that the HP term applies only to humans; broad match: MP term covers more than the HPO term; narrow match: MP term is more restircted than the HPO term, ignoring the species restriction; related match: MP term and HPO term share some similarity but would not have a parent-child relation in a shared ontology"
#creator_id: orcid:0000-0003-4606-0597|orcid:0000-0002-6490-7723|ror:021sy4w91|wikidata:Q1951035
#license: https://creativecommons.org/licenses/by/4.0/
#object_source: obo:hp
#subject_source: obo:mp
```

The prefixes are not `object_source` or `creator_id` as your current code seems to suggest. The prefixes are `obo:` as used in the `object_source` field, and `orcid:` as used in the `creator_id` field. So to get all the prefixes in the metadata, you have to loop through the values of the metadata (be they literals or lists!) and make sure to extract all the prefixes used in there - not in the keys!

sssom/util.py
```diff
@@ -146,14 +146,18 @@ def __str__(self) -> str:  # noqa:D105
     def clean_prefix_map(self) -> None:
         """Remove unused prefixes from the internal prefix map based on the internal dataframe."""
         prefixes_in_map = get_prefixes_used_in_table(self.df)
```
matentzn, line 148: I know this is not part of the PR, but it is part of the confusion here:

1. rename `prefixes_in_map` to `prefixes_in_table`. This variable has nothing to do with the map!

sssom/util.py
```diff
@@ -146,14 +146,18 @@ def __str__(self) -> str:  # noqa:D105
...
+        if self.metadata:
+            prefixes_in_metadata = get_prefixes_used_in_metadata(self.metadata)
         new_prefixes: PrefixMap = dict()
         missing_prefixes = []
         for prefix in prefixes_in_map:
```
matentzn, line 153:
```suggestion
   all_prefixes = prefixes_in_metadata + prefixes_in_table
        for prefix in all_prefixes:
```

matentzn: This is just for illustration, you can just do it on your own branch

sssom/util.py
```diff
@@ -146,14 +146,18 @@ def __str__(self) -> str:  # noqa:D105
...
             if prefix in self.prefix_map:
                 new_prefixes[prefix] = self.prefix_map[prefix]
+            elif len(prefixes_in_metadata) > 0 and prefix in prefixes_in_metadata:
+                continue
```
matentzn, line 157:
```suggestion
```

matentzn: Delete rows 156 and 157 - no need.

sssom/util.py
```diff
@@ -146,14 +146,18 @@ def __str__(self) -> str:  # noqa:D105
...
             else:
                 logging.warning(
-                    f"{prefix} is used in the data frame but does not exist in prefix map"
+                    f"{prefix} is used in the dataframe but neither exists in prefix map nor metadata"
```
matentzn, line 160:
```suggestion
                    f"{prefix} is used in the SSSOM mapping set but it does not exist in the prefix map"
```

sssom/util.py
```diff
@@ -1088,12 +1092,22 @@ def get_prefixes_used_in_table(df: pd.DataFrame) -> List[str]:
...
+                    prefixes.append(get_prefix_from_curie(v))
     return list(set(prefixes))
 
 
+def get_prefixes_used_in_metadata(meta: MetadataType) -> List[str]:
+    """Get a list of prefixes used in CURIEs in the metadata."""
+    prefixes = []
+    if meta:
+        for v in meta.values():
+            prefixes.append(get_prefix_from_curie(str(v)))
```
matentzn, line 1107: What if the value is a list, which it can be?

sssom/util.py
```diff
@@ -1088,12 +1092,22 @@ def get_prefixes_used_in_table(df: pd.DataFrame) -> List[str]:
...
     prefixes = []
     if not df.empty:
-        for col in KEY_FEATURES:
-            for v in df[col].values:
-                prefixes.append(get_prefix_from_curie(v))
+        for col in ENTITY_REFERENCE_SLOTS:
+            if col in df.columns:
+                for v in df[col].values:
+                    prefixes.append(get_prefix_from_curie(v))
     return list(set(prefixes))
```
matentzn, line 1099: prefixes should _always_ contain `semapv` and `sssom` prefixes.

matentzn (changes requested):
I explained this badly. You can make use of prefixes in two places:
1. The data frame
2. The metadata

The goal of the clean_prefix_map method is to get rid of prefixes not used at all in either the data frame nor the metadata.

sssom/util.py
hrshdhgd: I thought of doing this initially : `all_prefixes = prefixes_in_metadata + prefixes_in_table` but then checkout line 165:
```
self.prefix_map = new_prefixes
```
The reason I keep them separate is because that way I'll know which one belongs to the `prefix_map` and which one to the `metadata`. The differentiation is needed (I think).

sssom/util.py
hrshdhgd: Should I add "semapv" to [SSSOM_BUILT_IN_PREFIXES](https://github.com/mapping-commons/sssom-py/blob/5490672f3bb8853f0a9a6d01a223a7f476a37f1a/sssom/context.py#L20)?

matentzn: Great!

matentzn (approved):
Awesome thats it! THANK YOU! and sorry again for the chaos.

PR #312, opened 2022-09-14
Fix obographs parser

matentzn:
obographs parser was not checking edges and equivalence nodes.


hrshdhgd (dismissed):
Makes sense to me!

matentzn:
Made some changes, if you are ok with them, just merge it. Thank you @hrshdhgd :)

hrshdhgd (approved):

PR #318, opened 2022-09-25
Delete all validation files since they may be old

### LINKED ISSUE #317: Clean up tests/data/validation_data
There are quite a few JSON files in there which were from an ancient SSSOM version.

We should

1. regenerate all the data in the tests/data/validation_data using the up-to-date sssom-py
2. check that all tests still pass

hrshdhgd:
Fixes #317


matentzn:
All of them? Oh wow, we must have not tested roundtripping for ages!

See https://github.com/mapping-commons/sssom-py/blob/validate-data-files/tests/test_conversion.py#L68

Can you reintroduce some of these checks? I felt quite protected having them there!

hrshdhgd:
Ok, so when I put a created files in the `validate_data` folder and run the validation check, it still fails (literally the same file copy-pasted). I dug deeper and the root cause is in the `filecmp.cmp()` function.
```
    s1 = _sig(os.stat(f1))
    s2 = _sig(os.stat(f2))
    if s1[0] != stat.S_IFREG or s2[0] != stat.S_IFREG:
        return False
    if shallow and s1 == s2:
        return True
    if s1[1] != s2[1]:
        return False
```
Apparently s1 is not == s2. 
```
s1 = (32768, 98892, 1664213293.0535548)
s2 = (32768, 98892, 1664212433.827459)

os.stat(f1) = os.stat_result(st_mode=33152, st_ino=16032794, st_dev=16777225, st_nlink=1, st_uid=502, st_gid=20, st_size=98892, st_atime=1664213294, st_mtime=1664213293, st_ctime=1664213293)

os.stat(f2) = os.stat_result(st_mode=33152, st_ino=49568904, st_dev=16777225, st_nlink=1, st_uid=502, st_gid=20, st_size=98892, st_atime=1664213022, st_mtime=1664212433, st_ctime=1664213021)
```
So the `st_atime` does not match and hence the files are determined to be not equal. I'm assuming this is the creation/modified datetime(?). Would this be the correct way of checking files? Should we just write a different test to test these files?

matentzn:
@hrshdhgd can you update this PR?

PR #323, opened 2022-10-21
`SchemaView` object should be instantiated when needed & not globally.

### LINKED ISSUE #322: constants.py should not invoke schemaview
This is called *when importing sssom-py:

https://github.com/mapping-commons/sssom-py/blob/5490672f3bb8853f0a9a6d01a223a7f476a37f1a/sssom/constants.py#L12-L17

in general it's not a good idea to have logic executed when a package is imported. logic should only be done on specific invocation of that logic.

in this case it's particularly problematic as this invokes a network call...

hrshdhgd:
Fixes #322 

This is because it invokes a network call when `sssom-py` is imported. This is unnecessary.


sssom/constants.py
```diff
@@ -3,23 +3,12 @@
 import pathlib
 from enum import Enum
 
-import pkg_resources
-from linkml_runtime.utils.schema_as_dict import schema_as_dict
-from linkml_runtime.utils.schemaview import SchemaView
+from sssom.schema import SCHEMA_VIEW
```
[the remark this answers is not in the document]
hrshdhgd: lol! you're correct... moved it to `schema.yaml`

matentzn (changes requested):

sssom/cli.py
```diff
@@ -628,7 +627,7 @@ def decorator(f):
 @main.command()
 @input_argument
 @output_option
-@dynamically_generate_sssom_options(MAPPING_SLOTS)
+@dynamically_generate_sssom_options(SSSOMSchemaView().mapping_slots)
```
matentzn, line 630: I wonder if this will defeat the purpose of the refactoring. I should have thought of this before. 

I am wondering now if we need to find a way to skip network access in SchemaView()?

sssom/constants.py
```diff
@@ -177,3 +158,19 @@ class SchemaValidationType(str, Enum):
...
+    """
+    SchemaView class from linkml which is instantiated when necessary.
+
+    Reason for this: https://github.com/mapping-commons/sssom-py/issues/322
+    Implemented via PR: https://github.com/mapping-commons/sssom-py/pull/323
+    """
+
+    entity_reference = "EntityReference"
+    yaml = pkg_resources.resource_filename("sssom_schema", "schema/sssom_schema.yaml")
+    view = SchemaView(yaml)
```
matentzn, line 173: I am not sure, but you may have to use init() to avoid this code being run whenever the class is referenced: 

https://stackoverflow.com/questions/9056957/correct-way-to-define-class-variables-in-python

But not 100% sure.

sssom/util.py
```diff
@@ -1040,6 +1036,18 @@ class NoCURIEException(ValueError):
 
 
 CURIE_RE = re.compile(r"[A-Za-z0-9_.]+[:][A-Za-z0-9_]")
+schema_view_object = SSSOMSchemaView()
```
matentzn, line 1039: this is again in the global context I think.. wont this be invoked any time anyone imports anything from utils package?

sssom/writers.py
```diff
@@ -284,7 +284,7 @@ def to_rdf_graph(msdf: MappingSetDataFrame) -> Graph:
     # os.remove("sssom.ttl")  # remove the intermediate file.
     graph = rdflib_dumper.as_rdf_graph(
         element=doc.mapping_set,
-        schemaview=SchemaView(SCHEMA_YAML),
+        schemaview=SchemaView(SSSOMSchemaView().yaml),
```
matentzn, line 287: This is a bit weird... :D

matentzn (changes requested):

sssom/util.py
hrshdhgd: I will have to repeat this code in 4 different lines in the same file. Seems messy.

sssom/constants.py
```diff
@@ -177,3 +159,55 @@ class SchemaValidationType(str, Enum):
...
+    entity_reference = "EntityReference"
+    yaml = pkg_resources.resource_filename("sssom_schema", "schema/sssom_schema.yaml")
+    _view = None
+    _dict = None
+
+    @property
+    def view(self) -> SchemaView:
+        """Return SchemaView object."""
+        if self._view is None:
+            self._view = SchemaView(self.yaml)
```
matentzn, line 181: smart, so, whenever self.view is accessed as a variable, this function is executed, effectively deferring the SchemaView() instantiation to the first time its executed? Cool!

sssom/util.py
```diff
@@ -832,7 +828,7 @@ def inject_metadata_into_df(msdf: MappingSetDataFrame) -> MappingSetDataFrame:
     :return: MappingSetDataFrame with metadata as columns
     """
     # TODO Check if 'k' is a valid 'slot' for 'mapping' [sssom.yaml]
-    with open(SCHEMA_YAML) as file:
+    with open(SSSOMSchemaView().yaml) as file:
         schema = yaml.safe_load(file)
     slots = schema["classes"]["mapping"]["slots"]
```
matentzn, line 833: seems redundant now.. Shouldnt this simply be part of the new `SSSOMSchemaView` class object?

matentzn: Why would you load the yaml again, rather than using `SSSOMSchemaView.dict`?

tests/test_schemaview.py
```diff
@@ -0,0 +1,16 @@
...
+class TestSort(unittest.TestCase):
+    """A test case for sorting msdf columns."""
+
+    def setUp(self) -> None:
+        """Test up the test cases with the third basic example."""
+        self.sv = SSSOMSchemaView()
+
+    def test_slots(self) -> None:
+        """Test slots."""
+        self.assertEqual(len(self.sv.mapping_slots), 39)
```
matentzn, line 16: probably better test `self.assertTrue(len(self.sv.mapping_slots)>=39)` because changes to the metadata model will break this test every time?

tox.ini
```diff
@@ -52,7 +52,7 @@ description = Run the flake8 code quality checker.
 [testenv:mypy]
 deps = mypy
 skip_install = true
-commands = mypy --install-types --non-interactive --ignore-missing-imports sssom/ setup.py
+commands = mypy --install-types --non-interactive --ignore-missing-imports --implicit-optional sssom/ setup.py
```
matentzn, line 55: What does that do?

tests/test_sort.py
```diff
@@ -19,6 +19,8 @@ def test_sort(self):
         """Test sorting of columns."""
         new_df = sort_df_rows_columns(self.msdf.df)
         column_sequence = [
-            col for col in SCHEMA_DICT["slots"].keys() if col in new_df.columns
+            col
+            for col in SSSOMSchemaView().dict["slots"].keys()
```
matentzn, line 23: so weird SchemaView() has no "get_slots()".

tox.ini
hrshdhgd: In its absence `mypy` throws a bunch of errors.

sssom/util.py
hrshdhgd: Good catch!

tests/test_schemaview.py
hrshdhgd: Good call, changed!

tox.ini
matentzn: nothing you think is serious enough to be fixed though? I will trust you on this, just doing due diligence

hrshdhgd:
See #324

PR #324, opened 2022-11-09
Moving SchemaView() calls from global space to parts where necessary.

### LINKED ISSUE #322: constants.py should not invoke schemaview

hrshdhgd:
Followup on #323 . Fixes #322

### DIFF
--- a/sssom/constants.py
+++ b/sssom/constants.py
@@ -2,6 +2,7 @@
 
 import pathlib
 from enum import Enum
+from typing import List
 
 import pkg_resources
 from linkml_runtime.utils.schema_as_dict import schema_as_dict
@@ -14,11 +15,8 @@ HERE = pathlib.Path(__file__).parent.resolve()
 SCHEMA_YAML = pkg_resources.resource_filename(
     "sssom_schema", "schema/sssom_schema.yaml"
 )
-SCHEMA_VIEW = SchemaView(SCHEMA_YAML)
+
 # SCHEMA_VIEW = package_schemaview("sssom_schema")
-SCHEMA_DICT = schema_as_dict(SCHEMA_VIEW.schema)
-MAPPING_SLOTS = SCHEMA_DICT["classes"]["mapping"]["slots"]
-MAPPING_SET_SLOTS = SCHEMA_DICT["classes"]["mapping set"]["slots"]
 
 OWL_EQUIV_CLASS = "http://www.w3.org/2002/07/owl#equivalentClass"
 RDFS_SUBCLASS_OF = "http://www.w3.org/2000/01/rdf-schema#subClassOf"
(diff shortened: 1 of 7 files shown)


sssom/constants.py
```diff
@@ -177,3 +166,43 @@ class SchemaValidationType(str, Enum):
     SchemaValidationType.JsonSchema,
     SchemaValidationType.PrefixMapCompleteness,
 ]
+
+
+class SSSOMSchemaViewSingleton(object):
```
cmungall, line 171: I think this works but we have one too many layers here

SSSOMSchemaViewSingleton -> SSSOMSchemaView -> SchemaView

I think you could have just made SSSOMSchemaView a singleton

sssom/constants.py
```diff
@@ -177,3 +166,43 @@ class SchemaValidationType(str, Enum):
...
+    SchemaView class from linkml which is instantiated when necessary.
+
+    Reason for this: https://github.com/mapping-commons/sssom-py/issues/322
+    Implemented via PR: https://github.com/mapping-commons/sssom-py/pull/323
+    """
+
+    def __init__(self):
+        """Initialize class attributes."""
+        self.view = SchemaView(SCHEMA_YAML)
+        self.dict = schema_as_dict(self.view.schema)
```
cmungall, line 192: so it looks like you backed out the changes that used lazy loading of these

the whole point of having an abstraction over SchemaView was to allow deferred loading...

hrshdhgd: I was testing it with minimal code and it seemed to run tests fractionally faster. I've rolled it back to the lazy loading.

sssom/constants.py
hrshdhgd: You're right. I combined them.

sssom/constants.py
```diff
@@ -177,3 +166,59 @@ class SchemaValidationType(str, Enum):
...
+    def dict(self) -> dict:
+        """Return SchemaView as a dictionary."""
+        if self._dict is None:
+            self._dict = schema_as_dict(self.view.schema)
+        return self._dict
+
+    @property
+    def mapping_slots(self) -> List[str]:
+        """Return list of mapping slots."""
+        return self.dict["classes"]["mapping"]["slots"]
```
cmungall, line 205: just making a non-blocking note here. We will later refactor this, there is zero need for a dict or for doing dict lookups. the entire point of SV is to abstract away from this low level representation!

however, we now have a buffering layer with SSSOMSchemaView, so this is something we can do transparently later on!

matentzn:
lets go over this during next call. Some small qs but generally looks good!

sssom/constants.py
hrshdhgd: I think I know what you mean: So basically the above should be `return self.view.get_class("mapping").slots` , correct?

cmungall (approved):

tox.ini
```diff
@@ -15,7 +15,7 @@ envlist =
 commands =
     python -m pytest
 deps = 
-    linkml
+    linkml==1.3.13
```
matentzn, line 18: Make an issue to unpin with a link to the blocking issue on linkml repo

PR #329, opened 2022-12-14
clean_prefix_map() errors when SSSOM tsv invalid.

### LINKED ISSUE #328: `clean_prefix_map()` drops mappings whose CURIEs are not in `prefix_map`
This was a design decision made that needs to be revisited. Currently it only throws a warning but we want to throw an error so the user corrects it rather than proceed with an invalid SSSOM tsv file.

hrshdhgd:
Fixes #328

### DIFF
--- a/sssom/context.py
+++ b/sssom/context.py
@@ -17,7 +17,7 @@ from .typehints import Metadata, MetadataType, PrefixMap
 # EXTERNAL_CONTEXT_PATH = HERE / "sssom.external.context.jsonld"
 
 SSSOM_URI_PREFIX = "https://w3id.org/sssom/"
-SSSOM_BUILT_IN_PREFIXES = ["sssom", "owl", "rdf", "rdfs", "skos", "semapv"]
+SSSOM_BUILT_IN_PREFIXES = ("sssom", "owl", "rdf", "rdfs", "skos", "semapv")
 DEFAULT_MAPPING_SET_ID = f"{SSSOM_URI_PREFIX}mappings/{uuid.uuid4()}"
 DEFAULT_LICENSE = f"{SSSOM_URI_PREFIX}license/unspecified"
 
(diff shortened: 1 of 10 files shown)


matentzn (approved):
Ok.. Since this is restricted to the "clean" method, its fine.

sssom/parsers.py
```diff
@@ -862,8 +862,8 @@ def split_dataframe(
     """
     if msdf.df is None:
         raise RuntimeError
-    subject_prefixes = set(msdf.df[SUBJECT_ID].str.split(":", 1, expand=True)[0])
-    object_prefixes = set(msdf.df[OBJECT_ID].str.split(":", 1, expand=True)[0])
+    subject_prefixes = set(msdf.df[SUBJECT_ID].str.split(":", n=1, expand=True)[0])
+    object_prefixes = set(msdf.df[OBJECT_ID].str.split(":", n=1, expand=True)[0])
```
hrshdhgd, line 866: minor edit ... [explanation](https://stackoverflow.com/questions/73886976/facing-error-in-pandas-all-arguments-of-stringmethods-split)

sssom/context.py
```diff
@@ -17,7 +17,7 @@
 # EXTERNAL_CONTEXT_PATH = HERE / "sssom.external.context.jsonld"
 
 SSSOM_URI_PREFIX = "https://w3id.org/sssom/"
-SSSOM_BUILT_IN_PREFIXES = ["sssom", "owl", "rdf", "rdfs", "skos", "semapv"]
+SSSOM_BUILT_IN_PREFIXES = ("sssom", "owl", "rdf", "rdfs", "skos", "semapv")
```
hrshdhgd, line 20: Apparently this was getting overwritten after being assigned to `prefixes` in `utils.py`. So converting it into a tuple so it is immutable.

matentzn (approved):
Looks ok to me!

PR #330, opened 2022-12-19
Added sort while write_table

hrshdhgd:


sssom/writers.py
```diff
@@ -68,6 +69,7 @@ def write_table(
     if msdf.prefix_map is not None:
         meta[PREFIX_MAP_KEY] = msdf.prefix_map
 
+    msdf.df = sort_df_rows_columns(msdf.df)
```
matentzn, line 72: please add a flag to the method which is `sort = false` and only do the sort when explicitly changed.

hrshdhgd: Done

matentzn (approved):

PR #341, opened 2023-02-02
Check attribute exists using `hasattr()` and anchor `SQLAlchemy` version.

### LINKED ISSUE #340: SSSOM code broken
It's really frustrating to rely on this package. Sometimes it just breaks and then my builds that are relying on it don't work. This makes me want to make my own SSSOM package, which I won't have time to maintain either, but at least I will be able to know how it works...

Can someone help figure out what's going on in https://github.com/biopragmatics/biomappings/actions/runs/4075209444/jobs/7021308103?

hrshdhgd:
- ~SSSOMSchemaView.instance if SSSOMSchemaView.instance else SSSOMSchemaView()~ => SSSOMSchemaView.instance if hasattr(SSSOMSchemaView,"instance") else SSSOMSchemaView() fixes #340
 - `SQLAlchemy`'s latest version (v2.x.x) is causing errors in legacy code. Need to update code to be v2.x.x compliant. For now anchoring it to prior version v1.4.46
   - `SQLAlchemy` is a dependency of `pandasql` and [there is ticket for this.](https://github.com/yhat/pandasql/issues/102)
 - Removed `setup.py` from `flake8` check in `tox` due to [this issue experienced](https://github.com/PyCQA/bandit/discussions/725)
 - Fix bug mentioned in https://github.com/monarch-initiative/mondo-ingest/issues/129#issuecomment-1414535714

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -49,7 +49,7 @@ keywords =
 
 [options]
 install_requires =
-    sqlalchemy=1.4.46
+    sqlalchemy==1.4.46
     pyparsing
     bioregistry
     click
--- a/setup.cfg
+++ b/setup.cfg
@@ -49,6 +49,7 @@ keywords =
 
 [options]
 install_requires =
+    sqlalchemy=1.4.46
     pyparsing
     bioregistry
     click
--- a/setup.cfg
+++ b/setup.cfg
@@ -49,7 +49,7 @@ keywords =
 
 [options]
 install_requires =
-    sqlalchemy==1.4.46
+    sqlalchemy<2.0.0
     pyparsing
     bioregistry
     click
--- a/tests/test_collapse.py
+++ b/tests/test_collapse.py
@@ -44,7 +44,7 @@ class TestCollapse(unittest.TestCase):
     def test_filter(self):
         """Test the row count after filtering redundant rows."""
         df = filter_redundant_rows(self.df)
-        self.assertEqual(len(df), 91)
+        self.assertEqual(len(df), 92)
 
     def test_ptable(self):
         """Test the row count of the ptable export."""
--- a/tox.ini
+++ b/tox.ini
@@ -34,7 +34,7 @@ description = Run code formatters and linters.
 [testenv:flake8]
 skip_install = true
 commands =
-    flake8 sssom/ tests/ setup.py
+    flake8 sssom/ tests/ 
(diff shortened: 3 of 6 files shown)


tox.ini
```diff
@@ -34,7 +34,7 @@ description = Run code formatters and linters.
 [testenv:flake8]
 skip_install = true
 commands =
-    flake8 sssom/ tests/ setup.py
```
matentzn, line 37: ?

hrshdhgd: `tox` throws an error. I have linked the ticket/discussion where users have complained about it in the PR description

setup.cfg
```diff
@@ -49,6 +49,7 @@ keywords =
 
 [options]
 install_requires =
+    sqlalchemy==1.4.46
```
cmungall, line 52: let's not do this

hrshdhgd: replaced with `<2.0.0` Once I figure out a solution, this will be taken out.

tests/test_collapse.py
```diff
@@ -44,7 +44,7 @@ def test_collapse(self):
     def test_filter(self):
         """Test the row count after filtering redundant rows."""
         df = filter_redundant_rows(self.df)
-        self.assertEqual(len(df), 91)
+        self.assertEqual(len(df), 92)
```
matentzn, line 47: Is this expected?

setup.cfg
```diff
@@ -49,6 +49,7 @@ keywords =
 
 [options]
 install_requires =
+    sqlalchemy<2.0.0
```
matentzn, line 52: Can we combine merging this with a new issue to unpin, that contains the link to the issue that is is the reason why pinning is important?

tests/test_collapse.py
hrshdhgd: yes

setup.cfg
hrshdhgd: We could but that means Charlie will have to wait .... your call.

matentzn: I meant, we can merge your proposal now, but create a second issue to unpin so we don't forget!

hrshdhgd: Aah! gotcha! [Done!](https://github.com/mapping-commons/sssom-py/issues/343)

PR #345, opened 2023-02-23
Just removed pinning of SQLAlchemy

### LINKED ISSUE #343: Unpin `sqlalchemy` version
SQLAlchemy version 2.0.x is not compatible with implementations upto version 1.4.46 as seen in #341 (bullet 2)

Dependent on https://github.com/yhat/pandasql/issues/102#issuecomment-1413970825

hrshdhgd:
Fixes #343 

So now we can use the latest version of SQLAlchemy.

The fix was pretty simple: 
 - Here's my PR in the `pandasql` repo: https://github.com/yhat/pandasql/pull/104
 - Until this is merged (seems unlikely since the repo has been inactive since 2017), I have added my fork as the dependency as a fix. 
 - Not sure if PyPI would accept this as a release.

Possible solutions:

1. Re-release `pandasql` as `pandasql2` but I don't want to take credit for something I didn't develop.
2. Bypass use of `pandasql` and directly use SQLAlchemy for our purposes. This I think will be quite an undertaking.

Thoughts @cmungall , @matentzn ?

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -49,7 +49,7 @@ keywords =
 
 [options]
 install_requires =
-    sqlalchemy<2.0.0
+    sqlalchemy
     pyparsing
     bioregistry
     click
--- a/setup.cfg
+++ b/setup.cfg
@@ -57,7 +57,7 @@ install_requires =
     networkx
     numpy
     pandas
-    pandasql@git+https://git@github.com/hrshdhgd/pandasql.git
+    pansql
     pyyaml
     rdflib>=6
     scipy
--- a/setup.cfg
+++ b/setup.cfg
@@ -49,7 +49,6 @@ keywords =
 
 [options]
 install_requires =
-    sqlalchemy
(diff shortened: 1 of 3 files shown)


cmungall:
I'm not sure I love the dependency on something distributed from github

I'm not so worried abut putting a fork on pypi if the maintainers are not willing (I am not sure I would claim foo2 as the name, maybe better to go with a different name..). However this is a bit of a commitment.

hrshdhgd:
Alright here's the solution. I just published my fork to pypi calling it [pansql](https://pypi.org/project/pansql/). In terms of code, nothing changes. You still import `pandasql` as is and all is well. Behind the scenes, it's using our fork rather than the original.

hrshdhgd:
Refactored `pandasql` to `pansql` to avoid confusion.

matentzn (approved):
THANK you for your patience to find a good solution!

PR #346, opened 2023-02-27
New method: flip

### LINKED ISSUE #315: New method: flip
@kevinscharper noted in https://github.com/monarch-initiative/monarch-ingest/issues/326 that if you are not careful and "blindly apply" a mapping file to a dataset, you may end up applying it the wrong way. For example, Mondo SSSOM mappings have Mondo IDs in as `subject_id` and external ID's such as OMIM as `object_id`s. 

I propose the development of a `flip` method that allows to swap `subject` and `object`. This is not just about "flipping IDs", you have to 

- make sure that all `subject` and `object` specific fields are flipped accordingly, and
- skos:broadMatch gets flipped to skos:narrowMatch etc.

This method could be useful, I think, because it allows us to centralise the logic by which we permit flipping. For example, we can simply ignore cases where flipping would be clearly wrong, or the "inverse" properties are not known.

hrshdhgd:
- Flips `subject_id`<=>`object_id` such that all prefixes for the `subject_id` are the one provided by the user. 
 - The `predicate_id` also gets flipped accordingly.

Fixes #315

### DIFF
--- a/sssom/cli.py
+++ b/sssom/cli.py
@@ -58,6 +58,7 @@ from .util import (
     remove_unmatched,
     sort_df_rows_columns,
     to_mapping_set_dataframe,
+    flip_nodes,
 )
 from .writers import write_table
 
@@ -713,6 +714,32 @@ def remove(input: str, output: TextIO, remove_map: str):
     remove_msdf = parse_sssom_table(remove_map)
     input_msdf.remove_mappings(remove_msdf)
     write_table(input_msdf, output)
+
+@main.command()
+@input_argument
+@output_option
+@click.option(
+    "-P",
+    "--subject-prefix",
+    help="Flip subject_id and object_id such that all subject_ids have the same prefix.",
+)
+@click.option(
+    "--merge-flipped/--no-merge-flipped",
+    default=True,
+    is_flag=True,
+    help="If True (default), add flipped dataframe to input else, just return flipped data.",
+)
+def flip(input: str, output: TextIO, subject_prefix: str, merge_flipped: bool):
+    """
+    Flip subject and object IDs such that all subjects have the prefix provided.
+
+    :param input: SSSOM TSV file.
+    :param prefix: Prefix of all subject_ids.
+    :param output: SSSOM TSV file with columns sorted.
+    """
+    msdf = parse_sssom_table(input)
+    msdf.df = flip_nodes(msdf.df, subject_prefix, merge_flipped)
+    write_table(msdf, output)
 
 
 if __name__ == "__main__":
--- a/sssom/constants.py
+++ b/sssom/constants.py
@@ -139,6 +139,17 @@ PREDICATE_LIST = [
     RDF_SEE_ALSO,
 ]
 
+PREDICATE_FLIP_DICTIONARY = {
+        "skos:closeMatch": "skos:closeMatch",
+        "skos:relatedMatch": "skos:relatedMatch",
+        "skos:narrowMatch" : "skos:broadMatch",
+        "skos:broadMatch" : "skos:narrowMatch",
+        "skos:exactMatch" : "skos:exactMatch",
+        "semapv:crossSpeciesExactMatch" : "semapv:crossSpeciesExactMatch",
+        "semapv:crossSpeciesNarrowMatch" : "semapv:crossSpeciesBroadMatch",
+        "semapv:crossSpeciesBroadMatch" : "semapv:crossSpeciesNarrowMatch",
+    }
+
 
 class SEMAPV(Enum):
     """SEMAPV Enum containing different mapping_justification."""
--- a/sssom/util.py
+++ b/sssom/util.py
@@ -46,10 +46,12 @@ from .constants import (
     OBJECT_CATEGORY,
     OBJECT_ID,
     OBJECT_LABEL,
+    OBJECT_MATCH_FIELD,
     OBJECT_SOURCE,
     OBO_HAS_DB_XREF,
     OWL_DIFFERENT_FROM,
     OWL_EQUIVALENT_CLASS,
+    PREDICATE_FLIP_DICTIONARY,
     PREDICATE_ID,
     PREDICATE_LIST,
     PREDICATE_MODIFIER,
@@ -67,6 +69,7 @@ from .constants import (
     SUBJECT_CATEGORY,
     SUBJECT_ID,
     SUBJECT_LABEL,
+    SUBJECT_MATCH_FIELD,
     SUBJECT_SOURCE,
     SSSOMSchemaView,
 )
@@ -1505,3 +1508,42 @@ def _get_sssom_schema_object() -> SSSOMSchemaView:
         else SSSOMSchemaView()
     )
     return sssom_sv_object
+
+def flip_nodes(df: pd.DataFrame, subject_prefix: str, merge_flipped: bool = True) -> pd.DataFrame:
+    """Switching subject and objects based on their prefixes and adjusting predicates accordingly.
+
+    :param df: Pandas dataframe.
+    :param prefix: Prefix of subjects desired.
+    :return: Pandas dataframe with all subject IDs having the same prefix.
+    """
+    condition_1 = df[SUBJECT_ID].str.startswith(subject_prefix+":")
+    condition_2 = df[OBJECT_ID].str.startswith(subject_prefix+":")
+    condition_3 = df[PREDICATE_MODIFIER] == PREDICATE_MODIFIER_NOT
+    predicate_flip_map = PREDICATE_FLIP_DICTIONARY
+
+    predicate_modified_df = pd.DataFrame(df[condition_3])
+    non_predicate_modified_df = pd.DataFrame(df[~condition_3])
+
+    prefixed_subjects_df = pd.DataFrame(non_predicate_modified_df[(condition_1 & ~condition_2)])
+    non_prefix_subjects_df = pd.DataFrame(non_predicate_modified_df[(~condition_1 & condition_2)])
+    df_to_flip = non_prefix_subjects_df.loc[non_prefix_subjects_df[PREDICATE_ID].isin(list(predicate_flip_map.keys()))]
+    flipped_df = df_to_flip.rename(columns={
+        SUBJECT_ID:OBJECT_ID, 
+        SUBJECT_LABEL:OBJECT_LABEL,
+        SUBJECT_CATEGORY: OBJECT_CATEGORY,
+        SUBJECT_MATCH_FIELD: OBJECT_MATCH_FIELD,
+        SUBJECT_SOURCE: OBJECT_SOURCE,
+        OBJECT_ID:SUBJECT_ID, 
+        OBJECT_LABEL:SUBJECT_LABEL,
+        OBJECT_CATEGORY: SUBJECT_CATEGORY,
+        OBJECT_MATCH_FIELD: SUBJECT_MATCH_FIELD,
+        OBJECT_SOURCE: SUBJECT_SOURCE,
+    })
+    flipped_df = flipped_df[df.columns]
+    flipped_df[PREDICATE_ID] = flipped_df[PREDICATE_ID].apply(lambda x: predicate_flip_map[x])
+
+    return_df = pd.concat([prefixed_subjects_df, predicate_modified_df, flipped_df]).drop_duplicates()
+    if merge_flipped:
+        return pd.concat([df, return_df]).drop_duplicates()
+    else:
+        return return_df
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -4,7 +4,7 @@ import unittest
 from sssom.constants import OBJECT_ID, SUBJECT_ID
 from sssom.io import extract_iri
 from sssom.parsers import parse_sssom_table
-from sssom.util import MappingSetDataFrame, filter_out_prefixes, filter_prefixes
+from sssom.util import MappingSetDataFrame, filter_out_prefixes, filter_prefixes, flip_nodes
 from tests.constants import data_dir
 
 
@@ -14,6 +14,7 @@ class TestIO(unittest.TestCase):
     def setUp(self) -> None:
         """Set up."""
         self.msdf = parse_sssom_table(f"{data_dir}/basic.tsv")
+        self.msdf2 = parse_sssom_table(f"{data_dir}/basic7.tsv")
         self.features = [SUBJECT_ID, OBJECT_ID]
 
     def test_broken_predicate_list(self):
@@ -81,3 +82,17 @@ class TestIO(unittest.TestCase):
                 set(new_msdf.prefix_map.keys())
             ),
         )
+
+    def test_flip_nodes(self):
+        """Test flip nodes."""
+        subject_prefix = "a"
+        original_msdf = self.msdf2
+        flipped_df = flip_nodes(original_msdf.df, subject_prefix, False)
+        self.assertEqual(len(flipped_df), 14)
+    
+    def test_flip_nodes_merged(self):
+        """Test flip nodes."""
+        subject_prefix = "a"
+        original_msdf = self.msdf2
+        flipped_df = flip_nodes(original_msdf.df, subject_prefix, True)
+        self.assertEqual(len(flipped_df), 36)


sssom/cli.py
```diff
@@ -53,6 +53,7 @@
     compare_dataframes,
     dataframe_to_ptable,
     filter_redundant_rows,
+    flip_nodes,
```
matentzn, line 56: better call it flip_mappings

sssom/cli.py
```diff
@@ -715,5 +716,34 @@ def remove(input: str, output: TextIO, remove_map: str):
...
+@click.option(
+    "-P",
+    "--subject-prefix",
+    help="Flip subject_id and object_id such that all subject_ids have the same prefix.",
+)
+@click.option(
+    "--merge-flipped/--no-merge-flipped",
+    default=True,
+    is_flag=True,
+    help="If True (default), add flipped dataframe to input else, just return flipped data.",
```
matentzn, line 731:
```suggestion
    help="If True (default), add flipped mappings to the input mapping set, else, just return flipped mappings as a separate mapping set.",
```

sssom/cli.py
```diff
@@ -715,5 +716,34 @@ def remove(input: str, output: TextIO, remove_map: str):
     write_table(input_msdf, output)
 
 
+@main.command()
+@input_argument
+@output_option
+@click.option(
+    "-P",
+    "--subject-prefix",
+    help="Flip subject_id and object_id such that all subject_ids have the same prefix.",
```
matentzn, line 725: Cool!

sssom/constants.py
```diff
@@ -139,6 +139,17 @@
     RDF_SEE_ALSO,
 ]
 
+PREDICATE_FLIP_DICTIONARY = {
+    "skos:closeMatch": "skos:closeMatch",
```
matentzn, line 143: Are there perhaps CONSTANT values somewhere we can use for this dict?

sssom/cli.py
hrshdhgd: Done

sssom/constants.py
hrshdhgd: Yes, same file..lol. Replaced

sssom/util.py
```diff
@@ -1505,3 +1508,58 @@ def _get_sssom_schema_object() -> SSSOMSchemaView:
...
+            OBJECT_MATCH_FIELD: SUBJECT_MATCH_FIELD,
+            OBJECT_SOURCE: SUBJECT_SOURCE,
+        }
+    )
+    flipped_df = flipped_df[df.columns]
+    flipped_df[PREDICATE_ID] = flipped_df[PREDICATE_ID].apply(
+        lambda x: predicate_flip_map[x]
+    )
+
+    return_df = pd.concat([prefixed_subjects_df, flipped_df]).drop_duplicates()
```
hrshdhgd, line 1561: @matentzn , I need a second pair of eyes to check this logic. Is this returning what is needed?

sssom/util.py
```diff
@@ -1505,3 +1508,58 @@ def _get_sssom_schema_object() -> SSSOMSchemaView:
...
+    if merge_flipped:
+        return pd.concat([df, predicate_modified_df, return_df]).drop_duplicates()
+    else:
+        return return_df
```
hrshdhgd, line 1565: @matentzn , I need a second pair of eyes to check this logic. Is this returning what is needed?

sssom/util.py
```diff
@@ -1505,3 +1508,58 @@ def _get_sssom_schema_object() -> SSSOMSchemaView:
...
+        }
+    )
+    flipped_df = flipped_df[df.columns]
+    flipped_df[PREDICATE_ID] = flipped_df[PREDICATE_ID].apply(
+        lambda x: predicate_flip_map[x]
+    )
+
+    return_df = pd.concat([prefixed_subjects_df, flipped_df]).drop_duplicates()
+    if merge_flipped:
+        return pd.concat([df, predicate_modified_df, return_df]).drop_duplicates()
```
matentzn, line 1563: `predicate_modified_df` -> Perhaps we should not flip these at all. Since df is the original dataframe, why is it merged here again?

sssom/constants.py
```diff
@@ -139,6 +142,17 @@
...
 
+PREDICATE_FLIP_DICTIONARY = {
+    SKOS_CLOSE_MATCH: SKOS_CLOSE_MATCH,
+    SKOS_RELATED_MATCH: SKOS_RELATED_MATCH,
+    SKOS_NARROW_MATCH: SKOS_BROAD_MATCH,
+    SKOS_BROAD_MATCH: SKOS_NARROW_MATCH,
+    SKOS_EXACT_MATCH: SKOS_EXACT_MATCH,
+    CROSS_SPECIES_EXACT_MATCH: CROSS_SPECIES_EXACT_MATCH,
+    CROSS_SPECIES_NARROW_MATCH: CROSS_SPECIES_BROAD_MATCH,
+    CROSS_SPECIES_BROAD_MATCH: CROSS_SPECIES_NARROW_MATCH,
```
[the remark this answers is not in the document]
hrshdhgd: Done

sssom/util.py
```diff
@@ -1505,3 +1508,58 @@ def _get_sssom_schema_object() -> SSSOMSchemaView:
...
+
+    :param df: Pandas dataframe.
+    :param subject_prefix: Prefix of subjects desired.
+    :param merge_flipped: If True (default), add flipped dataframe to input else,
+                          just return flipped data.
+    :return: Pandas dataframe with all subject IDs having the same prefix.
+    """
+    condition_1 = df[SUBJECT_ID].str.startswith(subject_prefix + ":")
+    condition_2 = df[OBJECT_ID].str.startswith(subject_prefix + ":")
+    condition_3 = df[PREDICATE_MODIFIER] == PREDICATE_MODIFIER_NOT
```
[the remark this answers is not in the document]
matentzn: "adding it to the final returned df" --> 

I would:

Keep the input DF as is (no changes).

create the subsets you want to flip. Flip these. If "merge=true" append to the original DF and return, else just change column order to what the original column order was and then return. 

So no need "adding it the final return" as they are already part of the original DF

sssom/util.py
```diff
@@ -1505,3 +1508,58 @@ def _get_sssom_schema_object() -> SSSOMSchemaView:
...
+            SUBJECT_ID: OBJECT_ID,
+            SUBJECT_LABEL: OBJECT_LABEL,
+            SUBJECT_CATEGORY: OBJECT_CATEGORY,
+            SUBJECT_MATCH_FIELD: OBJECT_MATCH_FIELD,
+            SUBJECT_SOURCE: OBJECT_SOURCE,
+            OBJECT_ID: SUBJECT_ID,
+            OBJECT_LABEL: SUBJECT_LABEL,
+            OBJECT_CATEGORY: SUBJECT_CATEGORY,
+            OBJECT_MATCH_FIELD: SUBJECT_MATCH_FIELD,
+            OBJECT_SOURCE: SUBJECT_SOURCE,
```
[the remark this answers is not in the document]
hrshdhgd: Done

sssom/util.py
```diff
@@ -1505,3 +1508,58 @@ def _get_sssom_schema_object() -> SSSOMSchemaView:
...
+) -> pd.DataFrame:
+    """Switching subject and objects based on their prefixes and adjusting predicates accordingly.
+
+    :param df: Pandas dataframe.
+    :param subject_prefix: Prefix of subjects desired.
+    :param merge_flipped: If True (default), add flipped dataframe to input else,
+                          just return flipped data.
+    :return: Pandas dataframe with all subject IDs having the same prefix.
+    """
+    condition_1 = df[SUBJECT_ID].str.startswith(subject_prefix + ":")
```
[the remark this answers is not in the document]
hrshdhgd: Done

sssom/util.py
hrshdhgd: Isn't that the whole point of the `merge_flipped` flag? Flip + add to original DataFrame?

matentzn: What is exactly predicate_modified_df as opposed to `return_df`?

sssom/util.py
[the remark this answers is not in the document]
hrshdhgd: I'm sorry I do not follow. I think I'm already doing what you are explaining here.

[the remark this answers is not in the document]
matentzn: Good then!

I was just confused that the return has 3 files that are merged, rather than 2.

sssom/util.py
hrshdhgd: `predicate_modified_df` is the df of all rows that have `Not` in the `predicate_modifier` column

hrshdhgd: I am not flipping that df at all. Just adding it back at the end.

matentzn: Yes, but isnt `df` entirely unchanged, and already contains `predicate_modified_df`?

sssom/util.py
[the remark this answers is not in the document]
matentzn: condition_3 = df[PREDICATE_MODIFIER] == PREDICATE_MODIFIER_NOT

I would still recommend to make this stricter: Not just if it is not, but any future modifier should lead for this to be ignored by the flip command

sssom/util.py
hrshdhgd: Made changes and committed

matentzn: drop_duplicates() would have indeed taken care of it, but better this way.

sssom/constants.py
```diff
@@ -120,6 +120,9 @@
 SKOS_NARROW_MATCH = "skos:narrowMatch"
 OBO_HAS_DB_XREF = "oboInOwl:hasDbXref"
 SKOS_RELATED_MATCH = "skos:relatedMatch"
+CROSS_SPECIES_EXACT_MATCH = "semapv:crossSpeciesExactMatch"
+CROSS_SPECIES_NARROW_MATCH = "semapv:crossSpeciesNarrowMatch"
+CROSS_SPECIES_BROAD_MATCH = "semapv:crossSpeciesBroadMatch"
 RDF_SEE_ALSO = "rdfs:seeAlso"
 SSSOM_SUPERCLASS_OF = "inverseOf(owl:subClassOf)"
```
matentzn, line 127:
```suggestion
SSSOM_SUPERCLASS_OF = "sssom:superClassOf"
```

sssom/util.py
```diff
@@ -1505,3 +1512,84 @@ def _get_sssom_schema_object() -> SSSOMSchemaView:
...
+                ~subject_starts_with_prefix_condition
+                & object_starts_with_prefix_condition
+            )
+        ]
+    )
+    df_to_flip = non_prefix_subjects_df.loc[
+        non_prefix_subjects_df[PREDICATE_ID].isin(list(predicate_flip_map.keys()))
+    ]
+    list_of_subject_object_columns = [
+        x for x in df_to_flip.columns if x.startswith(("subject", "object"))
```
matentzn, line 1560: Sorry I didnt notice this before but can we reconcile this with:

https://github.com/mapping-commons/sssom-py/blob/master/sssom/parsers.py#L741

matentzn: I dont really know how applicable these are but I would kind of like if there was only one util method in sssom-py doing the flipping, rather than the same logic being at various places.

matentzn (commented):
Sorry, this should be the final comments. This is an important feature, else I wouldn't spend that much time getting it right!

sssom/util.py
```diff
@@ -1505,3 +1512,84 @@ def _get_sssom_schema_object() -> SSSOMSchemaView:
...
+    ]
+    flipped_df = df_to_flip.rename(
+        columns=_flip_column_names(list_of_subject_object_columns)
+    )
+    flipped_df = flipped_df[df.columns]
+    flipped_df[PREDICATE_ID] = flipped_df[PREDICATE_ID].apply(
```
matentzn, line 1566: I think we also need to attach additional metadata:

- [ ] set https://mapping-commons.github.io/sssom/mapping_source/ (`flipped_df[MAPPING_SOURCE]`) to the `mapping_set_id` of the input df.
- [ ] set `flipped_df[MAPPING_JUSTIFICATION]` to "sempav:FlippedMapping" (see https://github.com/mapping-commons/semantic-mapping-vocabulary/pull/23 for context)

sssom/constants.py
```diff
@@ -139,6 +142,21 @@
     RDF_SEE_ALSO,
 ]
 
+PREDICATE_FLIP_DICTIONARY = {
```
cthoyt, line 145: Can we please externalize this mapping such that it can be re-used?

cthoyt: Also such that it can be configured further without changing Python code.

tests/test_utils.py
```diff
@@ -14,6 +19,7 @@ class TestIO(unittest.TestCase):
     def setUp(self) -> None:
         """Set up."""
         self.msdf = parse_sssom_table(f"{data_dir}/basic.tsv")
+        self.msdf2 = parse_sssom_table(f"{data_dir}/basic7.tsv")
```
cthoyt, line 22: can you please make better names for these test data and variables?

sssom/util.py
cthoyt:
```suggestion
    flipped_df[PREDICATE_ID] = flipped_df[PREDICATE_ID].map(predicate_flip_map)
```

sssom/cli.py
```diff
@@ -715,5 +716,34 @@ def remove(input: str, output: TextIO, remove_map: str):
     write_table(input_msdf, output)
 
 
+@main.command()
+@input_argument
+@output_option
+@click.option(
```
matentzn, line 722: Add one more option `--inverse-map` that allows passing in a custom inverse map to overwrite / extend the default one provided by sssom-py

sssom/cli.py
```diff
@@ -53,6 +53,7 @@
     compare_dataframes,
     dataframe_to_ptable,
     filter_redundant_rows,
+    flip_mappings,
```
matentzn, line 56: replace the entire `flip` terminology with `invert`

matentzn: i.e. `invert_mappings`

sssom/cli.py
matentzn: (parameters, method names, etc)

sssom/constants.py
matentzn: Yes we had already discussed it between us, but said to externlise in a future PR. I am less concerned with externalising (moving to a separately managed config file) the sssom default flip dictionary (I leave this up to @hrshdhgd to decide), but I think being able to override this from the user side is pretty important, so I requested the method to be extended.

sssom/util.py
matentzn: Applying this change @cthoyt will break the code, but yes the syntax seems more terse. Thanks, I didn't know you can do it this way!

sssom/util.py
```diff
@@ -1505,3 +1512,84 @@ def _get_sssom_schema_object() -> SSSOMSchemaView:
...
+    return_df = pd.concat([prefixed_subjects_df, flipped_df]).drop_duplicates()
+    if merge_flipped:
+        return pd.concat([df, return_df]).drop_duplicates()
+    else:
+        return return_df
+
+
+def _flip_column_names(column_names) -> dict:
+    """Return a dictionary for column renames in pandas DataFrame."""
+    column_flip_map = {
```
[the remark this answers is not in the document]
matentzn: Yes, lets add it to constants.py for now, and add it as a parameter to the method signature that can be overwritten by the user.

tests/test_utils.py
matentzn: @cthoyt if this is important lets make a separate issue about this, we have 25 tests written this way now, and I don't think we need to start now.

matentzn: (keep PRs small)

sssom/cli.py
hrshdhgd: If we extend external inverse map, what happens if there is a conflict in key-value pair (internal vs external)?

hrshdhgd: Also, what will be the file format of the inverse map?

sssom/cli.py
hrshdhgd: Even the CLI command `flip` => `invert`?

matentzn: yes lets go for unified termonology. Thank you and really sorry this comes up now, but better now than later.

sssom/cli.py
matentzn: Makes sense! Yes.

sssom/constants.py
hrshdhgd: Done

sssom/cli.py
hrshdhgd: Done

sssom/util.py
hrshdhgd: This is for a `Mapping` object. What we have in this PR is for `MappingSetDataFrame`. I'll need to think about this.

tests/test_reconcile.py
```diff
@@ -26,9 +26,9 @@ def test_filter(self):
     def test_deal_with_negation(self):
         """Test handling negating returns the right number of rows."""
         df1 = deal_with_negation(self.msdf1.df)
-        self.assertEqual(7, len(df1.index))
+        self.assertEqual(8, len(df1.index))
         df2 = deal_with_negation(self.msdf2.df)
-        self.assertEqual(5, len(df2.index))
+        self.assertEqual(13, len(df2.index))
```
hrshdhgd, line 32: These changes are due to a bug-fix in `util.py`

sssom/util.py
```diff
@@ -764,7 +766,7 @@ def deal_with_negation(df: pd.DataFrame) -> pd.DataFrame:
                 & (combined_normalized_subset[CONFIDENCE] == row_1[CONFIDENCE])
                 & (
                     combined_normalized_subset[MAPPING_JUSTIFICATION]
-                    == SEMAPV.ManualMappingCuration
+                    == SEMAPV.ManualMappingCuration.value
```
hrshdhgd, line 769: bug-fix

sssom/util.py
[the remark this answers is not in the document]
hrshdhgd: Added `inverse_map.yaml` which holds the default values and is parsed by default if no file path is provided for the `--inverse-map` param.

sssom/cli.py
hrshdhgd: Done

sssom/util.py
matentzn: Great!

sssom/cli.py
```diff
@@ -715,5 +716,52 @@ def remove(input: str, output: TextIO, remove_map: str):
...
+    is_flag=True,
+    help="If True (default), add inverted mappings to the input mapping set, else, just return inverted mappings as a separate mapping set.",
+)
+@click.option(
+    "--inverse-map", help="Path to file that contains the inverse predicate dictionary."
+)
+def invert(
+    input: str,
+    output: TextIO,
+    subject_prefix: str,
```
matentzn, line 739: subject_prefix should be optional!

sssom/util.py
```diff
@@ -1505,3 +1507,75 @@ def _get_sssom_schema_object() -> SSSOMSchemaView:
         else SSSOMSchemaView()
     )
     return sssom_sv_object
+
+
+def invert_mappings(
+    df: pd.DataFrame,
+    subject_prefix: str,
```
matentzn, line 1514: subject_prefix should be optional - it should be possible to flip an entire mapping set without worrying about subject prefix

matentzn (commented):
This looks awesome. I saw only one issue I didn't catch before, which is this subject_prefix should be optional. It should be possible to simply flip a dataframe without any parameters, flipping everything (that is not modified).

sssom/inverse_map.yaml
```diff
@@ -0,0 +1,14 @@
+inverse_predicate_map:
+  skos:closeMatch: skos:closeMatch
+  skos:relatedMatch: skos:relatedMatch
+  skos:exactMatch: skos:exactMatch
+  skos:narrowMatch: skos:broadMatch
+  skos:broadMatch: skos:narrowMatch
+  semapv:crossSpeciesExactMatch: semapv:crossSpeciesExactMatch
+  semapv:crossSpeciesNarrowMatch: semapv:crossSpeciesBroadMatch
+  semapv:crossSpeciesBroadMatch: semapv:crossSpeciesNarrowMatch
```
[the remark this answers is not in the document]
matentzn: We changed our mind on this - adding cross species close match

sssom/util.py
hrshdhgd: Done

sssom/util.py
matentzn: Can we make an issue to this end and resolve this comment?

hrshdhgd: https://github.com/mapping-commons/sssom-py/issues/349

matentzn (approved):
This looks great now! Looking forward to testing it! :)

PR #350, opened 2023-03-13
`merge` command returned nothing when input `msdfs` had no `confidence` column

### LINKED ISSUE #348: merge two data frames with --reconcile true leads to empty result
Merging these two data frames:

```
# curie_map:
#   UBERON: http://purl.obolibrary.org/obo/UBERON_
#   ZFS: http://purl.obolibrary.org/obo/ZFS_
#   oio: http://www.geneontology.org/formats/oboInOwl#
# license: https://w3id.org/sssom/license/unspecified
# mapping_set_id: https://w3id.org/sssom/mappings/72debc9d-ca69-45e8-b46d-aef8361bedf2
# object_source: ZFS
# subject_source: UBERON
subject_id	subject_label	predicate_id	object_id	mapping_justification	subject_source	object_source
UBERON:0000069	larval stage	oio:hasDbXref	ZFS:0000048	semapv:UnspecifiedMatching	UBERON	ZFS
UBERON:0000105	life cycle stage	oio:hasDbXref	ZFS:0100000	semapv:UnspecifiedMatching	UBERON	ZFS
UBERON:0000105	life cycle stage	oio:hasDbXref	ZFS:0000000	semapv:UnspecifiedMatching	UBERON	ZFS
```

and 

```
# curie_map:
#   UBERON: http://purl.obolibrary.org/obo/UBERON_
#   WBls: http://purl.obolibrary.org/obo/WBls_
#   oio: http://www.geneontology.org/formats/oboInOwl#
# license: https://w3id.org/sssom/license/unspecified
# mapping_set_id: https://w3id.org/sssom/mappings/c5e357f5-86df-4aaa-a30e-8a23ad523ab2
# object_source: WBls
# subject_source: UBERON
subject_id	subject_label	predicate_id	object_id	mapping_justification	subject_source	object_source
UBERON:0000066	fully formed stage	oio:hasDbXref	WBls:0000041	semapv:UnspecifiedMatching	UBERON	WBls
UBERON:0000068	embryo stage	oio:hasDbXref	WBls:0000003	semapv:UnspecifiedMatching	UBERON	WBls
UBERON:0000068	embryo stage	oio:hasDbXref	WBls:0000102	semapv:UnspecifiedMatching	UBERON	WBls
UBERON:0000068	embryo stage	oio:hasDbXref	WBls:0000092	semapv:UnspecifiedMatching	UBERON	WBls
```

results in an empty data frame.

Does not work:
```
sssom merge tmp/anatomy-mapping-uberon-FBbt.sssom.tsv tmp/anatomy-mapping-uberon-WBbt.sssom.tsv --reconcile true
```

Works:
```
sssom merge tmp/anatomy-mapping-uberon-FBbt.sssom.tsv tmp/anatomy-mapping-uberon-WBbt.sssom.tsv --reconcile false
```

We should also add a test to this end

hrshdhgd:
Fixes #348

### DIFF
--- a/sssom/util.py
+++ b/sssom/util.py
@@ -277,7 +277,7 @@ def filter_redundant_rows(
     # create a 'sort' method and then replce the following line by sort()
     df = sort_sssom(df)
     # df[CONFIDENCE] = df[CONFIDENCE].apply(lambda x: x + random.random() / 10000)
-    df, nan_df = assign_default_confidence(df)
+    df, nan_df, confidence_column_created = assign_default_confidence(df)
     if ignore_predicate:
         key = [SUBJECT_ID, OBJECT_ID]
     else:
@@ -359,7 +359,7 @@ def filter_redundant_rows(
             [get_row_based_on_hierarchy(concerned_df), return_df], axis=0
         ).drop_duplicates()
 
-    if return_df[CONFIDENCE].isnull().all():
+    if confidence_column_created:
         return_df = return_df.drop(columns=[CONFIDENCE], axis=1)
     return return_df
 
@@ -392,24 +392,26 @@ def get_row_based_on_hierarchy(df: pd.DataFrame):
 
 def assign_default_confidence(
     df: pd.DataFrame,
-) -> Tuple[pd.DataFrame, pd.DataFrame]:
+) -> Tuple[pd.DataFrame, pd.DataFrame, bool]:
     """Assign :data:`numpy.nan` to confidence that are blank.
 
     :param df: SSSOM DataFrame
     :return: A Tuple consisting of the original DataFrame and dataframe consisting of empty confidence values.
     """
+    confidence_column_created = False
     # Get rows having numpy.NaN as confidence
     if df is not None:
         new_df = df.copy()
         if CONFIDENCE not in new_df.columns:
             new_df[CONFIDENCE] = 0.0  # np.NaN
             nan_df = pd.DataFrame(columns=new_df.columns)
+            confidence_column_created = True
         else:
             new_df = df[~df[CONFIDENCE].isna()]
             nan_df = df[df[CONFIDENCE].isna()]
     else:
         ValueError("DataFrame cannot be empty to 'assign_default_confidence'.")
-    return new_df, nan_df
+    return new_df, nan_df, confidence_column_created
 
 
 def remove_unmatched(df: pd.DataFrame) -> pd.DataFrame:
@@ -703,7 +705,7 @@ def deal_with_negation(df: pd.DataFrame) -> pd.DataFrame:
     """
 
     # Handle DataFrames with no 'confidence' column (basically adding a np.NaN to all non-numeric confidences)
-    df, nan_df = assign_default_confidence(df)
+    df, nan_df, confidence_column_created = assign_default_confidence(df)
     if df is None:
         raise ValueError(
             "The dataframe, after assigning default confidence, appears empty (deal_with_negation)"
@@ -822,6 +824,9 @@ def deal_with_negation(df: pd.DataFrame) -> pd.DataFrame:
         return_df = reconciled_df
     else:
         return_df = reconciled_df.append(nan_df).drop_duplicates()
+
+    if confidence_column_created:
+        return_df = return_df.drop(columns=[CONFIDENCE], axis=1)
 
     return return_df
 
--- a/sssom/util.py
+++ b/sssom/util.py
@@ -277,7 +277,8 @@ def filter_redundant_rows(
     # create a 'sort' method and then replce the following line by sort()
     df = sort_sssom(df)
     # df[CONFIDENCE] = df[CONFIDENCE].apply(lambda x: x + random.random() / 10000)
-    df, nan_df, confidence_column_created = assign_default_confidence(df)
+    confidence_in_original = CONFIDENCE in df.columns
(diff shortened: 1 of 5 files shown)


hrshdhgd:
I made this decision. Does this seem logical @matentzn ?

sssom/util.py
```diff
@@ -273,6 +273,8 @@ def filter_redundant_rows(
     :param ignore_predicate: If true, the predicate_id column is ignored, defaults to False
     :return: Filtered pandas DataFrame
     """
+    if CONFIDENCE not in df.columns:
```
matentzn, line 276: What would happen if we assumed that missing confidence means 100% confidence? I think this assumption is more often true. Also, if you do ad the IF clause here, the dataframe will not be sorted?

hrshdhgd: I just realized, line 280 has `assign_default_confidence` which adds a new column 'confidence' if absent BUT it initiates to `np.NaN`. Is this what we still intend to do?

matentzn: I think so, but it is important that the absence of a confidence value does not lead to a removal of the entire mapping! While it is not the same as confidence 1.0, it should probably be interpreted as such during reconciliation.. Unless you see a strong reason against it?

hrshdhgd: The presence of `np.NaN` is causing the df returned to be empty at the codes [between lines 297 to 311](https://github.com/mapping-commons/sssom-py/blob/issue-348/sssom/util.py#L297-L311)

hrshdhgd: Would it make sense to itiniate it to "0.0" instead of `np.Nan`

sssom/util.py
```diff
@@ -402,7 +402,7 @@ def assign_default_confidence(
     if df is not None:
         new_df = df.copy()
         if CONFIDENCE not in new_df.columns:
-            new_df[CONFIDENCE] = np.NaN
+            new_df[CONFIDENCE] = 0.0  # np.NaN
```
matentzn, line 405: I dont have a problem with this for the reconciliation, but I really wouldn't want the 0% confidence to be exported in the exported mapping set - this is altering the data.

Also, this is not solving the issue where a mapping set has a `confidence` column with some missing values..

matentzn: What happens now if a mapping with a justification with  `NA` confidence (empty value) and one justification with a `0` mapping co-incide? It seems odd tha NA mappings are not set to 0 as well.

matentzn: Ok if this is the case it's fine. I would just like to generally err on the conservative side, removing only the obvious cases. Ok let's go with your solution, and also make reconcile false by default.

hrshdhgd: `reconcile` is already `False` by default.

matentzn: This is weird, because I only noticed this problem because the reconciliation method kicked in without me adding a parameter. Can you double check that there is no place in the code that accidentally sets it to `true`?

sssom/util.py
```diff
@@ -277,7 +277,7 @@ def filter_redundant_rows(
     # create a 'sort' method and then replce the following line by sort()
     df = sort_sssom(df)
     # df[CONFIDENCE] = df[CONFIDENCE].apply(lambda x: x + random.random() / 10000)
-    df, nan_df = assign_default_confidence(df)
+    df, nan_df, confidence_column_created = assign_default_confidence(df)
```
matentzn, line 280: Could you keep track of wether there was a confidence column in the context of filter redundant rows and avoid creating the confidence column created var?

matentzn (commented):
Left a xomment

sssom/util.py
hrshdhgd: I could drop the `confidence` column at the end `if` all rows are `0.0` for that column but that's making an assumption that it happens only when we add a `confidence` column.

matentzn: So this is not an option?

```
    confidence_in_original = CONFIDENCE in df.columns
    df, nan_df = assign_default_confidence(df)
    if not confidence_in_original:
         return_df = return_df.drop(columns=[CONFIDENCE], axis=1)
```

Maybe I am misunderstanding something

hrshdhgd: Aah .... I thought you didn't want a new variable in general. We can do this.

hrshdhgd: Done

sssom/util.py
```diff
@@ -823,6 +825,9 @@ def deal_with_negation(df: pd.DataFrame) -> pd.DataFrame:
     else:
         return_df = reconciled_df.append(nan_df).drop_duplicates()
 
+    if confidence_in_original:
```
matentzn, line 828:
```suggestion
    if not confidence_in_original:
```

?

sssom/util.py
```diff
@@ -359,7 +360,7 @@ def filter_redundant_rows(
             [get_row_based_on_hierarchy(concerned_df), return_df], axis=0
         ).drop_duplicates()
 
-    if return_df[CONFIDENCE].isnull().all():
+    if confidence_in_original:
```
matentzn, line 363:
```suggestion
    if not confidence_in_original:
```

hrshdhgd:
Good catch! sorry about that!

matentzn (commented):
Can you add a test case for reconcile false?

matentzn (approved):
Left a comment

matentzn:
Nice! Thanks!

PR #352, opened 2023-03-14
Makes `clean_prefix_map()` optionally strict.

### LINKED ISSUE #351: `clean_prefix_map()` should have a `strict` flag [default = False]
This need arises due to[ this conversation in `oaklib`](https://github.com/INCATools/ontology-access-kit/issues/462#issuecomment-1468972949).

hrshdhgd:
Fixes #351 

As of the current state, if all prefixes in a dataframe are not accounted for in the `curie_map`, no output is generated and the process errors out.

For now setting the default to be `True` but I'm not sure if it should be True/False. Thoughts @matentzn ?

### DIFF
--- a/sssom/util.py
+++ b/sssom/util.py
@@ -145,8 +145,14 @@ class MappingSetDataFrame:
             description += self.df.tail().to_string() + "\n"
         return description
 
-    def clean_prefix_map(self) -> None:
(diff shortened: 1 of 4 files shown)


hrshdhgd:
Closed because of code-mixing with another PR

PR #353, opened 2023-03-14
Added strict flag for `clean_prefix_map()`

### LINKED ISSUE #351: `clean_prefix_map()` should have a `strict` flag [default = False]

hrshdhgd:
Fixes #351 

As of the current state, if all prefixes in a dataframe are not accounted for in the `curie_map`, no output is generated and the process errors out.

For now setting the default to be `True` but I'm not sure if it should be True/False. Thoughts @matentzn ?

### DIFF
--- a/sssom/util.py
+++ b/sssom/util.py
@@ -147,12 +147,13 @@ class MappingSetDataFrame:
             description += self.df.tail().to_string() + "\n"
         return description
 
-    def clean_prefix_map(self, strict:bool = True) -> None:
+    def clean_prefix_map(self, strict: bool = True) -> None:
         """
         Remove unused prefixes from the internal prefix map based on the internal dataframe.
-        
+
         :param strict: Boolean if True, errors out if all prefixes in dataframe are not
                        listed in the 'curie_map'.
+        :raises ValueError: If prefixes absent in 'curie_map' and strict flag = True
         """
         all_prefixes = []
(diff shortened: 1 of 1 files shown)


matentzn (approved):
This PR is a no-brainer and can be merged, but it wont solve the OAK problem -the question is who is responsible for filling in missing prefixes?

hrshdhgd:
We could do it here using your solution using the missing IRI but does that mean when the missing IRI is added to the `curie_map`, all prefixes that aren't accounted for get an additional prefix?

matentzn:
I was thinking this:

For e.g. let's say ABCD:1234 is unaccounted for. Currently `curie_map` looks like

```
curie_map:
     MONDO:  http://purl.obolibrary.org/obo/MONDO_
```

Now what you do is this inject the ABC prefix:


```
curie_map:
     MONDO:  http://purl.obolibrary.org/obo/MONDO_
     ABCD:  http://w3id.org/sssom/unknown_prefix/abcd/
```

Thats it?

hrshdhgd:
Aaahhh .... much elegant. I likes!

hrshdhgd:
But does that mean the `strict` flag would be useless at that point?

matentzn:
Definetely not. this should only be done in strict=false mode.

PR #358, opened 2023-03-28
Bug fixes

### LINKED ISSUE #359: parse adds Mapping fields to MappingSet header
In my mapping file (`obo_omop.tsv`) I am using the "comment" field to add notes for some of the mappings.  I have the mapping set level metadata in a separate `metadata.yml` file and it does *not* include the "comment" field.  When I run:
```
sssom parse obo_omop.tsv --metadata metadata.yml -o obo_omop.sssom.tsv
```
the resulting `obo_omop.sssom.tsv` file contains the "comment" field as specified in `obo_omop.tsv`, but it also adds "comment" to the MappingSet header along with the last value from the comment field in `obo_omop.tsv`.

I did some experimenting and found that this issue occurs for any of the fields from the MappingSet class that have the same name as a field in the Mapping class (e.g., comment, mapping_date, object_source, subject_source, other, etc.).  So, if you use one of these common name fields in your mapping file and don't use it in the metadata file, then `parse` will add the field to the mapping set header.

hrshdhgd:
- [x] If `predicate_modifier` absent in a `sssom` file, `invert_mappings` errors
- [x] Fixed #359

### DIFF
--- a/sssom/util.py
+++ b/sssom/util.py
@@ -1543,7 +1543,7 @@ def invert_mappings(
     :return: Pandas dataframe with all subject IDs having the same prefix.
     """
     if PREDICATE_MODIFIER not in df.columns:
-        df['predicate_modifier'] = ""
+        df["predicate_modifier"] = ""
     blank_predicate_modifier = df[PREDICATE_MODIFIER] == ""
     if predicate_invert_dictionary:
         predicate_invert_map = predicate_invert_dictionary
--- a/sssom/util.py
+++ b/sssom/util.py
@@ -65,7 +65,6 @@ from .constants import (
     SKOS_EXACT_MATCH,
     SKOS_NARROW_MATCH,
     SKOS_RELATED_MATCH,
-    SLOTS_NOT_IN_METADATA,
     SSSOM_SUPERCLASS_OF,
     SUBJECT_CATEGORY,
     SUBJECT_ID,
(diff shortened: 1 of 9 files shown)


matentzn (dismissed):

sssom/util.py
```diff
@@ -1542,6 +1542,8 @@ def invert_mappings(
     :param predicate_invert_dictionary: YAML file providing the inverse mapping for predicates.
     :return: Pandas dataframe with all subject IDs having the same prefix.
     """
+    if PREDICATE_MODIFIER not in df.columns:
+        df["predicate_modifier"] = ""
```
matentzn, line 1546: Remind me why this has to be here? Will the column removed if this turns out empty?

matentzn (dismissed):
One small question but otherwise looks great.

sssom/util.py
matentzn: Will the column appear in the inverted data frame even if its empty?

hrshdhgd: for now , yes. But sometimes even if the column exists aren't the chances of it being completely empty high?

matentzn: I would prefer if the column is not added if it didn't exist earlier. Is this difficult to achieve?

hrshdhgd: Fixed.

matentzn (approved):
Excellent looks good to me!

PR #361, opened 2023-03-29
Added flag `--strict-clean-prefixes` to `parse()` CLI

### LINKED ISSUE #360: During `parse`, if a prefix is not known, a proxy prefix should be created
https://github.com/mapping-commons/sssom-py/blob/cc22dab18ed4ca00ccd261404c1099a6a10e4b31/sssom/util.py#L151

Probably we only need to add that parameter to the CLI?

hrshdhgd:
Fixes #360


matentzn (dismissed):
WOW! Thanks for doing this so fast!

PR #367, opened 2023-04-13
Added 2 more columns to sorting by rows

hrshdhgd:


sssom/util.py
```diff
@@ -1391,7 +1391,7 @@ def sort_df_rows_columns(
         ]
         df = df.reindex(column_sequence, axis=1)
     if by_rows and len(df) > 0:
-        df = df.sort_values(by=df.columns[0], ignore_index=True)
+        df = df.sort_values(by=[SUBJECT_ID, PREDICATE_ID, OBJECT_ID], ignore_index=True)
```
matentzn, line 1394: how hard would it be to do this (pseudocode):

```
all_slots_in_sssom_sorted = ...
columns_in_df = sort_columns_in_the_same_order_as_list(df.columns, all_slots_in_sssom_sorted)
df = df.sort_values(by=columns_in_df, ignore_index=True)
```

This would generalise the problem and the whole dataframe would be canonical?

matentzn (approved):
Thank you!! Looks great!

PR #369, opened 2023-04-18
Fixing missing variable `non_inverted_df_by_predicate` when `subject_prefix` is provided.

hrshdhgd:
This addresses: https://github.com/monarch-initiative/monarch-mapping-commons/pull/17#issuecomment-1509277116


matentzn (approved):
Can you change the title of the PR before merging this? This has nothing to do with biomappings, right?

PR #371, opened 2023-04-19
Create gh action to update sssom context

### LINKED ISSUE #298: Add GitHub action that refreshes external context
You can sync the bioregistry context like this:

```
make sssom/external_context.py -B
```

We should add a github action that does this and creates a pull request so we can always be informed of changes in the bioregistry context.

### LINKED ISSUE #362: Add github actions workflow that refreshes sssom.context.external on a weekly bases
Add github actions workflow that refreshes sssom.context.external on a weekly basis. This should also make a PR, which I can review.

You can model this action after this one:
https://github.com/INCATools/ontology-development-kit/blob/master/.github/workflows/constraints.yml

Right now, our prefixes are not up to date with bioregistry.

- [ ] ensure the semapv prefix is `https://w3id.org/semapv/vocab/`, note the trailing /vocab. Also search the code for any hard coded references to this namespace that may be wrong.

anitacaron:
Fixes #362
Fixes #298

### DIFF
--- a/.github/workflows/update_context.yml
+++ b/.github/workflows/update_context.yml
@@ -1,0 +1,27 @@
+name: Update sssom context
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: "0 12 * * 1"
+jobs:
+  update_context:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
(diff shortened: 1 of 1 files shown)


matentzn (approved):
Excellent @anitacaron thank you!

PR #378, opened 2023-05-31
Uncommented existing test_cli for "oaei-ordo-hp"

hrshdhgd:
The test for parsing `oaei-ordo-hp.rdf` was commented. This PR just uncomments it and wires is properly in `test_cli.py` to verify if it passes.

### DIFF
--- a/sssom/parsers.py
+++ b/sssom/parsers.py
@@ -552,12 +552,10 @@ def from_alignment_minidom(
                         e.firstChild.nodeValue
                     )
                 elif node_name == "uri1":
-                    import pdb; pdb.set_trace()
+                    # import pdb; pdb.set_trace()
                     ms[SUBJECT_SOURCE] = e.firstChild.nodeValue  # CURIEfy this node value && If endswith(".extension"), replace it with _)
                 elif node_name == "uri2":
-                    import pdb
-
-                    pdb.set_trace()
+                    # import pdb; pdb.set_trace()
                     ms[OBJECT_SOURCE] = e.firstChild.nodeValue
 
     ms.mappings = mlist  # type: ignore
--- a/sssom/parsers.py
+++ b/sssom/parsers.py
@@ -51,6 +51,7 @@ from sssom.constants import (
     SUBJECT_SOURCE_ID,
     SSSOMSchemaView,
 )
+
 from .context import (
     DEFAULT_LICENSE,
     DEFAULT_MAPPING_SET_ID,
--- a/sssom/parsers.py
+++ b/sssom/parsers.py
@@ -18,10 +18,6 @@ import yaml
 from deprecation import deprecated
 from linkml_runtime.loaders.json_loader import JSONLoader
 from rdflib import Graph, URIRef
-
-# from .sssom_datamodel import Mapping, MappingSet
-from sssom_schema import Mapping, MappingSet
-
 from sssom.constants import (
     CONFIDENCE,
     CURIE_MAP,
@@ -51,6 +47,9 @@ from sssom.constants import (
     SUBJECT_SOURCE_ID,
     SSSOMSchemaView,
 )
+
+# from .sssom_datamodel import Mapping, MappingSet
+from sssom_schema import Mapping, MappingSet
 
 from .context import (
     DEFAULT_LICENSE,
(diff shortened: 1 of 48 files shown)


hrshdhgd:
@matentzn : ok, so I'm running :```sssom parse tests/data/oaei-ordo-hp.rdf -o test.tsv --input-format alignment-api-xml ```

and the test.tsv begins
```
# curie_map:
#   HP: http://purl.obolibrary.org/obo/HP_
#   orphanet.ordo: http://www.orpha.net/ORDO/Orphanet_
# license: https://w3id.org/sssom/license/unspecified
# mapping_set_id: https://w3id.org/sssom/mappings/105b111d-1f45-44e3-bb6d-8c33670537b7
# object_source: http://purl.obolibrary.org/hp.owl
# subject_source: http://purl.obolibrary.org/ordo.owl
subject_id	predicate_id	object_id	mapping_justification	confidence
orphanet.ordo:100047	owl:equivalentClass	HP:0100681	semapv:UnspecifiedMatching	0.27449
orphanet.ordo:100069	owl:equivalentClass	HP:0030219	semapv:UnspecifiedMatching	0.875
orphanet.ordo:100088	owl:equivalentClass	HP:0002890	semapv:UnspecifiedMatching	0.875
```

Is the `object_source` value incorrect? The reason I ask is `test_conversion.py` is throwing an error stating that the value is incorrect.

sssom/parsers.py
```diff
@@ -540,8 +547,14 @@ def from_alignment_minidom(
                 elif node_name == "onto2":
                     ms[OBJECT_SOURCE_ID] = e.firstChild.nodeValue
                 elif node_name == "uri1":
-                    ms[SUBJECT_SOURCE] = e.firstChild.nodeValue
+                    # import pdb; pdb.set_trace()
+                    ms[
+                        SUBJECT_SOURCE
+                    ] = (
+                        e.firstChild.nodeValue
+                    )  # CURIEfy this node value && If endswith(".extension"), replace it with _)
```
matentzn, line 555: What is your plan here re curification if an URI prefix is not in the prefix map?

matentzn (commented):
What is missing in this PR?

hrshdhgd:
> What is missing in this PR?

Just the curieficaion

sssom/parsers.py
matentzn: I would vote for leaving the URIs as is if the URI prefix is not known to the prefix map. We should produce a single warning at the end of the pipeline which lists the first 5 URIs that didn't have a proper CURIE, and log a warning "the extracted file is not valid SSSOM because some of the URIs could not be converted. Please provide appropriate URI prefixes and repeat  the process".


Charlie prefers to not process the node at all if the URI prefix is unknown; this is an alternative, but only if we can overwrite it with a flag, and then we are back to the first point.

hrshdhgd: That is the problem. The `sssom` definition for `subject_source` and `object_source` is `EntityReference`. Does this mean only CURIEs? `uriOrCurie` might be a better solution?

matentzn: I think a URI _is_ a curie, where the prefix is `http` and the URI expansion `http` so everwhere where there is an entity reference, you should be able to drop a URI.

Its fine to write a URI, just also write the warning I mention above.

sssom/util.py
```diff
@@ -1440,7 +1440,7 @@ def get_all_prefixes(msdf: MappingSetDataFrame) -> list:
                         # print(
                         #     f"Slot '{slot}' has an incorrect value: {msdf.metadata[slot]}"
                         # )
-                        raise ValidationError(
+                        logging.warning(
```
matentzn, line 1443: Why was this necessary?

sssom/validators.py
```diff
@@ -83,7 +83,7 @@ def check_all_prefixes_in_curie_map(msdf: MappingSetDataFrame) -> None:
 
     missing_prefixes = []
     for pref in prefixes:
-        if pref not in list(msdf.prefix_map.keys()):
+        if pref not in list(msdf.prefix_map.keys()) and pref != "":
```
matentzn, line 86: Shouldn't pref != "" be tested first?

matentzn (commented):
Looks great, two minor comments!

sssom/util.py
matentzn: This does not seem the right way to solve this though..  Which process adds an `object_source` that is not a valid CURIE?

hrshdhgd: [test_conversion.py](https://github.com/mapping-commons/sssom-py/blob/h2-patch3/tests/test_conversion.py#L106) <= [writers.py](https://github.com/mapping-commons/sssom-py/blob/h2-patch3/sssom/writers.py#L107) <= [validators.py](https://github.com/mapping-commons/sssom-py/blob/h2-patch3/sssom/validators.py#L73)

matentzn: None of this _adds_ an invalid object_source though that would justify this change in this PR now. 

Some observations (partially related):

1. OBJECT_SOURCE_ID does _not_ exist in SSSOM: https://github.com/mapping-commons/sssom-py/blob/master/sssom/parsers.py#L541
2. The URI parameter should validate correctly as an object_source: https://github.com/mapping-commons/sssom-py/blob/master/sssom/parsers.py#L543

Can you say what exactly the error is saying?

hrshdhgd: https://github.com/mapping-commons/sssom-py/actions/runs/5137740141/jobs/9246227909

matentzn: Got it! Proceed then, its not the "validate" methods, its something else. Thanks!

PR #381, opened 2023-06-11
Update sssom py SSSOM version to 0.12.0

matentzn:
@hrshdhgd can you deal with this release? I have done no tests etc, but there was a major change in the RDF serialiser (semapv namespace was wrong), so best run some tests?

### DIFF
--- a/sssom/util.py
+++ b/sssom/util.py
@@ -760,9 +760,9 @@ def deal_with_negation(df: pd.DataFrame) -> pd.DataFrame:
 
     # GroupBy and SELECT ONLY maximum confidence
     max_confidence_df: pd.DataFrame
-    max_confidence_df = combined_normalized_subset.groupby(
-        TRIPLE_IDS, as_index=False
-    )[CONFIDENCE].max()
+    max_confidence_df = combined_normalized_subset.groupby(TRIPLE_IDS, as_index=False)[
+        CONFIDENCE
+    ].max()
 
     # If same confidence prefer "HumanCurated".
     reconciled_df_subset = pd.DataFrame(columns=combined_normalized_subset.columns)
--- a/sssom/util.py
+++ b/sssom/util.py
@@ -100,8 +100,9 @@ SSSOM_DEFAULT_RDF_SERIALISATION = "turtle"
 
 URI_SSSOM_MAPPINGS = f"{SSSOM_URI_PREFIX}mappings"
 
-#: The 3 columns whose combination would be used as primary keys while merging/grouping
-KEY_FEATURES = [SUBJECT_ID, PREDICATE_ID, OBJECT_ID]
+#: The 4 columns whose combination would be used as primary keys while merging/grouping
+KEY_FEATURES = [SUBJECT_ID, PREDICATE_ID, OBJECT_ID, PREDICATE_MODIFIER]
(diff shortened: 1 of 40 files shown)


hrshdhgd:
There will be 1 error in tests because the main branch still needs the updated `SEMAPV` prefix map as seen [here](https://github.com/mapping-commons/sssom-py/blob/master/tests/test_parsers.py#L35). The rest of the tests pass.

hrshdhgd:
@matentzn : All tests pass!

sssom/util.py
```diff
@@ -760,9 +761,9 @@ def deal_with_negation(df: pd.DataFrame) -> pd.DataFrame:
 
     # GroupBy and SELECT ONLY maximum confidence
     max_confidence_df: pd.DataFrame
-    max_confidence_df = combined_normalized_subset.groupby(
-        KEY_FEATURES, as_index=False
-    )[CONFIDENCE].max()
+    max_confidence_df = combined_normalized_subset.groupby(TRIPLES_IDS, as_index=False)[
```
matentzn, line 764: If this `combined_normalized_subset` has no negative mappings in this context, this is ok, otherwise why not use `KEY_FEATURES`?

hrshdhgd:
@matentzn : Changed the logic a little bit:
 - Earlier I was splitting the dataframe based on the value of `predicate_modifier` and acting separately.
 - Now, I act on the dataframe as a whole (`Not` and `""` combined) while reconciling dataframes.

sssom/util.py
```diff
@@ -727,61 +727,40 @@ def deal_with_negation(df: pd.DataFrame) -> pd.DataFrame:
...
-    #     PREDICATE_ID
-    # ].str.replace("!", "")
-    ########################################################
-    normalized_negation_df = normalized_negation_df.drop(["index"], axis=1)
-
-    # remove the NOT rows from the main DataFrame
-    condition = negation_df.isin(df)
-    positive_df = df.drop(condition.index)
-    positive_df = positive_df.reset_index().drop(["index"], axis=1)
-
```
matentzn, line 746: I am very worried about this. can you describe the concrete consequences of this change? I didn't mean to prompt a redesign in this case, only to say that there very few situations in which `predicate_modifer` can be ignored as a part of the primary key..

matentzn: Trusting it now :P

tests/test_reconcile.py
```diff
@@ -28,7 +28,7 @@ def test_deal_with_negation(self):
         df1 = deal_with_negation(self.msdf1.df)
         self.assertEqual(8, len(df1.index))
         df2 = deal_with_negation(self.msdf2.df)
-        self.assertEqual(12, len(df2.index))
+        self.assertEqual(14, len(df2.index))
```
matentzn, line 31: This is what concerns me most, can you figure out what these two lines are - the update should not have change the number of rows in the reconciled part.

hrshdhgd: Yeah now I remember. ..... The negative - positive splitting was because we wanted to enforce the rule of "negative trumps positive" in case of a confidence tie. Because we are doing this, grouping by `predicate_modifier` makes no sense. Hence ~`KEY_FEATURES`~ => `TRIPLE_IDS`

hrshdhgd:
@matentzn : Ready again!

matentzn (commented):
approved (cant approve as its my PR)

PR #382, opened 2023-06-13
`sssom` using cookiecutter

### LINKED ISSUE #375: Replace linkml dependency with linkml-runtime dependency
In general there is no need for most python modules including this one to include linkml as a full dependency. LinkML is typically only needed for the build part of the lifecycle, same as all the other dev dependencies.

also why isn't this repo using poetry?

hrshdhgd:
- [x] Uses `poetry`: fixes #375
- [x] Generated from monarch-cookiecutter.
- [x] Removed reference to `linkml` and replaced it with `linkml-runtime` Fixes #375 
- [x] Uses context that comes packaged within `sssom-schema`. 
- [x] Corrected `author_id`: ~"cjm"~ => "orcid:1234"


src/sssom/validators.py
```diff
@@ -4,8 +4,11 @@
 from typing import List
 
 from jsonschema import ValidationError
-from linkml.validators.jsonschemavalidator import JsonSchemaDataValidator
-from linkml.validators.sparqlvalidator import SparqlDataValidator  # noqa: F401
+
+# from linkml.validators.jsonschemavalidator import JsonSchemaDataValidator
+# from linkml.validators.sparqlvalidator import SparqlDataValidator  # noqa: F401
+from linkml_runtime.processing.referencevalidator import ReferenceValidator
```
matentzn, line 10: What is ReferenceValidator?

tests/data/bad_basic.tsv
```diff
@@ -1,7 +1,7 @@
 #license: "https://creativecommons.org/publicdomain/zero/1.0/"
 #mapping_set_id: https://w3id.org/sssom/mapping/tests/data/basic3.tsv
 #mapping_tool: "https://github.com/cmungall/rdf_matcher"
-#creator_id: "cjm"
+#creator_id: "orcid:1234"
```
matentzn, line 4: GOOD!

tests/data/basic4.tsv
```diff
@@ -1,7 +1,7 @@
 #license: "https://creativecommons.org/publicdomain/zero/1.0/"
 #mapping_set_id: https://w3id.org/sssom/mapping/tests/data/basic4.tsv
 #mapping_tool: "https://github.com/cmungall/rdf_matcher"
-#creator_id2: "cjm"
+#creator_id2: "orcid:1234"
```
matentzn, line 4: creator_id2 should not be allowed in header - right now validation of non defined properties is pretty lax.

matentzn: (this does not have to solved in this PR!)

tests/data/pato.json
```diff
@@ -78610,7 +78610,7 @@
             "basicPropertyValues": [
               {
                 "pred": "http://www.geneontology.org/formats/oboInOwl#created_by",
-                "val": "cjm"
+                "val": "orcid:1234"
```
matentzn, line 78613: This should not be fixed (pato.json), but instead _skipped_ by the obographs->sssom parser. No need to fix now, but make an issue to that end.

tests/test_cli.py
```diff
@@ -372,5 +362,5 @@ def test_convert_cli(self) -> Result:
             "--output-format",
             "owl",
         ]
-        result = subprocess.run(command)  # noqa
+        result = subprocess.run(command, shell=True)  # noqa
```
matentzn, line 365: Why was this change necessary?

tox.ini
```diff
@@ -4,79 +4,103 @@
...
+description = Run the flake8 tool with several plugins (bandit, docstrings, import order, pep8 naming).
+
+#########################
+# Flake8 Configuration  #
+# (.flake8)             #
+#########################
+[flake8]
+max-line-length=120
+ignore =
+    DAR101 # Missing parameter(s) in Docstring: - with_git_hash
```
matentzn, line 76: Isn't this a bit central? How can we enforce this moving forward? Can we autogenerate docstrings for everything and leave it empty, so that in the future we have that check active?

sssom/__init__.py
```diff
@@ -1,26 +0,0 @@
...
-
-from .util import (  # noqa:401
-    collapse,
-    compare_dataframes,
-    dataframe_to_ptable,
-    filter_redundant_rows,
-    group_mappings,
-    parse,
-    reconcile_prefix_and_data,
-)
```
matentzn, line 26: It is slightly dangerous that these are no longer in _init_. Is there any way we can make it safe? Now, all code that will do 

```
import reconcile_prefix_and_data from sssom
```

Will break wo a warning..

matentzn (commented):
FANTASTIC effort, I am so glad!! Thanks thanks!!

I made some comments!

Also let's get a @souzadevinicius review on this one, maybe he sees something I missed.

src/sssom/validators.py
hrshdhgd: Validates a reference (schema). Takes `SchemaView` as an input

tests/data/pato.json
hrshdhgd: I did a complete find-replace for this substitution.

tox.ini
hrshdhgd: This error shows up in spite of parameters being present in the docstring. Not sure what it expects.

hrshdhgd: removing these `ignores is a PR in itself. Could be assigned to someone who I wants to learn linting.

sssom/__init__.py
hrshdhgd: Added them back.

src/sssom/validators.py
matentzn: Does this cover all of what the JSON schema checker was doing previously? What else does it do? Do you know why it was renamed?

tests/test_cli.py
matentzn: Gr8 thx

src/sssom/validators.py
hrshdhgd: `ReferenceValidator` is a generalized validator that validates whatever is passed through and automatically detects the format. Hence it substitutes JsonSchema and SPARQL.

matentzn: Ok, I read all of it and don't understand it. If you understand it well enough to know that this change actually does at least as many checks as the previous JSON-specific thing, I am fine. RIght now, I am not clear wether that reference implementation actually runs all the JSON schema checks...

matentzn (approved):

src/sssom/validators.py
hrshdhgd: [test_validate.py](https://github.com/mapping-commons/sssom-py/blob/sssom-using-cookiecutter/tests/test_validate.py#L23-L39) confirms that it works.

tests/test_parsers.py
```diff
@@ -32,7 +32,9 @@ class TestParse(unittest.TestCase):
     def setUp(self) -> None:
         """Set up the test case."""
         # TODO: change back to the commented url.
-        self.df_url = "https://raw.githubusercontent.com/mapping-commons/sssom-py/master/tests/data/basic.tsv"
+        self.df_url = (
```
souzadevinicius, line 35: It could be replaced by a constant to improve readability

PR #385, opened 2023-06-23
Change parse methods to support StringIO

matentzn:
This ended up doing a bit more than just adding support for StringIO:

- Localising all the file processing to a private `_open_input()` method. No other function will deal with opening the filestream, and the target format is io.StreamIO
- Providing a single method _separate_metadata_and_table_from_stream to take in the full stream for an sssom mapping table and splitting it into the metadata and dataframe component, before parsing it with pandas/yaml.
- Simplified a few tests and removed the broken `to_dataframe()` method (which is never needed).

### DIFF
--- a/src/sssom/util.py
+++ b/src/sssom/util.py
@@ -871,7 +871,7 @@ def get_file_extension(file: Union[str, Path, TextIO]) -> str:
             return f_format.strip(punctuation)
         else:
             logging.warning(f"Cannot guess format from {filename}")
-    logging.info(f"Cannot guess format extension for this file, assuming TSV.")
+    logging.info("Cannot guess format extension for this file, assuming TSV.")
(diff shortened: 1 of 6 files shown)


src/sssom/util.py
```diff
@@ -923,23 +930,24 @@ def read_metadata(filename: str) -> Metadata:
...
-        if extension == "tsv":
-            sep = "\t"
-        elif extension == "csv":
-            sep = ","
-        else:
-            sep = "\t"
-            logging.warning("Cannot automatically determine table format, trying tsv.")
-        df = read_csv(file, comment="#", sep=sep).fillna("")
+    sep_new = sep
+    if set is None:
```
hrshdhgd, line 942: `set` or `sep`?

PR #387, opened 2023-06-28
Add reports to validators

matentzn:


matentzn:
Replaced by https://github.com/mapping-commons/sssom-py/pull/662

PR #388, opened 2023-06-29
Add default `confidence` param to sssom file while creating `ptable`

hrshdhgd:
If the confidence column is absent, `ptable` cannot be generated.

Fix: Add the `confidence` column if absent and initialize it to a user provided value. If present, then fill `None`s by that value

### DIFF
--- a/src/sssom/cli.py
+++ b/src/sssom/cli.py
@@ -49,8 +49,6 @@ from .util import (
     SSSOM_EXPORT_FORMATS,
     SSSOM_READ_FORMATS,
     MappingSetDataFrame,
-    add_default_confidence,
-    collapse,
     compare_dataframes,
     dataframe_to_ptable,
     filter_redundant_rows,
--- a/src/sssom/cli.py
+++ b/src/sssom/cli.py
@@ -254,7 +254,7 @@ def split(input: str, output_directory: str):
 @click.option("-W", "--inverse-factor", help="Inverse factor.")
 @click.option(
     "--default-confidence",
-    default=None,
+    type=float,
     help="Default confidence to be assigned if absent.",
 )
 def ptable(input, output: TextIO, inverse_factor: float, default_confidence: float):
--- a/src/sssom/cli.py
+++ b/src/sssom/cli.py
@@ -24,6 +24,7 @@ from rdflib import Graph
 from scipy.stats import chi2_contingency
 
 from sssom.constants import (
+    CONFIDENCE,
     DEFAULT_VALIDATION_TYPES,
     PREFIX_MAP_MODES,
     SchemaValidationType,
@@ -49,6 +50,7 @@ from .util import (
     SSSOM_EXPORT_FORMATS,
     SSSOM_READ_FORMATS,
     MappingSetDataFrame,
+    add_default_confidence,
     collapse,
     compare_dataframes,
     dataframe_to_ptable,
@@ -266,6 +268,7 @@ def ptable(input, output: TextIO, inverse_factor, default_confidence: bool):
     # TODO should maybe move to boomer (but for now it can live here, so cjm can tweak
     msdf = parse_sssom_table(input)
     # df = parse(input)
+    msdf.df = add_default_confidence(msdf.df)
     df = collapse(msdf.df)
     # , priors=list(priors)
     rows = dataframe_to_ptable(
--- a/src/sssom/cli.py
+++ b/src/sssom/cli.py
@@ -268,7 +268,9 @@ def ptable(input, output: TextIO, inverse_factor: float, default_confidence: flo
         msdf.df = add_default_confidence(msdf.df, default_confidence)
     df = collapse(msdf.df)
     # , priors=list(priors)
-    rows = dataframe_to_ptable(df, inverse_factor=inverse_factor, confidence=default_confidence)
+    rows = dataframe_to_ptable(
+        df, inverse_factor=inverse_factor, default_confidence=default_confidence
+    )
     for row in rows:
         print(*row, sep="\t", file=output)
 
--- a/src/sssom/cli.py
+++ b/src/sssom/cli.py
@@ -253,14 +253,24 @@ def split(input: str, output_directory: str):
 @input_argument
 @output_option
 @click.option("-W", "--inverse-factor", help="Inverse factor.")
-def ptable(input, output: TextIO, inverse_factor):
+@click.option(
+    "--default-confidence / --no-default-confidence",
+    default=False,
+    is_flag=True,
+    help="If True and SSSOM file does not have the `confidence` column,\
+          it is created and initialized to 0.95. But if the `confidence` column exists,\
+          only `None` values are imputed to 0.95.",
+)
+def ptable(input, output: TextIO, inverse_factor, default_confidence: bool):
     """Convert an SSSOM file to a ptable for kboom/`boomer <https://github.com/INCATools/boomer>`_."""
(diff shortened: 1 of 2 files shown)


src/sssom/cli.py
```diff
@@ -253,14 +254,25 @@ def split(input: str, output_directory: str):
 @input_argument
 @output_option
 @click.option("-W", "--inverse-factor", help="Inverse factor.")
-def ptable(input, output: TextIO, inverse_factor):
+@click.option(
+    "--default-confidence / --no-default-confidence",
+    default=False,
+    is_flag=True,
```
matentzn, line 260: 1. Dont make it a flag, just make it 

```
--confidence X
```

where X is a number, e.g.

```
--confidence 0.9
```

Much more flexible, especially if you deal with mapping sets you trust differently!

Also, I thought about it and we can catch two flies with one here: Lets say this is the confidence of the person processing the mapping. We can use this not only to fill in missing confidence, but perhaps (in a second step, not this PR) to re-calculate existing confidence values.

So lets say we have a mapping that already has 0.8 confidence. If the mapping user provides `--confidence 0.9` then we could recalcularte the confidence like this `0.8*0.9=0.71` I think that is useful for many use cases..

src/sssom/util.py
```diff
@@ -504,17 +504,38 @@ def compare_dataframes(df1: pd.DataFrame, df2: pd.DataFrame) -> MappingSetDiff:
...
+    :param df: DataFrame whose `confidence` column needs to be filled.
+    :return: DataFrame with a complete `confidence` column.
+    """
+    df[CONFIDENCE] = df.get(CONFIDENCE, 0.95)
+    df.loc[df[CONFIDENCE].isnull(), CONFIDENCE] = 0.95
+    return df
+
+
+def dataframe_to_ptable(
+    df: pd.DataFrame, *, inverse_factor: float = None, default_confidence: bool = False
```
matentzn, line 521: Rename to just "confidence"?

hrshdhgd: I'll rename to `default_confidence` since there's already a variable named `confidence` down below in the same function.

src/sssom/cli.py
hrshdhgd: So in other words if an external confidence is provided (say 0.9 in your example) and existing confidence reduces (0.72)?

hrshdhgd: I can do recalculation in this PR itself, it's very simple!

src/sssom/cli.py
```diff
@@ -253,14 +254,23 @@ def split(input: str, output_directory: str):
...
+    default=None,
+    help="Default confidence to be assigned if absent.",
+)
+def ptable(input, output: TextIO, inverse_factor: float, default_confidence: float):
     """Convert an SSSOM file to a ptable for kboom/`boomer <https://github.com/INCATools/boomer>`_."""
     # TODO should maybe move to boomer (but for now it can live here, so cjm can tweak
     msdf = parse_sssom_table(input)
     # df = parse(input)
+    if default_confidence:
+        msdf.df = add_default_confidence(msdf.df, default_confidence)
```
matentzn, line 268: Why is this added twice? Once here, and once in `dataframe_to_ptable`?

src/sssom/util.py
```diff
@@ -504,17 +504,39 @@ def compare_dataframes(df1: pd.DataFrame, df2: pd.DataFrame) -> MappingSetDiff:
...
+
+    If `confidence` column already exists, only fill in the None ones by 0.95.
+
+    :param df: DataFrame whose `confidence` column needs to be filled.
+    :return: DataFrame with a complete `confidence` column.
+    """
+    # df[CONFIDENCE] = df.get(CONFIDENCE, confidence)
+    if df.get(CONFIDENCE) is not None:
+        df[CONFIDENCE] = confidence * df[CONFIDENCE]
+    df.loc[df[CONFIDENCE].isnull(), CONFIDENCE] = confidence
```
matentzn, line 518: Is this doing the exact same thing as this (which is how I thought of the problem):

```
   if 'confidence' not in df.columns:
        df['confidence'] = confidence
    else:
        df['confidence'] = np.where(
            (df['confidence'].between(0, 1)) & (pd.notnull(df['confidence'])),
            df['confidence'] * confidence,
            confidence
        )
```

src/sssom/cli.py
hrshdhgd: Yeah I had that question too... this is how it was from the beginning. Let me streamline it.

hrshdhgd: Done

src/sssom/util.py
hrshdhgd: Yes

src/sssom/cli.py
```diff
@@ -253,14 +252,18 @@ def split(input: str, output_directory: str):
...
+    "--default-confidence",
+    default=None,
+    help="Default confidence to be assigned if absent.",
+)
+def ptable(input, output: TextIO, inverse_factor: float, default_confidence: float):
     """Convert an SSSOM file to a ptable for kboom/`boomer <https://github.com/INCATools/boomer>`_."""
     # TODO should maybe move to boomer (but for now it can live here, so cjm can tweak
     msdf = parse_sssom_table(input)
-    # df = parse(input)
-    df = collapse(msdf.df)
```
matentzn, line 261: ? Why is collapse removed?

matentzn: Because this already happens in `dataframe_to_ptable`.

src/sssom/cli.py
matentzn: Great.

src/sssom/cli.py
```diff
@@ -253,14 +252,18 @@ def split(input: str, output_directory: str):
 @input_argument
 @output_option
 @click.option("-W", "--inverse-factor", help="Inverse factor.")
-def ptable(input, output: TextIO, inverse_factor):
+@click.option(
+    "--default-confidence",
+    default=None,
```
matentzn, line 257:
```suggestion
    default=None,
    type=click.FloatRange(0, 1),
```

Not sure if that works, maybe check

src/sssom/util.py
```diff
@@ -504,17 +504,40 @@ def compare_dataframes(df1: pd.DataFrame, df2: pd.DataFrame) -> MappingSetDiff:
...
+def add_default_confidence(df: pd.DataFrame, confidence: float = None) -> pd.DataFrame:
+    """Add `confidence` column to DataFrame if absent and initializes to 0.95.
+
+    If `confidence` column already exists, only fill in the None ones by 0.95.
+
+    :param df: DataFrame whose `confidence` column needs to be filled.
+    :return: DataFrame with a complete `confidence` column.
+    """
+    if CONFIDENCE in df.columns:
+        df[CONFIDENCE] = confidence * df.get(CONFIDENCE, 1)
```
matentzn, line 516: What happens if there is a 

NaN (a non numeric string)
None or ""
Or NULL

In the `confidence` column?

src/sssom/cli.py
hrshdhgd: Exactly

src/sssom/cli.py
hrshdhgd: I've changed it as per click documentation

src/sssom/util.py
hrshdhgd: Good catch! Fixed it

matentzn (dismissed):
Fantastic effort! THANK YOU!

PR #389, opened 2023-07-07
Predicate modifier fix

hrshdhgd:
`predicate_modifier` is a optional column but is essential for `groupBy` while removing mappings etc. This adds a blank column if the column does not exist. This PR is in response to https://github.com/monarch-initiative/mondo-ingest/issues/343

### DIFF
--- a/.github/workflows/update_context.yml
+++ b/.github/workflows/update_context.yml
@@ -17,7 +17,9 @@ jobs:
         run: make all_schema
 
       - name: Update sssom context
-        run: make deploy-dm
+        run: |
+          mkdir -p sssom
+          make deploy-dm
 
       - name: Create Pull Request
         uses: peter-evans/create-pull-request@v3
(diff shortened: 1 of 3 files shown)


matentzn (commented):
Ok, in "remove mappings" this makes some sense, but before I approve: isn't it better to fix the code that assumes that `PREDICATE_MODIFIER` MUST be present, rather than injecting an empty column?

hrshdhgd:
This error came into being after we included `predicate_modifier` into `KEY_FEATURES`. You mean something like https://github.com/mapping-commons/sssom-py/pull/389/commits/06a227c318bf33b219e5810f9cbfba336cb34e19

matentzn:
No I am thinking more to remove predicate_modifier from KEY_FEATURES everytime KEY_FEATURES is used as a means to subset a dataframe.. Or something along this line. I don't know exactly what is right here - from a user perspective, I just don't want empty `predicate_modifier` columns to show up in my data frames when I am processing them with sssom py.

matentzn (approved):

PR #391, opened 2023-07-14
`scipy` latest version creating incompatibility errors.

hrshdhgd:
I switched `sssom-py` to using any version of python >= 3.8 but < 4.0.0 for now, (it was <3.12 earlier). Due to this, poetry complained about `scipy` having conflicting python version compatibility. 

Solution `poetry add scipy="*"`: What this does is finds a compatible version of `scipy` and installs it. This will fix some headache in the future.

For now this will fix the following problem pin ODK:

```
The current project's Python requirement (>=3.8,<4.0.0) is not compatible with some of the required packages Python requirement:
  - sssom requires Python >=3.8,<3.12, so it will not be satisfied for Python >=3.12,<4.0.0

Because sssom (0.3.35) requires Python >=3.8,<3.12
 and no versions of sssom match >0.3.35,<0.4.0, sssom is forbidden.
So, because test depends on sssom (^0.3.35), version solving failed.
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -11,7 +11,7 @@ license = "MIT"
 readme = "README.md"
 
 [tool.poetry.dependencies]
-python = ">=3.8,<3.12"
+python = "^3.8"
 click = "*"
 importlib-metadata = ">=4.8.0"
 pyparsing = "^3.0.9"
@@ -26,7 +26,7 @@ rdflib = "^6.3.2"
 sparqlwrapper = "^2.0.0"
 sssom-schema = "^0.13.0"
 validators = "^0.20.0"
-scipy = "^1.10.1"
+scipy = "*"
 
 [tool.poetry.group.dev.dependencies]
 pytest = {version = ">=7.1.2"}
--- a/poetry.lock
+++ b/poetry.lock
@@ -63,20 +63,20 @@ lxml = ["lxml"]
 
 [[package]]
 name = "bioregistry"
-version = "0.9.51"
+version = "0.9.75"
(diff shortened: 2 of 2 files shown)


pyproject.toml
```diff
@@ -11,7 +11,7 @@ license = "MIT"
 readme = "README.md"
 
 [tool.poetry.dependencies]
-python = ">=3.8,<3.12"
+python = "^3.8"
```
matentzn, line 14: Will this allow 3.12?

matentzn: (in other words, why not `>=3.8,<4.0.0`)

matentzn (approved):

matentzn:
See my question before merging, else LGTM

pyproject.toml
hrshdhgd: Yes it will allow 3.12.x. This means exactly that `<=3.8, <4.0.0`

cmungall:
Why not fix all the other `^` dependencies to '>='. As a core library that
is reused in multiple other projects, sssom-py simply has to be less
opinionated about versions.

Do we even need scipy as a core dependency?

I think I am responsible for introducing this:
https://github.com/mapping-commons/sssom-py/blame/master/src/sssom/cli.py#L488C32-L488C32

Back when sssom-py was a bit more experimental. But if people are really
using this non-core function perhaps this could be an extra.

We really need to be making our core python libraries more lightweight.
Note every design decision from 3+ years ago is set in stone. We can start
moving out non-core functionality into separate packages, plugins, or make
the dependency an extra.

Issue #363, opened 2023-04-04
Replace CURIE inference mechanism with curies.Converter.from_extended_prefix_map


matentzn:
Right now, we are using the bioregistry bimap, and old bioregistry tooling for getting from CU:RIE string to URIs and back.

@cthoyt has now introduced the concept of "extended prefix maps" which are much more powerful for guessing and standardising prefixes then the bimap.

Here is a snippet:

```python
import curies

#: OBO context's extended prefix map
url = "https://github.com/biopragmatics/bioregistry/raw/main/exports/contexts/obo.epm.json"
converter = curies.Converter.from_extended_prefix_map(url)

>>> converter.expand("GO:1234567")
'http://purl.obolibrary.org/obo/GO_1234567'

>>> converter.compress("http://purl.obolibrary.org/obo/GO_1234567")
'GO:1234567'
>>> converter.compress("http://amigo.geneontology.org/amigo/term/GO:1234567")
'GO:1234567'

>>> converter.standardize_prefix("gomf")
'GO'
>>> converter.standardize_curie"gomf:1234567")
'GO:1234567'
>>> converter.standardize_uri("http://amigo.geneontology.org/amigo/term/GO:1234567")
'http://purl.obolibrary.org/obo/GO_1234567'
```

We should replace the old way of doing this with the new one.

For now, I still want to mirror the prefix map as part of the package rather than relying on bioregistry to doing the hosting/versioning.

matentzn:
This is a mid-priority refactoring that can be done as a subsequent PR to #396

matentzn:
As per discussion in #396, we can replace everything, including the built-in prefix map, with `curies`. 

Hwoever, I would like to ensure we have a test in place in the testing framework that ensures that whenever the epm is updated, the built-in prefixes are exactly what we expect them to be (which is **what is in sssom context**).

cthoyt:
@matentzn this is effectively done besides #431

PR #397, opened 2023-07-22
Begin using `curies.Converter` in more places

cthoyt:
Part of #363 

This PR does the following:

1. Adds a minimum version of `curies` that has the strict compress and expand functions
2. Rewrites the SPARQL utils and RDF utils to use `curies` functionality
3. Updates custom `curie_from_uri` to use `curies` (will make a follow-up PR that replaces this completely)


src/sssom/cli.py
```diff
@@ -356,11 +356,8 @@ def sparql(
         endpoint.limit = limit
     if object_labels is not None:
         endpoint.include_object_labels = object_labels
-    if prefix is not None:
-        if endpoint.prefix_map is None:
-            endpoint.prefix_map = {}
-        for k, v in prefix:
-            endpoint.prefix_map[k] = v
+    for k, v in prefix or []:
```
matentzn, line 359: Nice syntax!

src/sssom/sparql_util.py
```diff
@@ -59,7 +64,8 @@ def query_mappings(config: EndpointConfig) -> MappingSetDataFrame:
         cols.insert(-1, "object_label")
     colstr = " ".join([f"?{c}" for c in cols])
     olq = "OPTIONAL { ?object_id rdfs:label ?object_label }" if config.include_object_labels else ""
-    q = f"""\
+    sparql = dedent(
```
matentzn, line 67: Cool method, textwrap.dedent!

src/sssom/sparql_util.py
```diff
@@ -72,56 +78,16 @@ def query_mappings(config: EndpointConfig) -> MappingSetDataFrame:
...
-    """Convert all values in the dict from URIs to CURIEs.
-
-    :param row: A dictionary of string keys to URIs
-    :param config: Configuration
-    :return: A dictionary of string keys to CURIEs
-    """
-    return {k: contract_uri(v, config) for k, v in row.items()}
-
-
-def contract_uri(uri: str, config: EndpointConfig) -> str:
```
matentzn, line 99: I am so glad this is going away.

src/sssom/util.py
```diff
@@ -1107,13 +1107,12 @@ def get_prefix_from_curie(curie: str) -> str:
         return ""
 
 
-def curie_from_uri(uri: str, prefix_map: Mapping[str, str]) -> str:
+def curie_from_uri(uri: str, prefix_map: Union[Mapping[str, str], Converter]) -> str:
```
matentzn, line 1110: Add deprecation annotation?

src/sssom/util.py
```diff
@@ -1126,21 +1125,10 @@ def curie_from_uri(uri: str, prefix_map: Mapping[str, str]) -> str:
...
-            if is_curie(curie):
-                return f"{prefix}:{remainder}"
-            else:
-                logging.warning(f"{prefix}:{remainder} is not a CURIE ... skipping")
-                continue
-    raise NoCURIEException(f"{uri} does not follow any known prefixes")
+    warnings.warn("Use safe_compress() instead", DeprecationWarning)
+    if not isinstance(prefix_map, Converter):
+        converter = Converter.from_prefix_map(prefix_map)
+    return safe_compress(uri, converter)
```
matentzn, line 1131: What happens if `prefix_map` _is_ an instance of `converter`? I mean, this case should basically never happen, but shouldn't there be an

```
else: converter = prefix_map 
```

There?

src/sssom/util.py
```diff
@@ -1601,3 +1589,18 @@ def invert_mappings(
...
+
+    :param uri: The URI to parse. If this is already a CURIE, return directly.
+    :param converter: Converter used for compression
+    :return: A CURIE
+    """
+    if not is_curie(uri):
+        return converter.compress_strict(uri)
+    rv = converter.standardize_curie(uri)
+    if rv is None:
+        raise ValueError
```
matentzn, line 1605: Is there a reason the ValueError is not a bit more verbose?

matentzn (dismissed):
Looks awesome!! THANK YOU!

@hrshdhgd I will leave it to you to deal with the PR - I left 1/2 comments here and there to check, but nothing critical.

src/sssom/util.py
cthoyt: oop this is a typo, thank you for the catch.

src/sssom/util.py
cthoyt: Fixed in 7e19639

src/sssom/util.py
cthoyt: Fixed in ec3452f

src/sssom/util.py
cthoyt: I wasn't aware of this, thanks for the suggestion. Done in 118017d.

cthoyt:
@matentzn @hrshdhgd it appears the issues in the build now are due to poetry. I will again give a gentle nudge to totally drop poetry, it is way more of an issue than it helps

hrshdhgd:
There are 4 errors as of now. First one is a CLI error which will be fixed automatically once the following are fixed.

Three pytest errors occurring as of now:
```
FAILED tests/test_parsers.py::TestParse::test_parse_alignment_minidom - curies.api.DuplicateURIPrefixes: Duplicate URI prefixes:
FAILED tests/test_parsers.py::TestParse::test_parse_obographs - curies.api.DuplicateURIPrefixes: Duplicate URI prefixes:
FAILED tests/test_parsers.py::TestParse::test_parse_sssom_rdf - ValueError: CURIE appeared where there should be a URI, and could not be standardized: oio:hasBroadSynonym
```

From what I can tell the first two errors are thrown by `curies/api.py`.

The third error is because the prefix `oio:` is not in the `prefix_map`.

hrshdhgd:
>it appears the issues in the build now are due to poetry. I will again give a gentle nudge to totally drop poetry, it is way more of an issue than it helps

I agree but it seems to me that only GitHub environments seem to be problematic. Locally, it seems to work fine. Plus Chris insists all our projects following a standard architecture, hence my hands are sort of tied 🙂 .

cthoyt:
@hrshdhgd thanks for following up on all of this. can you take care of looking into these tests that have inconsistent data? Great to see that we are implicitly now checking data integrity using `curies`!

hrshdhgd:
I've fixed the `poetry` issue @cthoyt . You may be in a position to fix these errors quicker than I and I have other higher priorities to tend to atm. I'd really appreciate if you could complete your PR.

src/sssom/context.py
```diff
@@ -86,6 +86,8 @@ def add_built_in_prefixes_to_prefix_map(
                 raise ValueError(
                     f"Built-in prefix {k} is specified ({prefix_map[k]}) but differs from default ({builtinmap[k]})"
                 )
+    if "dc" in prefix_map and "dcterms" in prefix_map and prefix_map["dc"] == prefix_map["dcterms"]:
+        del prefix_map["dc"]
```
matentzn, line 90: I dont know who wrote this but this seems misplaced. Why is this here?

src/sssom/sparql_util.py
```diff
@@ -26,13 +28,16 @@ class EndpointConfig:
...
-    prefix_map: Optional[Dict[str, str]]
     include_object_labels: bool = False
+    prefix_map: Dict[str, str] = field(default_factory=dict)
 
 
 def query_mappings(config: EndpointConfig) -> MappingSetDataFrame:
     """Query a SPARQL endpoint to obtain a set of mappings."""
-    sparql = SPARQLWrapper(config.url)
+    if not config.prefix_map:
+        raise TypeError
```
matentzn, line 38: Maybe a more expressive error message here

cthoyt: Done in 1c25c94, but I inherited this not having a useful error message from before.

src/sssom/context.py
cthoyt: See also https://github.com/mapping-commons/sssom/pull/302 for a potential upstream fix. I don't think this should be a blocker for this PR, since prefix map loading isn't very principled in sssom-py already, I don't see any harm in adding some additional specific hacks. A future PR can switch over to using `curies.Converter` natively to handle prefix map loading and support synonyms.

matentzn: Yea, these entries should not exist in the sssom schema context. We want to move away from them entirely. 

I would prefer to wait for the ssssom schema release and then get rid of this dc superfluous code..

matentzn: This is for now necessary because the LinkML context generator automatically adds those prefixes.

src/sssom/context.py
```diff
@@ -108,8 +122,12 @@ def get_default_metadata() -> Metadata:
             if "@id" in v and "@prefix" in v:
                 if v["@prefix"]:
                     prefix_map[key] = v["@id"]
+    del prefix_map["@vocab"]
 
     prefix_map.update({(k, v) for k, v in contxt_external.items() if k not in prefix_map})
+    clean_dc(prefix_map)
+    # Tests if the prefix map is a valid bijective map
+    Converter.from_prefix_map(prefix_map)
```
matentzn, line 130: How does this line of code affect the rest of the code? It seems a bit disconnected?

tests/test_parsers.py
```diff
@@ -63,6 +64,7 @@ def setUp(self) -> None:
         self.alignmentxml_file = f"{test_data_dir}/oaei-ordo-hp.rdf"
         self.alignmentxml = minidom.parse(self.alignmentxml_file)
         self.metadata = get_default_metadata()
+        Converter.from_prefix_map(self.metadata.prefix_map)
```
matentzn, line 67: How does this line of code work?

src/sssom/context.py
cthoyt: it will throw an error if the prefix map is not bijective, I can update the comment to make this more clear

matentzn: Ah cool. No this makes sense now!

tests/test_parsers.py
matentzn: Same as above.

cthoyt: I added a more explicit helper function for checking prefix map validity in 439f561

src/sssom/context.py
cthoyt: I added a more explicit helper function for checking prefix map validity in 439f561

tests/test_parsers.py
```diff
@@ -115,12 +117,15 @@ def test_parse_obographs(self):
...
-
+        # TODO this number seems arbitrary versus having proper tests
+        #  this can be fixed later
+        # self.assertEqual(
+        #     len(msdf.df),
+        #     9881,
+        #     f"{self.obographs_file} has the wrong number of mappings.",
+        # )
+
+    @unittest.skip(reason="Not sure what is broken in this graph")
```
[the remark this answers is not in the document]
cthoyt: looking into it

.gitignore
```diff
@@ -158,3 +158,4 @@ schema/sssom_datamodel.py
 sssom/internal_context.py
 sssom/sssom_datamodel.py
 *.rej
+poetry.lock
```
cthoyt, line 161: @hrshdhgd if we're giving up on poetry locks, can we put that into a different PR? I don't want people saying later that I snuck this in :)

src/sssom/context.py
cthoyt: We actually fixed this the right way with the SSSOM Schema v0.14, I removed related code for checking/fixing DC/DCTERMS in e24fec2.

tests/test_parsers.py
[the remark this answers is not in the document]
cthoyt: so yes, I have now added a detailed explanation of why this test is not needed. We should keep it skipped, or even better, remove it.

.gitignore
hrshdhgd: Fair enough. The tests will fail for now but I think I have a solution.

cthoyt: Yes, I mean we can solve the problem in a separate PR, merge it into master, then just update this PR on top of it. Thanks!

hrshdhgd: Done: https://github.com/mapping-commons/sssom-py/pull/399

cthoyt:
@hrshdhgd it appears this PR is passing, even with the Poetry changes removed.

hrshdhgd:
> @hrshdhgd it appears this PR is passing, even with the Poetry changes removed.

Yes , I had a quick solution that worked!

hrshdhgd:
I merged #399 to main. Now if you merge the main branch down to this one, all should be well. I'll let you do it since you're actively working on this branch.

tests/test_parsers.py
```diff
@@ -117,21 +118,10 @@ def test_parse_obographs(self):
             write_table(msdf, file)
         self.assertEqual(
             len(msdf.df),
-            9881,
+            8099,
```
matentzn, line 121: @cthoyt I reinstated this test. Analysing the warnings, I noticed that the 800 less parsed entities (which is good, since they are not in the prefix map) are due to https://github.com/biopragmatics/bioregistry/issues/917. 

I think this is entirely independent of this PR, so I will approve this, this is just FYI.

matentzn (approved):
I have convinced myself that deleting that broken obographs test was the right thing to do! Thanks for your work on this, I am happy now; the one comment is purely an FYI, the behaviour is desireable.

matentzn:
@hrshdhgd ready for your review and merge, I am happy with it!

hrshdhgd (approved):
Looks much cleaner than before! Thanks @cthoyt !

PR #400, opened 2023-07-27
Add poetry.lock back to versioning and trying minimal dependency

hrshdhgd:
- [x] Added `poetry.lock` back to version control
- [ ] ~Removed pinning of poetry version in GitHub Actions~
- [ ] ~Pruned dependency list to the minimal required one based on the dependencies that already bring in other packages that are needed.~
- [x] figure out why the error was occurring:
  - So `linkml-runtime` has a dependency on `prefixcommons` which has a dependency on `pytest-logging`. This package (pytest-logging) is causing issues with the windows instance within GitHub Actions. 
  - I do not know why this is **not** happening in any of the other projects we have.
  - The fix in `qc.yml` is : 
  ```
    poetry run pip wheel --use-pep517 "pytest-logging (==2015.11.4)"
  ```
    before `poetry install`

 - Another issue was with Windows environment not detecting `poetry` as a command in spite of being installed.
   - Reason: `poetry` path was not automatically being added to environment variables when it is invoked using an reusable gh-action ( `snok/install-poetry@v1`).
   - Fix:
    ```
    - name: Windows specific step.
            if: matrix.os == 'windows-latest'
            run: echo "${HOME}/.local/bin" | Out-File -FilePath $env:GITHUB_PATH -Encoding utf8 -Append

    ```
 - `pyparsing` is not really used in the project as far as I know. It is also causing errors in GH Actions. So removing it 

Note: `poetry` is still pinned to version 1.3.2 as before.


pyproject.toml
```diff
@@ -12,21 +12,15 @@ readme = "README.md"
...
-bioregistry = ">=0.9.43"
-deprecation = ">=2.1.0"
-linkml-runtime = ">=1.5.3"
+linkml-runtime = ">=1.5.5"
+importlib-metadata = ">=6.8.0"
+pansql = ">=0.0.1"
+sssom-schema = ">=0.14.0"
+bioregistry = ">=0.9.87"
 networkx = ">=3.1"
-curies = ">=0.5.7"
```
cthoyt, line 22: I don't think this makes total sense - `curies` is a transitive dependency of other dependencies of sssom-py, but we are using 0.5.7+ only features, so this needs to be annotated. I think the rule of thumb is you should mark everything you explicitly import as explicit dependencies

matentzn: Same for pandas pyyaml (and perhaps others) - whenever we directly use a library we should add it here.

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "sssom"
-version = "0.3.33rc1.post1.dev0+968123b"
+version = "0.0.0"
```
matentzn, line 3: Is this safe?

pyproject.toml
```diff
@@ -12,21 +12,15 @@ readme = "README.md"
...
-importlib-metadata = ">=4.8.0"
-pyparsing = "^3.0.9"
-bioregistry = ">=0.9.43"
-deprecation = ">=2.1.0"
-linkml-runtime = ">=1.5.3"
+linkml-runtime = ">=1.5.5"
+importlib-metadata = ">=6.8.0"
+pansql = ">=0.0.1"
+sssom-schema = ">=0.14.0"
+bioregistry = ">=0.9.87"
```
matentzn, line 19: Id this needed?

.github/workflows/qc.yml
```diff
@@ -22,13 +22,37 @@ jobs:
...
-          pip install --upgrade pip
-          pip install poetry==1.3.2
+        uses: snok/install-poetry@v1
+        with:
+          version: 1.3.2
 
+      #----------------------------------------------
+      #       load cached venv if cache exists
+      #----------------------------------------------
+      - name: Load cached venv
```
matentzn, line 36: What is the reason for this step? Is there ever going to be a cached environment?

.github/workflows/qc.yml
```diff
@@ -22,13 +22,37 @@ jobs:
...
+      - name: Windows specific step.
+        if: matrix.os == 'windows-latest'
+        run: echo "${HOME}/.local/bin" | Out-File -FilePath $env:GITHUB_PATH -Encoding utf8 -Append
+
+      #----------------------------------------------
+      # install dependencies if cache does not exist
+      #----------------------------------------------
       - name: Install dependencies
-        run: poetry install
+        if: steps.cached-poetry-dependencies.outputs.cache-hit != 'true'
```
matentzn, line 51: Id this ever going to be true?

pyproject.toml
hrshdhgd: These are the same dependencies that existed before, just the order has changed.

pyproject.toml
hrshdhgd: I am starting with the minimum possible dependency list to see where the error is occurring and also satisfies the tests to pass. This is still a work in progress and nothing is final yet. Be rest assured everything will be put back.

cmungall: @cthoyt and @matentzn are correct, if you depend on it, declare it. Just because it is a transitive dependency today does not mean it will be tomorrow.

What you should be doing is refactoring the code so these are not  mandatory direct requirements. There are standard patterns here. E.g move to extras and do a dynamic import.

We need to do this. This library has too many dependencies for a library that should be providing minimal core functionality for a simple standard

PR #401, opened 2023-07-28
Use `curies.Converter` more natively

cthoyt:
Blocked by #418

This PR is the next step towards using `curies.Converter`. 

- Replace the `prefix_map` inside the `Metadata` definition with a `converter`. Next steps are to do this for all models.
- Throw away many duplicate implementations of prefix map merging and replace with `curies.chain` as the one true implementation


src/sssom/context.py
```diff
@@ -25,108 +22,79 @@
...
+def get_internal_converter() -> Converter:
+    """Get a converter from the SSSOM internal context."""
+    context = _get_jsonld_context()
+    prefix_map = {}
+    for key in context["@context"]:
+        v = context["@context"][key]
+        if isinstance(v, str):
+            prefix_map[key] = v
+        elif isinstance(v, dict):
+            if "@id" in v and "@prefix" in v:
```
cthoyt, line 41: is this ID syntax actually part of the context map standard? We can upstream this into `curies` if so

src/sssom/context.py
```diff
@@ -25,108 +22,79 @@
...
 
 
-def get_jsonld_context():
-    """Get JSON-LD form of sssom_context variable from auto-generated 'internal_context.py' file.
-
-    [Auto generated from sssom.yaml by jsonldcontextgen.py]
-
-    :return: JSON-LD context
-    """
+def _get_jsonld_context():
```
matentzn, line 25: This wont be needed anymore after this PR is done.

src/sssom/context.py
```diff
@@ -3,18 +3,15 @@
 import json
 import logging
 import uuid
-from typing import Optional
+from typing import Any, Mapping, Optional, Union
```
matentzn, line 6: Mapping -> Can we alias it? It sort of conflicts with sssom.Mapping?

src/sssom/context.py
```diff
@@ -25,108 +22,79 @@
...
-def get_extended_prefix_map():
-    """Get prefix map from bioregistry (obo.epm.json).
-
-    :return: Prefix map.
-    """
-    converter = Converter.from_extended_prefix_map(EXTENDED_PREFIX_MAP)
-    return {record.prefix: record.uri_prefix for record in converter.records}
+def get_internal_converter() -> Converter:
+    """Get a converter from the SSSOM internal context."""
+    context = _get_jsonld_context()
```
matentzn, line 34: wont be needed anymore after this is done.

src/sssom/context.py
matentzn: None of this will be needed anymore when your PR is done.

src/sssom/context.py
```diff
@@ -25,108 +22,79 @@
...
 
 
-def get_extended_prefix_map():
-    """Get prefix map from bioregistry (obo.epm.json).
-
-    :return: Prefix map.
-    """
-    converter = Converter.from_extended_prefix_map(EXTENDED_PREFIX_MAP)
-    return {record.prefix: record.uri_prefix for record in converter.records}
+def get_internal_converter() -> Converter:
```
matentzn, line 32: No need for "internal" - "external" distinction. This control is established by controlling the OBO context.

src/sssom/context.py
```diff
@@ -25,108 +22,79 @@
...
+            if "@id" in v and "@prefix" in v:
+                if v["@prefix"]:
+                    prefix_map[key] = v["@id"]
+    del prefix_map["@vocab"]
+    return Converter.from_prefix_map(prefix_map)
 
 
-def get_built_in_prefix_map() -> PrefixMap:
-    """Get built-in prefix map from the sssom_context variable in the auto-generated 'internal_context.py' file.
+def get_external_converter() -> Converter:
```
matentzn, line 48: we only ned this from now on, so "external" / "internal" dichotomy is superfluous.

src/sssom/context.py
```diff
@@ -25,108 +22,79 @@
...
+    """Get prefix map from bioregistry (obo.epm.json)."""
+    return Converter.from_extended_prefix_map(EXTENDED_PREFIX_MAP)
 
-    [Auto generated from sssom.yaml by jsonldcontextgen.py]
 
-    :return: Prefix map
-    """
-    contxt = get_jsonld_context()
+def get_built_in_converter() -> Converter:
+    """Get built-in prefix map from the sssom_context variable in the auto-generated 'internal_context.py' file."""
```
matentzn, line 54: see above. not needed.

src/sssom/context.py
```diff
@@ -25,108 +22,79 @@
...
-        for k, v in builtinmap.items():
-            if k not in prefix_map:
-                prefix_map[k] = v
-            elif builtinmap[k] != prefix_map[k]:
-                raise ValueError(
-                    f"Built-in prefix {k} is specified ({prefix_map[k]}) but differs from default ({builtinmap[k]})"
-                )
-    return prefix_map
+        converter = Converter.from_prefix_map(prefix_map)
+    return curies.chain([converter, get_built_in_converter()])
```
matentzn, line 80: cool this is possible but not needed anymore: we just use the bioregistry converter and make sure through unit testing that it never violates our prefix preferences when updating the EPM.

src/sssom/io.py
```diff
@@ -138,23 +139,14 @@ def split_file(input_path: str, output_directory: Union[str, Path]) -> None:
...
-                if prefix not in prefix_map:
-                    prefix_map[prefix] = uri_prefix
-    return prefix_map
+def _get_converter(metadata: Metadata, prefix_map_mode: str = None) -> Converter:
+    if prefix_map_mode is None or prefix_map_mode == PREFIX_MAP_MODE_METADATA_ONLY:
+        return metadata.converter
+    if prefix_map_mode == PREFIX_MAP_MODE_SSSOM_DEFAULT_ONLY:
+        return get_default_metadata().converter
+    if prefix_map_mode == PREFIX_MAP_MODE_MERGED:
+        return curies.chain([metadata.converter, get_default_metadata().converter])
```
matentzn, line 148: Nice!

src/sssom/parsers.py
```diff
@@ -556,7 +561,7 @@ def from_sssom_rdf(
 
     ms.mappings = mlist  # type: ignore
     _set_metadata_in_mapping_set(mapping_set=ms, metadata=meta)
-    mdoc = MappingSetDocument(mapping_set=ms, prefix_map=prefix_map)
+    mdoc = MappingSetDocument(mapping_set=ms, prefix_map=converter.prefix_map)
```
matentzn, line 564: Now I get it. Great.

matentzn (commented):
Some quick comments, didnt do a thorough review yet.

There wont be any reason to distinguish internal and external converters moving forward, because we update the EPM during release and can ensure consistency with QC.

src/sssom/context.py
cthoyt: okay great then i'll just delete it. using this is a terrible pain

src/sssom/parsers.py
cthoyt: Next step will be replace the prefix_map object in the mapping set document with a converter directly

matentzn:
Lost track of this. @cthoyt I know you will be swamped, so let me know when this is ready for review.

cthoyt:
This has been split into many other PRs

PR #403, opened 2023-08-03
Add SSSOM rewire to rewire method

matentzn:
This PR introduces the possibility of rewiring an SSSOM table based on another SSSOM table. This is in scope for sssom-py because of the complex handing of SSSOM mapping set metadata.

### DIFF
--- a/src/sssom/cli.py
+++ b/src/sssom/cli.py
@@ -42,7 +42,7 @@ from .io import (
     split_file,
     validate_file,
 )
-from .parsers import parse_sssom_table
+from .parsers import from_sssom_dataframe, parse_sssom_table
 from .rdf_util import rewire_graph
 from .sparql_util import EndpointConfig, query_mappings
 from .util import (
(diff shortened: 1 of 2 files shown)


ehartley:
@matentzn yes, I can take this on.

matentzn:
Super!

cthoyt:
#426 is about fixing the reconciliation code, which also effectively includes the solution to this

matentzn:
@cthoyt Amazing! Ok then. @ehartley hold off then on any further thinking in this direction, lets wait for Charlie to tell us what, if anything, will be needed on top of his insane work right now :) Thanks @cthoyt

cthoyt:
@ehartley I added the code that standardizes a MSDF based on recent updates. You can continue on here. I realize I may have misunderstood what was going on here before, so I highly suggest writing several clear, explicit examples as unit tests before starting any implementation.

PR #415, opened 2023-09-09
Refactor SPARQL utilities to use `curies.Converter`

cthoyt:
This PR updates the `sparql_util.py` as well as the CLI endpoint that wraps it to use `curies.Converter` and also play more nicely with the `MappingSetDataFrame` constructor


hrshdhgd (dismissed):

hrshdhgd (commented):
Seems like there is some error in GH Actions

cthoyt:
@hrshdhgd is this an instance of how poetry is programmed to fail randomly 1/20 times? This really happens on purpose. The devs do some sketchy stuff

hrshdhgd:
It seems like poetry and GH Actions environment have some incompatibilities from time to time. It is truly frustrating. Locally poetry works without any of these issues.

hrshdhgd (approved):

PR #416, opened 2023-09-10
Use `curies.Converter` in OntoPortal exporter

cthoyt:
This PR begins to switch the OntoPortal exporter to use `curies.Converter` for expansion and compression. Notably, this replaces ad hoc code with battle tested and provably correct expansion code.

However, it's not so obvious what the data is supposed to look like in the end since the test cases only check that there is some data, and not actually what it looks like.

@syphax-bouazzouni can you help build up the test cases so we can be more sure that this is working correctly? Is there documentation somewhere for the schema of JSON that this exporter corresponds to?


syphax-bouazzouni:
Hi, thanks for the refactoring.
An example of a valid ontoportal schema is as following  
```json
[
  {
    "classes": [
      "http://bioontology.org/ontologies/BiomedicalResourceOntology.owl#Image_Algorithm",
      "http://purl.org/incf/ontology/Computational_Neurosciences/cno_alpha.owl#cno_0000202"
    ],
    "name": "This is the mappings produced to test the bulk load",
    "source": "https://w3id.org/semapv/LexicalMatching",
    "comment": "mock data",
    "relation": [
      "http://www.w3.org/2002/07/owl#subClassOf"
    ],
    "subject_source_id": "http://bioontology.org/ontologies/BiomedicalResources.owl",
    "object_source_id": "http://purl.org/incf/ontology/Computational_Neurosciences/cno_alpha.owl",
    "source_name": "https://w3id.org/sssom/mapping/tests/data/basic.tsv",
    "source_contact_info": "orcid:1234,orcid:5678",
    "date": "2020-05-30"
  }
]
```

Required fields are classes are names.

cthoyt:
@syphax-bouazzouni @hrshdhgd this PR has a great example of a clear, explicit unit test. something to take inspiration of in the future :)

matentzn (approved):
Assuming @syphax-bouazzouni agrees with this, looks good!

PR #420, opened 2023-09-12
Better handle URI fields when getting prefixes from metadata

### LINKED ISSUE #419: get_prefixes_used_in_metadata breaks when using URIs in the *_source columns
From a discussion with @cthoyt on the sssom Slack channel: 
> https://github.com/mapping-commons/sssom-py/blob/fda75658f687171cc2de439fae9faa64620e7a97/src/sssom/util.py#L1119C1-L1130C20 breaks when you use URIs in the subject_source and object_source columns.

version: sssom, version 0.3.41

## Steps needed to reproduce the issue:
Download the files:
- https://raw.githubusercontent.com/jmillanacosta/mappings_enm/58d31d60d9d2cc4df6a0ea0990ab4805d1b3f351/sssom/ENMO_mappings.tsv
- https://raw.githubusercontent.com/jmillanacosta/mappings_enm/58d31d60d9d2cc4df6a0ea0990ab4805d1b3f351/sssom/ENMO_mapping_set.YAML

And run: 
```sssom parse ENMO_mappings.tsv -m ENMO_mapping_set.YAML -o enmo.sssom.tsv```

```
WARNING:root:https is used in the SSSOM mapping set but it does not exist in the prefix map
WARNING:root:http is used in the SSSOM mapping set but it does not exist in the prefix map
  File "/home/javier/miniconda3/bin/sssom", line 8, in <module>
    sys.exit(main())
  File "/home/javier/miniconda3/lib/python3.9/site-packages/click/core.py", line 1157, in __call__
    return self.main(*args, **kwargs)
  File "/home/javier/miniconda3/lib/python3.9/site-packages/click/core.py", line 1078, in main
    rv = self.invoke(ctx)
  File "/home/javier/miniconda3/lib/python3.9/site-packages/click/core.py", line 1688, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/javier/miniconda3/lib/python3.9/site-packages/click/core.py", line 1434, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/javier/miniconda3/lib/python3.9/site-packages/click/core.py", line 783, in invoke
    return __callback(*args, **kwargs)
  File "/home/javier/miniconda3/lib/python3.9/site-packages/sssom/cli.py", line 215, in parse
    parse_file(
  File "/home/javier/miniconda3/lib/python3.9/site-packages/sssom/io.py", line 112, in parse_file
    doc.clean_prefix_map(strict=strict_clean_prefixes)
  File "/home/javier/miniconda3/lib/python3.9/site-packages/sssom/util.py", line 185, in clean_prefix_map
    raise ValueError(
ValueError: ['https', 'http'] are used in the SSSOM mapping set but it does not exist in the prefix map
```

### PR DESCRIPTION
**Pull request by cthoyt**
Closes #419, cc @jmillanacosta

This PR adds a list of fields that should be given as URIs to be skipped when looking through metadata associated with a given mapping set dataframe so that `http` and `https` aren't interpreted as CURIE prefixes.

@hrshdhgd is there a way to introspect over the SSSOM schema and identify fields that are only supposed to be URIs (and therefore not CURIEs)?

### CODE REVIEW

**Comment by hrshdhgd** [comment]
From [`url.py -> inject_metadata_into_df()`](https://github.com/mapping-commons/sssom-py/blob/master/src/sssom/util.py#L837)

https://github.com/mapping-commons/sssom-py/blob/035bbd447322e867b5f0800ae23fc43031c32b4c/src/sssom/util.py#L845-L846


You can get types using `schema['slots'][**name-of-slot**]['range']` gives you the type of the slot. 

For e.g.: `schema['slots']['subject_id']['range'] = 'EntityReference'` and the definition of `EntityReference` is [here](https://github.com/mapping-commons/sssom/blob/74aeffb4e6dc07751df5d143566e3d120cc9b347/src/sssom_schema/schema/sssom_schema.yaml#L64-L69) which is a `uriorcurie`. 

Based on this information, there may be a different solution we may want to think about.

File: src/sssom/context.py [hunk 1]
```diff
@@ -16,7 +16,7 @@
 # EXTERNAL_CONTEXT_PATH = HERE / "obo.epm.json"
 
 SSSOM_URI_PREFIX = "https://w3id.org/sssom/"
-SSSOM_BUILT_IN_PREFIXES = ("sssom", "owl", "rdf", "rdfs", "skos", "semapv")
+SSSOM_BUILT_IN_PREFIXES = ("sssom", "owl", "rdf", "rdfs", "skos", "semapv", "orcid")
```
gouttegd, line 19: Is `orcid` now supposed to be a built-in prefix in SSSOM?

If so, can you please make sure that [the specification](https://mapping-commons.github.io/sssom/spec/#tsv) is updated accordingly? Currently, the spec still stays that only `sssom`, `owl`, `rdf`, `rdfs`, `skos`, `semapv` are built-in.

cthoyt: Hmm, it hadn't occurred to me that this was part of the spec and not an implementation detail. I don't have the motivation to get into a discussion about this so I will change the PR

gouttegd: The list of built-in prefixes is part of the spec because it wouldn’t be great if a given prefix was considered built-in by one implementation and not built-in by another. A mapping set that makes use of that prefix without explicitly declaring it (thinking it is universally built-in) would be accepted by the first implementation and rejected by the second.

hrshdhgd (approved):

PR #421, opened 2023-09-12
Use `curies.Converter` in Metadata object

cthoyt:
This is the next step in the process to switching everything to use converters.


src/sssom/typehints.py
```diff
@@ -19,5 +24,33 @@
...
+
+        return cls(
+            converter=get_converter(),
+            metadata=get_default_metadata(),
+        )
+
+
+def generate_mapping_set_id() -> str:
+    """Getnerate a mapping set ID."""
+    return f"{SSSOM_URI_PREFIX}mappings/{uuid.uuid4()}"
```
matentzn, line 48: Note that this logic seems duplicated with https://github.com/mapping-commons/sssom-py/pull/421/files#diff-fa45af370531fd5131bc80371b323932f856b104add1550cb1e7fe281f821133R14

src/sssom/typehints.py
```diff
@@ -19,5 +24,33 @@
...
+        from .context import get_converter
+
+        return cls(
+            converter=get_converter(),
+            metadata=get_default_metadata(),
+        )
+
+
+def generate_mapping_set_id() -> str:
+    """Getnerate a mapping set ID."""
```
matentzn, line 47:
```suggestion
    """Generate a mapping set ID."""
```

matentzn (commented):
Two comments, tehn I will approve

src/sssom/typehints.py
cthoyt: Resolved in 98122ac

matentzn (approved):

PR #422, opened 2023-09-14
Improve MSDF instantiation and context functions

cthoyt:
We're getting dangerously close to pulling the converter conversion trigger :)


cthoyt:
@hrshdhgd seems like more spurious poetry errors in the build, otherwise this should be passing all tests

hrshdhgd (approved):
Still not sure if it is GH Actions or Poetry. But this happens with just this project.

PR #426, opened 2023-09-15
Re-implement reconciliation code with `curies`

cthoyt:
This PR replaces the `reconcile_prefix_and_data` with more carefully defined operations from the `curies` package (implemented in https://github.com/cthoyt/curies/pull/74 and https://github.com/cthoyt/curies/pull/78).

The new implementation is built on `curies.remap_curie_prefixes` and `curies.rewire`. Note that if you want to overwrite a CURIE prefix, e.g., in the Bioregistry extended prefix map, you need to provide a place for the old one to go as in `{"geo": "ncbi.geo", "geogeo": "geo"}`. Just doing `{"geogeo": "geo"}` would not work since `geo` already exists.

This is a slight change from the existing implementation, but I think it is better since it allows for solving a much more general problem of dependent (i.e., _transitive_) remappings


cthoyt:
This PR is finished but is currently failing CI because of non-deterministic issues with poetry (is it time to get rid of poetry yet 😛?)

cthoyt:
@hrshdhgd please help, we have a poetry issue again

.github/workflows/qc.yml
```diff
@@ -27,8 +27,6 @@ jobs:
       #----------------------------------------------
       - name: Install Poetry
         uses: snok/install-poetry@v1
-        with:
-          version: 1.3.2
```
matentzn, line 31: @hrshdhgd double check if right

src/sssom/util.py
```diff
@@ -1105,62 +1105,25 @@ def reconcile_prefix_and_data(
     """
     # Discussion about this found here:
     # https://github.com/mapping-commons/sssom-py/issues/216#issue-1171701052
+    converter = msdf.converter
+    converter = curies.remap_curie_prefixes(converter, prefix_reconciliation["prefix_synonyms"])
+    converter = curies.rewire(converter, prefix_reconciliation["prefix_expansion_reconciliation"])
```
matentzn, line 1110: Use constants instead of strings to satisfy SRP?

tests/test_collapse.py
```diff
@@ -76,13 +74,51 @@ def test_reconcile_prefix(self):
...
+            msdf.prefix_map,
+        )
+        prefix_reconciliation = {
+            "prefix_synonyms": {
+                "a": "c",
+                "b": "bravo",
+                "r": "rdfs",
+                "o": "owl",
+            },
+            "prefix_expansion_reconciliation": {
```
matentzn, line 100: Rename `expansion_reconciliation` to `rewiring` to maintain consistent terminology?

tests/test_collapse.py
```diff
@@ -76,13 +74,51 @@ def test_reconcile_prefix(self):
...
+                "c": "http://test.owl/c/",
+                "bravo": "http://test.owl/bravo",
+                "rdfs": "http://www.w3.org/2000/01/rdf-schema#",
+                "owl": "http://www.w3.org/2002/07/owl#",
+            },
+        }
 
         recon_msdf = reconcile_prefix_and_data(msdf, prefix_reconciliation)
```
matentzn, line 108: I know this is not part of this PR, but consider turning the `prefix_reconciliation` object into a proper class instance for better documentation. Having the client need to construct a specifically formed yaml may not be the best way to ensure their happiness :P

tests/test_collapse.py
```diff
@@ -76,13 +74,51 @@ def test_reconcile_prefix(self):
...
-
-        prefix_expansion = prefix_reconciliation["prefix_expansion_reconciliation"]
-
-        for pfx, exp in prefix_expansion.items():
-            if pfx in recon_msdf.prefix_map.keys():
-                self.assertEqual(recon_msdf.prefix_map[pfx], exp)
+        self.assertEqual(
+            {
+                "a": "http://example.org/a/",
```
matentzn, line 111: Why is "a" still present? Shouldn't it have been remapped to c before it go rewired?

tests/test_collapse.py
```diff
@@ -76,13 +74,51 @@ def test_reconcile_prefix(self):
...
+                "sssom": "https://w3id.org/sssom/",
+            },
+            msdf.prefix_map,
+        )
+        prefix_reconciliation = {
+            "prefix_synonyms": {
+                "a": "c",
+                "b": "bravo",
+                "r": "rdfs",
+                "o": "owl",
```
matentzn, line 98: Isnt the `"o": "owl",` redundant with the `"r": "rdfs",` check? In both cases, non existing prefix is remapped - without an effect.

.github/workflows/qc.yml
cthoyt: This deletion didn’t work

tests/test_collapse.py
cthoyt: Agreed

tests/test_collapse.py
cthoyt: The mapping "a":"c" got ignored since `c` is already present in the prefix map and this would cause an inconsistency

.github/workflows/qc.yml
hrshdhgd: It won't work. We have determined that `poetry==1.3.2` is the latest version that plays well with GH Actions. I'm not aware if the `poetry` developers have fixed their issues with GH actions but for now this works.

It seems like all the checks pass now.

tests/test_collapse.py
cthoyt: So the question is: do we want to add an "overwrite" setting? I don't think that having this be default functionality makes sense. if you don't want original prefix `c` in the map, then maybe there should be a 3rd part to this workflow for explicit deletions

matentzn: Hmmm interesting. I think we would like a --force option, which results in the removal of the already existing entry, and a logging.warning(). What do you think?

cthoyt: Yes, I think having a flag would be the way to go. I will mock that in the `curies` package.

src/sssom/util.py
cthoyt: I don't know what SRP means, but let's save updating this data structure for a subsequent PR

tests/test_collapse.py
cthoyt: I have a solution! Pushing now

cthoyt: Seems to work nicely! My new solution doesn't allow deletions, but requires that if you remap a->c and there is already a c, you need to give c-> something else. Seems reasonable to me

.github/workflows/qc.yml
```diff
@@ -22,6 +22,9 @@ jobs:
         with:
           python-version: ${{ matrix.python-version }}
 
+      - name: Install poetry dependencies
+        run: python -m pip install -U pip setuptools
```
matentzn, line 26: I assume @hrshdhgd has approved this; I don't really know exactly what may and may not go into the qc.yml file.

src/sssom/util.py
matentzn: [Single responsibility principle](https://en.wikipedia.org/wiki/Single-responsibility_principle): classes/methods/functions should only have one reason to change. If this string changes for whatever reason, many different classes and places in the code will have to change as well.

tests/test_collapse.py
matentzn: What about this?

matentzn (approved):
No reasons not to accept, but two questions still open.

matentzn:
Ok, I merged it now under the flag of progress 🙏

matentzn:
@hrshdhgd can you comment on wether you agree to the change in the action?

.github/workflows/qc.yml
hrshdhgd: I had not seen this before but this won't do any harm as far as I can tell.

cthoyt: On the next PR, this continued failing, so this change didn't do anything :/

PR #429, opened 2023-09-26
Convert SSSOM TSV => JSON => TSV and confirm the `MappingSetDataFrame` object remains consistent.

hrshdhgd:
Fixes https://github.com/mapping-commons/sssom/issues/321

Given a simple sssom tsv
```
#curie_map:
#  FBbt: "http://purl.obolibrary.org/obo/FBbt_"
#  ORCID: "https://orcid.org/"
#  UBERON: "http://purl.obolibrary.org/obo/UBERON_"
#creator_id:
#  - "ORCID:0000-0002-6095-8718"
subject_id	subject_label	predicate_id	object_id	mapping_justification
FBbt:00000001	organism	semapv:crossSpeciesExactMatch	UBERON:0000468	semapv:ManualMappingCuration
```

- I first parse it using `parse_sssom_table()` 
-  `msdf1.clean_prefix_map()`
- Convert to JSON using `to_json()` and write to file
- Parse again using `parse_sssom_json()`
- `msdf2.clean_prefix_map()`

The prefix maps of both should be the same.

### DIFF
--- a/src/sssom/parsers.py
+++ b/src/sssom/parsers.py
@@ -212,21 +212,28 @@ def parse_sssom_table(
                     logging.info(f"Externally provided metadata {k}:{v} is added to metadata set.")
                     sssom_metadata[k] = v
         meta = sssom_metadata
-
-        if "curie_map" in sssom_metadata:
+        if CURIE_MAP in sssom_metadata:
             if prefix_map:
+                # Convert sssom_metadata[CURIE_MAP] keys to lowercase for case-insensitive comparison
+                curie_map_lower = {k.lower(): k for k in sssom_metadata[CURIE_MAP].keys()}
+
                 for k, v in prefix_map.items():
-                    if k in sssom_metadata[CURIE_MAP]:
-                        if sssom_metadata[CURIE_MAP][k] != v:
+                    k_lower = k.lower()
+
+                    if k_lower in curie_map_lower:
+                        original_key = curie_map_lower[k_lower]
+                        if sssom_metadata[CURIE_MAP][original_key] != v:
                             logging.warning(
-                                f"SSSOM prefix map {k} ({sssom_metadata[CURIE_MAP][k]}) "
-                                f"conflicts with provided ({prefix_map[k]})."
+                                f"SSSOM prefix map {original_key} ({sssom_metadata[CURIE_MAP][original_key]}) "
+                                f"conflicts with provided ({v})."
                             )
+                            sssom_metadata[CURIE_MAP][original_key] = v
                     else:
                         logging.info(
                             f"Externally provided metadata {k}:{v} is added to metadata set."
                         )
                         sssom_metadata[CURIE_MAP][k] = v
+
             prefix_map = sssom_metadata[CURIE_MAP]
 
     meta_all = _get_prefix_map_and_metadata(prefix_map=prefix_map, meta=meta)
@@ -264,10 +271,32 @@ def parse_sssom_json(
 ) -> MappingSetDataFrame:
     """Parse a TSV to a :class:`MappingSetDocument` to a  :class`MappingSetDataFrame`."""
     raise_for_bad_path(file_path)
-    metadata = _get_prefix_map_and_metadata(prefix_map=prefix_map, meta=meta)
 
     with open(file_path) as json_file:
         jsondoc = json.load(json_file)
+
+    # Get prefix map from jsondoc and update metadata.
+    # This takes priority over default prefix_map in case of a tie.
+    jsondoc_prefix_map = jsondoc["@context"]
+
+    # Convert keys in both maps to lower case for comparison
+    if prefix_map:
+        lowercase_prefix_map = {k.lower(): v for k, v in prefix_map.items()}
+
+        # Iterate over jsondoc_prefix_map
+        for key, value in jsondoc_prefix_map.items():
+            # If lowercase key exists in lowercase_prefix_map, update the value and key
+            if key.lower() in lowercase_prefix_map:
+                # Remove the old key-value pair
+                if key in prefix_map:
+                    del prefix_map[key]
+                elif key.lower() in prefix_map:
+                    del prefix_map[key.lower()]
+                # Add the new key-value pair
+                prefix_map[key] = value
+
+    metadata = _get_prefix_map_and_metadata(prefix_map=prefix_map, meta=meta)
+
     msdf = from_sssom_json(jsondoc=jsondoc, prefix_map=metadata.prefix_map, meta=metadata.metadata)
     # df: pd.DataFrame = msdf.df
     # if mapping_predicates and not df.empty():
--- a/src/sssom/parsers.py
+++ b/src/sssom/parsers.py
@@ -6,12 +6,13 @@ import json
 import logging as _logging
 import re
 import typing
-from collections import Counter
+from collections import ChainMap, Counter
 from pathlib import Path
 from typing import Any, Callable, Dict, Iterable, List, Optional, TextIO, Tuple, Union, cast
 from xml.dom import Node, minidom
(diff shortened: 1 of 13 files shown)


hrshdhgd:
At the moment this fails. Here's the reason why:

Here's the test code:

https://github.com/mapping-commons/sssom-py/blob/831a58eb637eacba38355c32558ab5eaf7f34961/tests/test_parsers.py#L250-L265


`msdf1.prefix_map` looks like this: 
```yaml
{
'owl': 'http://www.w3.org/2002/07/owl#', 
'rdf': 'http://www.w3.org/1999/02/22-rdf-syntax-ns#', 
'rdfs': 'http://www.w3.org/2000/01/rdf-schema#',
'semapv': 'https://w3id.org/semapv/vocab/', 
'skos': 'http://www.w3.org/2004/02/skos/core#',
'sssom': 'https://w3id.org/sssom/', 
'FBbt': 'http://purl.obolibrary.org/obo/FBbt_', 
'ORCID': 'https://orcid.org/', 
'UBERON': 'http://purl.obolibrary.org/obo/UBERON_'
}
```

whereas `msdf2.prefix_map` looks like this
```yaml
{
'owl': 'http://www.w3.org/2002/07/owl#', 
'rdf': 'http://www.w3.org/1999/02/22-rdf-syntax-ns#', 
'rdfs': 'http://www.w3.org/2000/01/rdf-schema#', 
'semapv': 'https://w3id.org/semapv/vocab/', 
'skos': 'http://www.w3.org/2004/02/skos/core#', 
'sssom': 'https://w3id.org/sssom/', 
'FBbt': 'http://purl.obolibrary.org/obo/FBbt_', 
'UBERON': 'http://purl.obolibrary.org/obo/UBERON_'
}

```
`'ORCID': 'https://orcid.org/', ` is missing from `msdf2.prefix_map`

In the `clean_prefix_map()` function, the `prefixes_in_table` variable for `msdf1` = `{'owl', 'rdfs', 'skos', 'sssom', 'UBERON', 'ORCID', 'semapv', 'rdf', 'FBbt'}` which is the same for `msdf2`. In spite of that, the `self.prefix_map` generated for both are different.
```python
self.prefix_map = dict(subconverter.bimap)
```


This line is where the `ORCID` prefix gets lost:

```python
subconverter = self.converter.get_subconverter(prefixes_in_table)
```

@cthoyt , would you have any idea why this would be happening?

cc: @matentzn

cthoyt:
The function that gets all of the prefixes appearing in the msdf is probably the culprit

src/sssom/parsers.py
```diff
@@ -212,21 +212,28 @@ def parse_sssom_table(
                     logging.info(f"Externally provided metadata {k}:{v} is added to metadata set.")
                     sssom_metadata[k] = v
         meta = sssom_metadata
-
-        if "curie_map" in sssom_metadata:
+        if CURIE_MAP in sssom_metadata:
```
cthoyt, line 215: I have rewritten all of this code using `curies` already and am waiting for #426 (ready) then #401 (needs some more work) to get merged. I think it's better that you reduce effort on re-writing it again, since the `curies` solution is much more clean and elegant.

src/sssom/parsers.py
```diff
@@ -264,10 +271,32 @@ def parse_sssom_json(
...
-    metadata = _get_prefix_map_and_metadata(prefix_map=prefix_map, meta=meta)
 
     with open(file_path) as json_file:
         jsondoc = json.load(json_file)
+
+    # Get prefix map from jsondoc and update metadata.
+    # This takes priority over default prefix_map in case of a tie.
+    jsondoc_prefix_map = jsondoc["@context"]
+
+    # Convert keys in both maps to lower case for comparison
```
cthoyt, line 282: Is this wise? I think we should not be doing all of this lowercasing. The prefix maps should be correct as they are. Doing all of these ad-hoc operations makes sssom-py insanely hard to work with (I can tell you this from first hand experience since I have tried to understand the whole history of the package)

cthoyt (changes requested):
Please wait on making progress with this until the PRs that replace a lot of the same code with `curies` are merged - hopefully this encourages much more consistent and elegant handling of prefix maps and reduces the number of changes to address the issue here

tests/test_parsers.py
```diff
@@ -245,3 +246,20 @@ def test_parse_obographs_merged(self):
             )
         msdf = parse_sssom_table(outfile)
         self.assertTrue(custom_curie_map.items() <= msdf.prefix_map.items())
+
+    def test_tsv_to_json_and_back(self):
+        """Test converting SSSOM TSV => JSON => SSSOM TSV such that it is reproducible."""
+        sample_tsv = f"{test_data_dir}/sample1.sssom.tsv"
+        json_outfile = f"{test_out_dir}/sample1.json"
+        msdf1 = parse_sssom_table(sample_tsv)
```
cthoyt, line 254: @hrshdhgd other hints: when writing tests, make the things you're testing more explicit. The way this test is written, it doesn't really test anythign specific. Why not explicitly check that `ORCID` is in the places you expect to to be? Otherwise, someone reading this test learns nothing about what's actually supposed to be happening here.

tests/test_parsers.py
```diff
@@ -245,3 +246,24 @@ def test_parse_obographs_merged(self):
...
+        self.assertIn("ORCID", msdf1.prefix_map)
+        json_doc = to_json(msdf1)
+
+        self.assertEqual(msdf1.prefix_map, json_doc["@context"])
+
+        with open(json_outfile, "w") as file:
+            write_json(msdf1, file)
+
+        msdf2 = parse_sssom_json(json_outfile)
+        self.assertIn("ORCID", msdf2.prefix_map)
```
cthoyt, line 266: I added some more explicit tests so you can see where things are actually going wrong, which appers to be in the SSSOM JSON parsing, and has nothing to do with cleaning the prefix map like I suggested in a previous comment

hrshdhgd:
@cthoyt , thanks for the insight! This branch was just to investigate the issue Damien pointed out (hence the draft). As for the lowercasing, I don't like that solution at all. I was just trying to understand the issue better so I could explain it to someone else via code (you and Nico).

src/sssom/parsers.py
cthoyt: I split the code for updating this into #431

cthoyt:
@hrshdhgd I think this is now solved with https://github.com/mapping-commons/sssom-py/pull/431. Maybe you can rebase then see if you can add some more tests to make sure

cthoyt (changes requested):
I am now 100% against adding any ad hoc code that works on prefixes or prefix maps into sssom-py.

If people add incorrect prefixes that aren't defined, why don't we give them appropriate warnings/errors?

Besides, this code shouldn't be necessary to address the original issue, which was something to do with incorrectly writing SSSOM files that was fixed in another PR

matentzn (changes requested):
Can  you add a failing test here that roundtrips through TSV->JSON->TSV before we do anything else? I think the problem is somewhere deeper, but perhaps the solution is somewhere along the lines of https://github.com/mapping-commons/sssom-py/pull/431/files#diff-aed80c160ecfb8db6a235671a6d4ed0ae74470301705533ff4ec2a7a36dd989fR444

hrshdhgd:
Charlie's test validates the issue:

https://github.com/mapping-commons/sssom-py/blob/e06a4c690127bb58db1fb29fea71793f2139879b/tests/test_parsers.py#L256-L325

matentzn:
This test does not seem to have a JSON serialise and parse step in it though - what we need is a test that takes Damiens test case, serialises it to JSON, then parses the JSON, then ensures that pre JSON and post JSON msdf objects are the same..

PR #430, opened 2023-09-26
Switch MSDF to use `curies.Converter` natively

cthoyt:


cthoyt:
@hrshdhgd this is ready to go besides the mangling I did for getting tests to work... which they still don't

hrshdhgd (approved):

PR #431, opened 2023-09-26
Simplify parsing SSSOM w/ `curies.chain`

### LINKED ISSUE #363: Replace CURIE inference mechanism with curies.Converter.from_extended_prefix_map

cthoyt:
Closes #363 (final nail in the coffin)

This provides an alternative to https://github.com/mapping-commons/sssom-py/pull/429 that makes more explicit the chaining operations done on the metadata and prefix maps. This is also a good change to carefully document the way that this is handled, since I might not have captured it accurately. As it is, The priority order for combining prefix maps are:

1. Internal prefix map inside the document
2. Prefix map passed through this function inside the ``meta``
3. Prefix map passed through this function to ``prefix_map``
4. Default prefix map (handled with ensure_converter)

### DIFF
--- a/src/sssom/context.py
+++ b/src/sssom/context.py
@@ -58,7 +58,7 @@ def ensure_converter(prefix_map: HINT = None, *, use_bioregistry: bool = True) -
         if use_bioregistry:
             return get_converter()
         else:
-            return Converter([])
+            return _get_built_in_prefix_map()
     if not isinstance(prefix_map, Converter):
         prefix_map = Converter.from_prefix_map(prefix_map)
     return curies.chain([_get_built_in_prefix_map(), prefix_map])
--- a/src/sssom/context.py
+++ b/src/sssom/context.py
@@ -52,10 +52,29 @@ def _get_built_in_prefix_map() -> Converter:
 HINT = Union[None, PrefixMap, Converter]
 
 
-def ensure_converter(prefix_map: HINT = None, *, use_bioregistry: bool = True) -> Converter:
-    """Ensure a converter is available."""
+def ensure_converter(prefix_map: HINT = None, *, use_defaults: bool = True) -> Converter:
+    """Ensure a converter is available.
+
+    :param prefix_map: One of the following:
+
+        1. An empty dictionary or ``None``. This results in using the default
+           extended prefix map (currently based on a variant of the Bioregistry)
+           if ``use_defaults`` is set to true, otherwise just the builtin prefix
+           map including the prefixes in :data:`SSSOM_BUILT_IN_PREFIXES`
+        2. A non-empty dictionary representing a prefix map. This is loaded as a
+           converter with :meth:`Converter.from_prefix_map`. It is chained
+           behind the builtin prefix map to ensure none of the
+           :data:`SSSOM_BUILT_IN_PREFIXES` are overwritten with non-default values
+        3. A pre-instantiated :class:`curies.Converter`. Similarly to a prefix
+           map passed into this function, this is chained behind the builtin prefix
+           map
+    :param use_defaults: If an empty dictionary or None is passed to this function,
+        this parameter chooses if the extended prefix map (currently based on a
+        variant of the Bioregistry) gets loaded.
+    :returns: A re-usable converter
+    """
     if not prefix_map:
-        if use_bioregistry:
+        if use_defaults:
             return get_converter()
         else:
             return _get_built_in_prefix_map()
--- a/src/sssom/context.py
+++ b/src/sssom/context.py
@@ -52,10 +52,13 @@ def _get_built_in_prefix_map() -> Converter:
 HINT = Union[None, PrefixMap, Converter]
 
 
-def ensure_converter(prefix_map: HINT = None) -> Converter:
+def ensure_converter(prefix_map: HINT = None, *, use_bioregistry: bool = True) -> Converter:
     """Ensure a converter is available."""
     if not prefix_map:
-        return get_converter()
+        if use_bioregistry:
+            return get_converter()
+        else:
+            return Converter([])
     if not isinstance(prefix_map, Converter):
         prefix_map = Converter.from_prefix_map(prefix_map)
     return curies.chain([_get_built_in_prefix_map(), prefix_map])
--- a/src/sssom/parsers.py
+++ b/src/sssom/parsers.py
@@ -202,13 +202,12 @@ def parse_sssom_table(
     #  2. Internal prefix map inside the document
     #  3. Prefix map passed through this function inside the ``meta``
     #  4. Prefix map passed through this function to ``prefix_map`` (handled with ensure_converter)
-    #  5. Default prefix map (handled with ensure_converter)
     converter = curies.chain(
         [
             _get_built_in_prefix_map(),
             Converter.from_prefix_map(sssom_metadata.pop(CURIE_MAP, {})),
             Converter.from_prefix_map(meta.pop(CURIE_MAP, {})),
-            ensure_converter(prefix_map),
+            ensure_converter(prefix_map, use_bioregistry=False),
         ]
     )
 
--- a/src/sssom/parsers.py
+++ b/src/sssom/parsers.py
@@ -6,12 +6,13 @@ import json
 import logging as _logging
 import re
 import typing
-from collections import Counter
+from collections import ChainMap, Counter
 from pathlib import Path
 from typing import Any, Callable, Dict, Iterable, List, Optional, TextIO, Tuple, Union, cast
 from xml.dom import Node, minidom
 from xml.dom.minidom import Document
 
+import curies
 import numpy as np
 import pandas as pd
 import requests
@@ -55,7 +56,13 @@ from sssom.constants import (
 
 from .context import HINT, ensure_converter
 from .sssom_document import MappingSetDocument
-from .typehints import Metadata, MetadataType, PrefixMap, generate_mapping_set_id
+from .typehints import (
+    Metadata,
+    MetadataType,
+    PrefixMap,
+    generate_mapping_set_id,
+    get_default_metadata,
+)
 from .util import (
     PREFIX_MAP_KEY,
     SSSOM_DEFAULT_RDF_SERIALISATION,
@@ -193,44 +200,36 @@ def parse_sssom_table(
     stream: io.StringIO = _open_input(file_path)
     sep_new = _get_seperator_symbol_from_file_path(file_path)
     df, sssom_metadata = _read_pandas_and_metadata(stream, sep_new)
-    # if mapping_predicates:
-    #     # Filter rows based on presence of predicate_id list provided.
-    #     df = df[df["predicate_id"].isin(mapping_predicates)]
+    if meta is None:
+        meta = {}
 
-    # If SSSOM external metadata is provided, merge it with the internal metadata
+    # The priority order for combining prefix maps are:
+    #  1. Internal prefix map inside the document
+    #  2. Prefix map passed through this function inside the ``meta``
+    #  3. Prefix map passed through this function to ``prefix_map``
+    #  4. Default prefix map (handled with ensure_converter)
+    converter = curies.chain(
+        [
+            Converter.from_prefix_map(sssom_metadata.pop(CURIE_MAP, {})),
+            Converter.from_prefix_map(meta.pop(CURIE_MAP, {})),
+            ensure_converter(prefix_map),
+        ]
+    )
 
-    if sssom_metadata:
-        if meta:
-            for k, v in meta.items():
-                if k in sssom_metadata:
-                    if sssom_metadata[k] != v:
-                        logging.warning(
-                            f"SSSOM internal metadata {k} ({sssom_metadata[k]}) "
-                            f"conflicts with provided ({meta[k]})."
-                        )
-                else:
-                    logging.info(f"Externally provided metadata {k}:{v} is added to metadata set.")
-                    sssom_metadata[k] = v
-        meta = sssom_metadata
+    # The priority order for combining metadata is:
+    #  1. Metadata appearing in the SSSOM document
+    #  2. Metadata passed through ``meta`` to this function
+    #  3. Default metadata
+    combine_meta = dict(
+        ChainMap(
+            sssom_metadata,
+            meta,
+            get_default_metadata(),
+        )
+    )
 
-        if "curie_map" in sssom_metadata:
-            if prefix_map:
-                for k, v in prefix_map.items():
-                    if k in sssom_metadata[CURIE_MAP]:
-                        if sssom_metadata[CURIE_MAP][k] != v:
-                            logging.warning(
-                                f"SSSOM prefix map {k} ({sssom_metadata[CURIE_MAP][k]}) "
-                                f"conflicts with provided ({prefix_map[k]})."
-                            )
-                    else:
-                        logging.info(
-                            f"Externally provided metadata {k}:{v} is added to metadata set."
-                        )
-                        sssom_metadata[CURIE_MAP][k] = v
-            prefix_map = sssom_metadata[CURIE_MAP]
-
-    meta_all = _get_prefix_map_and_metadata(prefix_map=prefix_map, meta=meta)
-    msdf = from_sssom_dataframe(df, prefix_map=meta_all.prefix_map, meta=meta_all.metadata)
+    msdf = from_sssom_dataframe(df, prefix_map=converter, meta=combine_meta)
+    msdf.clean_prefix_map()
     return msdf
 
 
@@ -394,7 +393,7 @@ def parse_alignment_xml(
 
 def from_sssom_dataframe(
     df: pd.DataFrame,
-    prefix_map: Optional[PrefixMap] = None,
+    prefix_map: HINT = None,
     meta: Optional[MetadataType] = None,
 ) -> MappingSetDataFrame:
     """Convert a dataframe to a MappingSetDataFrame.
--- a/src/sssom/parsers.py
+++ b/src/sssom/parsers.py
@@ -207,7 +207,7 @@ def parse_sssom_table(
             _get_built_in_prefix_map(),
             Converter.from_prefix_map(sssom_metadata.pop(CURIE_MAP, {})),
             Converter.from_prefix_map(meta.pop(CURIE_MAP, {})),
-            ensure_converter(prefix_map, use_bioregistry=False),
+            ensure_converter(prefix_map, use_defaults=False),
         ]
     )
 
--- a/src/sssom/parsers.py
+++ b/src/sssom/parsers.py
@@ -198,7 +198,7 @@ def parse_sssom_table(
         meta = {}
 
     # The priority order for combining prefix maps are:
-    #  1. Internal prefix map inside the document inside
+    #  1. Internal prefix map inside the document
     #  2. Prefix map passed through this function inside the ``meta``
     #  3. Prefix map passed through this function to ``prefix_map`` (handled with ensure_converter)
     #  4. Default prefix map (handled with ensure_converter)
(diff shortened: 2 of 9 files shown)


src/sssom/parsers.py
```diff
@@ -193,44 +200,36 @@ def parse_sssom_table(
...
-                        logging.info(
-                            f"Externally provided metadata {k}:{v} is added to metadata set."
-                        )
-                        sssom_metadata[CURIE_MAP][k] = v
-            prefix_map = sssom_metadata[CURIE_MAP]
+    if meta is None:
+        meta = {}
+
+    # The priority order for combining prefix maps are:
+    #  1. Internal prefix map inside the document inside
```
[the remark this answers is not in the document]
cthoyt: thanks

hrshdhgd (approved):
Super clean!

hrshdhgd:
But tests failing.

cthoyt:
@hrshdhgd before I fix the tests, I want to make sure that everyone agrees on the domain logic (i.e., order of priority for metadata and prefix maps)

matentzn:
Will check tomorrow don't merge

cthoyt:
> Will check tomorrow don't merge

We're not at merge yet, but if you can give feedback on the business logic for merging that would be great

tests/test_validate.py
```diff
@@ -36,6 +36,7 @@ def test_validate_json_fail(self):
         In this particular test case, the 'mapping_justification' slot
         does not have EntityReference objects, but strings.
         """
+        # FIXME
```
cthoyt, line 39: @hrshdhgd can you help me debug this?

cthoyt: I was able to figure out what was going on. The test was incorrect. In order to avoid issues like this in the future, it would be good to make tests much more explicit (i.e., check that the error actually is what is expected)

Therefore, I'm skipping this test and adding a more detailed explanation as to why.

cthoyt: New issue opened in https://github.com/mapping-commons/sssom-py/issues/434

cthoyt:
@matentzn all of the unit test issues are now addressed, so this is ready for review (though testing is broken because of poetry, yet again)

src/sssom/writers.py
```diff
@@ -441,7 +441,7 @@ def to_fhir_json(msdf: MappingSetDataFrame) -> Dict:
 def to_json(msdf: MappingSetDataFrame) -> JsonObj:
     """Convert a mapping set dataframe to a JSON object."""
     doc = to_mapping_set_document(msdf)
-    data = JSONDumper().dumps(doc.mapping_set, contexts=doc.prefix_map)
+    data = JSONDumper().dumps(doc.mapping_set, contexts={"@context": doc.prefix_map})
```
cthoyt, line 444: I think it's the case that none of the test data had ever had a properly loaded prefix map, so the fact that this wasn't working before was never caught by any tests

tests/test_collapse.py
```diff
@@ -73,6 +73,7 @@ def test_diff(self):
     def test_reconcile_prefix(self):
         """Test curie reconciliation is performing as expected."""
         msdf = parse_sssom_table(data_dir / "basic3.tsv")
+        msdf.clean_prefix_map()
```
cthoyt, line 76: we now backfill the prefix map with the default one, so it was simpler to remove the unused parts here rather than changing the test to be explicit

tests/test_conversion.py
```diff
@@ -146,9 +146,10 @@ def _test_to_dataframe(self, mdoc, test):
             test.ct_data_frame_rows,
             f"The pandas data frame has less elements than the orginal one for {test.filename}",
         )
-        df.to_csv(test.get_out_file("roundtrip.tsv"), sep="\t")
-        # data = pd.read_csv(test.get_out_file("roundtrip.tsv"), sep="\t")
-        data = parse_sssom_table(test.get_out_file("roundtrip.tsv")).df
+        path = test.get_out_file("roundtrip.tsv")
```
cthoyt, line 149: This makes it so the prefix map is output with the pandas dataframe. Before, none of the test data had properly loaded prefix maps, so it was checking two empty prefix maps were the same. This update makes tests properly output the prefix map with the data.

tests/test_utils.py
```diff
@@ -129,16 +129,10 @@ def test_clean_prefix_map_strict(self):
     def test_clean_prefix_map_not_strict(self):
         """Test clean prefix map with 'strict'=False."""
         msdf = parse_sssom_table(f"{data_dir}/test_clean_prefix.tsv")
-        original_curie_map = msdf.prefix_map
```
cthoyt, line 132: this is removed for the same reason as before, since a lot of extra content gets added during the new loading, and I didn't want to make the test so big and explicit

src/sssom/parsers.py
```diff
@@ -193,44 +194,35 @@ def parse_sssom_table(
...
-                        sssom_metadata[CURIE_MAP][k] = v
-            prefix_map = sssom_metadata[CURIE_MAP]
+    if meta is None:
+        meta = {}
+
+    # The priority order for combining prefix maps are:
+    #  1. Internal prefix map inside the document
+    #  2. Prefix map passed through this function inside the ``meta``
+    #  3. Prefix map passed through this function to ``prefix_map`` (handled with ensure_converter)
+    #  4. Default prefix map (handled with ensure_converter)
```
matentzn, line 204: I never realised how strangely redundant `prefix_map` and `meta[curie_map]` are. It is hard for me in my mind to tease the use cases apart that require these to be separate, which makes it hard to specify a good default order. Your suggestion seems in line with the status quo so I would say lets roll with it!

src/sssom/parsers.py
```diff
@@ -193,44 +194,35 @@ def parse_sssom_table(
...
+    #  1. Metadata appearing in the SSSOM document
+    #  2. Metadata passed through ``meta`` to this function
+    #  3. Default metadata
+    combine_meta = dict(
+        ChainMap(
+            sssom_metadata,
+            meta,
+            get_default_metadata(),
+        )
+    )
```
matentzn, line 223: Again, this appears to be the default, so I am ok with keeping it!

src/sssom/writers.py
matentzn: Is this working now?

src/sssom/parsers.py
```diff
@@ -193,44 +194,35 @@ def parse_sssom_table(
...
-        meta = sssom_metadata
-
-        if PREFIX_MAP_KEY in sssom_metadata:
-            if prefix_map:
-                for k, v in prefix_map.items():
-                    if k in sssom_metadata[CURIE_MAP]:
-                        if sssom_metadata[CURIE_MAP][k] != v:
-                            logging.warning(
-                                f"SSSOM prefix map {k} ({sssom_metadata[CURIE_MAP][k]}) "
-                                f"conflicts with provided ({prefix_map[k]})."
```
matentzn, line 223: Is this warning preserved in curies.chain?

src/sssom/parsers.py
```diff
@@ -193,44 +194,35 @@ def parse_sssom_table(
...
-    # If SSSOM external metadata is provided, merge it with the internal metadata
-
-    if sssom_metadata:
-        if meta:
-            for k, v in meta.items():
-                if k in sssom_metadata:
-                    if sssom_metadata[k] != v:
-                        logging.warning(
-                            f"SSSOM internal metadata {k} ({sssom_metadata[k]}) "
-                            f"conflicts with provided ({meta[k]})."
```
matentzn, line 209: Is the warning preserved in `ChainMap`?

tests/test_conversion.py
matentzn: Thanks!

matentzn (commented):
This looks great, made a few questions and comments. My primary concern right now is roundtripping. For example, what happens right now if someone loads a SSSOM mapping file to, say, annotate it with a new piece of metadata, not having included the SSSOM_BUILT_IN_PREFIXES? Will they just be added to the file that is written out? 

TLDR: the remain thing I want to feel convinced about is that if a legal SSSOM file is passed to `parse_sssom_table`, and then written back out, it remains entirely unchanged. Is that, as far as you can see, the case?

src/sssom/writers.py
cthoyt: Yes.

src/sssom/parsers.py
cthoyt: No. The notion of logging conflicts here was more to support how complicated this code was. The `curies.chain` goes one step further than the previous code and creates a coherent, merged prefix map that loses no information from any of the inputs. Only in a case where there is an irreconcilable difference does `curies.chain` throw an error.

src/sssom/parsers.py
cthoyt: No, by the nature of how ChainMap works, it picks the highest priority and disregards everything else. If requested, I can re-implement identification of discrepancies between the internal metadata and metadata passed to this function, but I think there's some comfort in knowing that ChainMap (and the accompanying docs) means that merging this data follows a clear path now that can be understood by any reader

src/sssom/writers.py
matentzn: Thanks!

src/sssom/parsers.py
matentzn: So when the original sssom file provided map contains `a: http://a.b` and say the `prefix_map` parameter provided map contains `a: http://a.c` it will just maintain the on provided by the sssom file provided map. What happens if the `prefix_map` provided map contains `x: http://a.b`?

src/sssom/parsers.py
```diff
@@ -193,44 +194,35 @@ def parse_sssom_table(
...
+        meta = {}
+
+    # The priority order for combining prefix maps are:
+    #  1. Internal prefix map inside the document
+    #  2. Prefix map passed through this function inside the ``meta``
+    #  3. Prefix map passed through this function to ``prefix_map`` (handled with ensure_converter)
+    #  4. Default prefix map (handled with ensure_converter)
+    converter = curies.chain(
+        [
+            Converter.from_prefix_map(sssom_metadata.pop(CURIE_MAP, {})),
```
matentzn, line 207: what about the built-in prefix map? is there any security to prevent the `sssom_metadata` provided curie_map to not redefine one of the 8 or so built in prefixes?

src/sssom/parsers.py
matentzn: Yeah, I think this makes sense to me, for now;

src/sssom/parsers.py
cthoyt: > So when the original sssom file provided map contains a: http://a.b and say the prefix_map parameter provided map contains a: http://a.c it will just maintain the on provided by the sssom file provided map. 

This results in an EPM with:

```json
[ {"prefix": "a", "uri_prefix": "http://a.b", "uri_prefix_synonyms": ["http://a.c"]} ]
```

> What happens if the prefix_map provided map contains x: http://a.b?

You would get

```json
[ {"prefix": "a", "prefix_synonyms": ["x"], "uri_prefix": "http://a.b"} ]
```

src/sssom/parsers.py
```diff
@@ -193,44 +194,35 @@ def parse_sssom_table(
...
+            Converter.from_prefix_map(meta.pop(CURIE_MAP, {})),
+            ensure_converter(prefix_map),
+        ]
+    )
+
+    # The priority order for combining metadata is:
+    #  1. Metadata appearing in the SSSOM document
+    #  2. Metadata passed through ``meta`` to this function
+    #  3. Default metadata
```
matentzn, line 216: The only downside of this is that you cannot at all _change_ the metadata in the mapping set. A common use case would be to copy a mapping set object, then updating the metadata in there with additional things, replacing, for example the contributor set. Did you check how this approaches behaves with the `sssom merge` method?

src/sssom/parsers.py
matentzn: Ok cool, this makes sense. Thx.

src/sssom/parsers.py
cthoyt: As far as I can tell, `sssom merge` doesn't explicitly do any explicit metadata merging

cthoyt: If you want me to change the priority order, then I am happy to do that. Before, it wasn't really defined in a way that made sense, so this was just my best guess. Maybe we should put the `meta` argument higher so you can override what's in the document, but does that make sense? Shouldn't we assume ahead of time that the document it internally self-consistent? If we want to update the way the data is inside, we can do a combination of reconciliation + (re-)standardization

src/sssom/parsers.py
cthoyt: Yes, I will put that as an explicit first entry.

cthoyt: Done in 55df2c1

cthoyt:
> This looks great, made a few questions and comments. My primary concern right now is roundtripping. For example, what happens right now if someone loads a SSSOM mapping file to, say, annotate it with a new piece of metadata, not having included the SSSOM_BUILT_IN_PREFIXES? Will they just be added to the file that is written out?

@matentzn I added an explicit test to show that outputting a coherent prefix map is working in b73ee19.

I am pretty confident that this makes round-tripping even safer than before, since it always tries to backfill the prefix map with the built-in one.

src/sssom/parsers.py
matentzn: (feel free to click resolve on this comment no further action required)

src/sssom/parsers.py
matentzn: Sweet, yes. Thank you!

src/sssom/parsers.py
matentzn: > As far as I can tell, sssom merge doesn't explicitly do any explicit metadata merging 

I remember now, it should not actually. Brain fart.

>  Shouldn't we assume ahead of time that the document it internally self-consistent?

I cannot currently think through all eventuality just in my mind, but intuitively I think what you have here makes sense. We will probably have to provide explicit methods to update annotations if necessary (and we already have the `sssom annotate` command so that should be covered.

I will resolve this issue for now! Thanks for indulging me!

tests/test_parsers.py
```diff
@@ -245,3 +251,70 @@ def test_parse_obographs_merged(self):
...
+            "object_id",
+            "object_label",
+            "mapping_justification",
+            "creator_id",
+        ]
+        df = pd.DataFrame(rows, columns=columns)
+        msdf = MappingSetDataFrame(df=df, converter=ensure_converter())
+        msdf.clean_prefix_map(strict=True)
+        self.assertEqual(
+            {"DOID", "semapv", "orcid", "skos", "UMLS"}.union(SSSOM_BUILT_IN_PREFIXES),
```
matentzn, line 285:
```suggestion
            {"DOID", "orcid", "UMLS"}.union(SSSOM_BUILT_IN_PREFIXES),
```

matentzn: (semapv and skos are built in prefixes)

src/sssom/context.py
```diff
@@ -52,10 +52,13 @@ def _get_built_in_prefix_map() -> Converter:
 HINT = Union[None, PrefixMap, Converter]
 
 
-def ensure_converter(prefix_map: HINT = None) -> Converter:
+def ensure_converter(prefix_map: HINT = None, *, use_bioregistry: bool = True) -> Converter:
```
matentzn, line 55:
```suggestion
def ensure_converter(prefix_map: HINT = None, *, use_default_prefixmap: bool = True) -> Converter:
```

The "bioregistry" names invites the faulty suggestion that it will actually use bioregistry, rather than a customised context which may or may not be up-to-date in sssom-py.

src/sssom/context.py
```diff
@@ -52,10 +52,13 @@ def _get_built_in_prefix_map() -> Converter:
...
     """Ensure a converter is available."""
     if not prefix_map:
-        return get_converter()
+        if use_bioregistry:
```
matentzn, line 58:
```suggestion
        if use_default_prefixmap:
```

src/sssom/context.py
```diff
@@ -52,10 +52,13 @@ def _get_built_in_prefix_map() -> Converter:
...
+            return get_converter()
```
matentzn, line 59: shouldnt this be a chain of built-in prefix map with get_converter()? I cant see a use case in which we would ever not want the built in prefixes to be added first.

src/sssom/parsers.py
```diff
@@ -193,44 +194,36 @@ def parse_sssom_table(
...
+    #  1. Built-in prefix map
+    #  2. Internal prefix map inside the document
+    #  3. Prefix map passed through this function inside the ``meta``
+    #  4. Prefix map passed through this function to ``prefix_map`` (handled with ensure_converter)
+    converter = curies.chain(
+        [
+            _get_built_in_prefix_map(),
+            Converter.from_prefix_map(sssom_metadata.pop(CURIE_MAP, {})),
+            Converter.from_prefix_map(meta.pop(CURIE_MAP, {})),
+            ensure_converter(prefix_map, use_bioregistry=False),
```
matentzn, line 210:
```suggestion
            ensure_converter(prefix_map, use_default_prefixmap=False),
```

matentzn (commented):
Awesome changes man, a few more comments and I think we are ready to merge!

tests/test_parsers.py
cthoyt: Yes, I have put them here because they are explicitly included. Ultimately, this makes the same set, but I can remove them if it's confusing

src/sssom/context.py
cthoyt: good point

src/sssom/context.py
cthoyt: `get_converter` has the built-in prefix map chained inside it as expected

tests/test_parsers.py
matentzn: If it makes sense to you, its fine, I thought i was just and oversight!

src/sssom/context.py
matentzn: Sweet.

tests/test_parsers.py
cthoyt: Addressed in 4d91007

src/sssom/context.py
cthoyt: Addressed in 20fb6d5

src/sssom/context.py
cthoyt: Addressed in 20fb6d5. I also wrote more detailed documentation for this function.

src/sssom/parsers.py
cthoyt: Addressed in 20fb6d5

matentzn (dismissed):
What a monumental effort. THANK YOU!

matentzn (approved):
@hrshdhgd feel free to merge and make a new release as well!

PR #433, opened 2023-09-26
Test old versions of Pandas

### LINKED ISSUE #432: Consider loosening Pandas dependency to include some 1.x versions
Recently sssom-py started requiring Pandas >= 2.0.3, which is colliding with other tooling we have which has a conservative upper limit on pandas of <2 (or ^1.x.x). In cat-merge, I found that 2.0.x was ok, but 2.1.x was not. I can look at kgx, but if sssom-py doesn't actually require a 2.x version of Pandas, it would be nice to walk back the requirement.

hrshdhgd:
Fixes #432


cthoyt:
@hrshdhgd did you do something magic that makes poetry not break? teach me the ways!

hrshdhgd:
lol! Honestly I didn't do anything. My guess is GitHub fixed things on their end or I just got lucky.

I actually wanted ask you something related to this PR. If I wanted to run tests with 2 versions of `pandas` (<2.0.0 and >=2.0.0), how do I do it?

Using `tox` is the first thing that comes to mind. Something like:
```
[testenv:pandas1]
deps =
  pandas<2.0.0
allowlist_externals =
    poetry
commands =
    poetry run pytest
description = Run unit tests with pytest. This is a special environment that does not get a name, and
              can be referenced with "py".
```

But this won't work since the `pandas` version is dictated by the `poetry.lock` which is not impacted by tox. Any ideas?

cthoyt:
@hrshdhgd you can see in https://github.com/cthoyt/curies/pull/64/files how I did this with a combination of tox and GHA

hrshdhgd:
I thought so. I think that is acceptable for this particular case. Thank you very much @cthoyt !

matentzn:
Lemme know when ready

hrshdhgd:
@matentzn  : Ready to merge

hrshdhgd:
> @hrshdhgd did you do something magic that makes poetry not break? teach me the ways!

@cthoyt : I think this may be the solution. You had it right with `pip install -U pip setuptools` , I just removed the `python -m` and it seemed to fix the error (this time at least! 😄)

cthoyt:
@hrshdhgd that's messed up. this means there are multiple independent versions of python floating around

.github/workflows/qc.yml
```diff
@@ -51,3 +51,11 @@ jobs:
 
       - name: Test with pytest and generate coverage file
         run: poetry run tox -e py
+
+      #----------------------------------------------
+      # QC with pandas 1.x.x
+      #----------------------------------------------
```
cmungall, line 57: add a comment that says that we run `poetry update` frequently so the lock file should in theory have an up to date >2 vesion, which gives us confidence we are testing against 1 and 2

cthoyt:
@hrshdhgd there are lots of places where pytrie is having issues, do you think we should vendor this in `curies`?

.github/workflows/qc.yml
```diff
@@ -51,3 +50,12 @@ jobs:
...
         run: poetry run tox -e py
+
+      #-----------------------------------------------------
+      # QC with pandas 1.x.x
+      # Note: The lock file will always have pandas > 2.0.0
+      #-----------------------------------------------------
+      - name: Test with pandas<2.0.0
+        run: |
+          poetry add "pandas<2.0.0"
+          poetry run pytest
```
matentzn, line 61: Do replicate the condiations above exactly, wouldnt this test have to run is tox?

matentzn (approved):
I approve, just made one non-essential comment @hrshdhgd, trust you on this!

cthoyt:
I would prefer we break these new tests out into different matrix entries - this will make the overall runtime shorter for each build. It might also be okay to reduce the number of different python versions running this test as it's less than mission critical

.github/workflows/qc.yml
hrshdhgd: Shouldn't matter, it is exactly the same. But just for the sake of clarity, I changed it.

PR #435, opened 2023-10-02
Explicitly use Python engine for pandas

cthoyt:
Pandas was giving the following warning:

```
/Users/cthoyt/dev/sssom-py/src/sssom/parsers.py:147: ParserWarning: Falling back to the 'python' engine because the 'c' engine does not support sep=None with delim_whitespace=False; you can avoid this warning by specifying engine='python'.
  df = pd.read_csv(table_stream, sep=sep, dtype=str)
```

This PR explicitly specifies the engine to be "python" as suggested.
Another alternative would be to set the separator to tab by default, since I don't think we allow SSSOM in CSV (or at least we shouldn't :p)

This PR also deletes another superfluous `parse()` function that just wraps `pandas.read_csv()`

### DIFF
--- a/src/sssom/parsers.py
+++ b/src/sssom/parsers.py
@@ -144,7 +144,7 @@ def _read_pandas_and_metadata(input: io.StringIO, sep: str = None):
     table_stream, metadata_stream = _separate_metadata_and_table_from_stream(input)
 
     try:
-        df = pd.read_csv(table_stream, sep=sep, dtype=str)
+        df = pd.read_csv(table_stream, sep=sep, dtype=str, engine="python")
         df.fillna("", inplace=True)
     except EmptyDataError as e:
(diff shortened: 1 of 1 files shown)


matentzn (commented):
parse() is not superfluous as it wraps the two parameters (now three). I would prefer we keep it. Remember the SRP. This PR is actually a perfect example: You want to fix a single parameter, which should only be a single line of code changed. Now, you have to change 13 or more lines..

cthoyt:
Yes, good point. I wanted to make sure there weren't any other places where pandas was getting used and we'd have to worry about this warning. I am not sold that there needs to be a function that is only used for two small tests (which should probably be loaded through the high-level parser anyway), but we can discuss that elsewhere

matentzn (approved):
Hahah you are less confrontational man then suits you! THANKS

PR #437, opened 2023-10-03
Be less draconian on alignment api parsing

matentzn:
This PR:

- Adds wrapper methods for ensuring the validity of an individually constructed mapping from a dictionary and adding that mapping to a list, if valid
- Adding a few tests to ensure the constructing a Mapping() object fails if the data provided is incomplete
- Removes the now redundant _is_valid_mapping method
- Adds tests for ensuring the alignment API parsers don't fail when prefixes are unknown.


src/sssom/parsers.py
```diff
@@ -828,12 +828,15 @@ def _cell_element_values(cell_node, converter: Converter, mapping_predicates) ->
...
 
     mdict[MAPPING_JUSTIFICATION] = MAPPING_JUSTIFICATION_UNSPECIFIED
-
-    m = Mapping(**mdict)
-    if _is_valid_mapping(m):
-        return m
-    else:
-        return None
+    try:
+        m = Mapping(**mdict)
```
matentzn, line 832: Maybe due to some change in LinkML, this line has started throwing an error when the mapping is not wellformed (it also means the below line is not necessary anymore, but we may use it in the future).

In essence, the only thing that happens here is wrapping the `m = Mapping(**mdict)` in try catch so it does not fail hard.

cthoyt: If the below line isn't used anymore, let's delete it. Saving stuff for the future is a big programming trap

tests/test_parsers.py
```diff
@@ -156,6 +157,79 @@ def test_parse_alignment_minidom(self):
...
+        msdf_with_prefixmap = from_alignment_minidom(
+            dom=alignmentxml,
+            prefix_map=prefix_map_with_prefix,
+        )
+
+        msdf_without_prefixmap = from_alignment_minidom(
+            dom=alignmentxml,
+        )
+
+        self.assertEqual(len(msdf_with_broken_prefixmap.df), 1)
```
cthoyt, line 227: just a reminder to make this an explicit test, adding something like:

```python
self.assertEqual(
    ("FBbt_00004924", "skos:exactMatch", "WBbt_0006760"),
    msdf_with_broken_prefixmap.df.iloc[0],
)
```

src/sssom/parsers.py
matentzn: Ok, I added a test now as well that ensures that line _actually_ does as I claim.

src/sssom/parsers.py
```diff
@@ -828,12 +828,15 @@ def _cell_element_values(cell_node, converter: Converter, mapping_predicates) ->
...
+        if _is_valid_mapping(m):
```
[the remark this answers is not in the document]
matentzn: Refactored entirely

tests/test_parsers.py
matentzn: Cool, done. Thanks! Its fun!

tests/test_model.py
```diff
@@ -0,0 +1,30 @@
...
+            subject_id="ID:123"
+        )  # This is missing object_id, predicate_id, mapping_justification
+        mdict_complete = dict(
+            subject_id="AID:123",
+            object_id="BID:123",
+            predicate_id="skos:exactMatch",
+            mapping_justification="semapv:LexicalMatching",
+        )
+
+        with self.assertRaises(ValueError):
```
cthoyt, line 23: can we check that the value error is giving back feedback about what's missing? Also, shouldn't this be a type error? These are probably issues with linkml upstream, so let's just make notes about them

tests/test_model.py
```diff
@@ -0,0 +1,30 @@
...
+            m = Mapping(**mdict_missing)
+
+        m = Mapping(**mdict_complete)
```
cthoyt, line 26: split these two independent tests into two different test functions

src/sssom/util.py
```diff
@@ -1358,3 +1358,34 @@ def safe_compress(uri: str, converter: Converter) -> str:
             f"CURIE appeared where there should be a URI, and could not be standardized: {uri}"
         )
     return rv
+
+
+def ensure_valid_mapping_from_dict(mdict):
```
cthoyt, line 1363: Given the recent discussion about people using stuff that should be marked as private, let's move these functions into the `parsers.py` module and keep them marked as private by having an underscore at the beginning of the name

tests/test_model.py
matentzn: I added a check to that end, but it is not perfect - the LinkML error only returns the first problem, not _all_ problems, in the error message..

tests/test_model.py
matentzn: Done.

tests/test_parsers.py
```diff
@@ -156,6 +157,91 @@ def test_parse_alignment_minidom(self):
...
+            prefix_map=prefix_map_without_prefix,
+        )
+        expected_row_values = [
+            "FBbt:00004924",
+            "skos:exactMatch",
+            "WBbt:0006760",
+            "semapv:UnspecifiedMatching",
+            0.75,
+        ]
+        self.assertEqual(msdf_with_broken_prefixmap.df.iloc[0].tolist(), expected_row_values)
```
cthoyt, line 224: expected value comes first in unit tests

src/sssom/util.py
matentzn: Ok done!

tests/test_parsers.py
matentzn: OOps, fixed.

src/sssom/parsers.py
```diff
@@ -926,3 +915,34 @@ def split_dataframe_by_prefix(
...
+
+
+def _ensure_valid_mapping_from_dict(mdict):
+    """
+    Return a valid mapping object if it can be constructed, else None.
+
+    :param mdict: A dictionary containing the mapping metadata.
+    :return: A valid Mapping object, or None.
+    """
+    try:
```
cthoyt, line 927: this function could set the default mapping justification

```python
mdict[MAPPING_JUSTIFICATION] = MAPPING_JUSTIFICATION_UNSPECIFIED
```

and any other reasonable defaults

matentzn: Fantastic idea.. This will make this a bit more failsafe!

matentzn:
Thank you Charlie, I am very out of practice with coding and that was a fun show! Thanks for your help and education.

src/sssom/parsers.py
```diff
@@ -643,7 +640,9 @@ def from_obographs(
                                     mdict[SUBJECT_LABEL] = label
                                     mdict[PREDICATE_ID] = "oboInOwl:hasDbXref"
                                     mdict[MAPPING_JUSTIFICATION] = MAPPING_JUSTIFICATION_UNSPECIFIED
-                                    mlist.append(Mapping(**mdict))
+                                    mapping = _ensure_valid_mapping_from_dict(mdict)
```
cthoyt, line 643: can you use `_add_valid_mapping_to_list` here?

src/sssom/parsers.py
```diff
@@ -411,7 +411,9 @@ def from_sssom_dataframe(
     bad_attrs: typing.Counter[str] = Counter()
     for _, row in df.iterrows():
         mdict, bad_attrs = _get_mdict_ms_and_bad_attrs(row, bad_attrs)
-        mlist.append(_prepare_mapping(Mapping(**mdict)))
+        mapping = _ensure_valid_mapping_from_dict(mdict)
+        if mapping:
+            mlist.append(_prepare_mapping(mapping))
```
cthoyt, line 416: maybe make the application of `_prepare_mapping` be a flaggable part of `_add_valid_mapping_to_list`

src/sssom/parsers.py
matentzn: Oversight, done.

src/sssom/parsers.py
matentzn: Yes, and I refactored it also to make more explicit what this method does.

src/sssom/parsers.py
```diff
@@ -926,3 +901,39 @@ def split_dataframe_by_prefix(
...
+def _add_valid_mapping_to_list(mdict, mlist, flip_superclass_assertions=False):
+    """
+    Validate the mapping and append to the list if valid.
+
+    Parameters:
+    - mdict (dict): A dictionary containing the mapping metadata.
+    - mlist (list): The list to which the valid mapping should be appended.
+    """
+    mapping = _ensure_valid_mapping_from_dict(mdict)
+    if mapping:
```
cthoyt, line 935: I am a fan of limiting the indentation, so this can be done in two ways:

1. Flip `if mapping` to `if not mapping: return None`, then you don't have to indent for all of this
2. Rather than repeating `mlist.append` twice, you can just do 
   ```python
   if flip_superclass_assertions:
       mapping = _flip_superclass_assertion(mapping)
   ```
   and have `mlist.append(mapping)` with no indendation

matentzn: Cool, done! Thank you Sensei!

src/sssom/parsers.py
```diff
@@ -926,3 +901,40 @@ def split_dataframe_by_prefix(
             df_subset, prefix_map=dict(subconverter.bimap), meta=meta
         )
     return split_to_msdf
+
+
+def _ensure_valid_mapping_from_dict(mdict):
```
cthoyt, line 906: type annotations?

cthoyt (changes requested):
After the 3 comments I made this is ready

src/sssom/parsers.py
```diff
@@ -926,3 +901,40 @@ def split_dataframe_by_prefix(
...
+
+    try:
+        m = Mapping(**mdict)
+        return m
+    except ValueError as e:
+        logging.warning(
+            f"One mapping in the mapping set is not well-formed, "
+            f"and therfore not included in the mapping set ({mdict}). Error: {e}"
+        )
+    return None
```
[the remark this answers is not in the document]
matentzn: I dont know what you mean by "else clause" here but I did:

```
    try:
        m = Mapping(**mdict)
    except ValueError as e:
        logging.warning(
            f"One mapping in the mapping set is not well-formed, "
            f"and therfore not included in the mapping set ({mdict}). Error: {e}"
        )
        return None
    return m
```

src/sssom/parsers.py
matentzn: Done

src/sssom/parsers.py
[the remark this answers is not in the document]
cthoyt: Yes, that works. An alternative is

```python
    try:
        m = Mapping(**mdict)
    except ValueError as e:
        logging.warning(
            f"One mapping in the mapping set is not well-formed, "
            f"and therfore not included in the mapping set ({mdict}). Error: {e}"
        )
        return None
    else:
        return m
```

src/sssom/parsers.py
```diff
@@ -740,9 +726,8 @@ def get_parsing_function(input_format: Optional[str], filename: str) -> Callable
     return func
 
 
-def _prepare_mapping(mapping: Mapping) -> Mapping:
-    p = mapping.predicate_id
-    if p == "sssom:superClassOf":
+def _flip_superclass_assertion(mapping: Mapping) -> Mapping:
+    if mapping.predicate_id == "sssom:superClassOf":
```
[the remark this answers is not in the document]
matentzn: Awesome!

src/sssom/parsers.py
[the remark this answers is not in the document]
matentzn: Interesting, I never heard of that.. You can only learn this is PR reviews from experienced coders I assume..

src/sssom/parsers.py
```diff
@@ -727,10 +727,11 @@ def get_parsing_function(input_format: Optional[str], filename: str) -> Callable
 
 
 def _flip_superclass_assertion(mapping: Mapping) -> Mapping:
-    if mapping.predicate_id == "sssom:superClassOf":
+    if mapping.predicate_id != "sssom:superClassOf":
+        return mapping
+    else:
```
cthoyt, line 732: no else needed here, now you can get rid of indentation

cthoyt (approved):
Amazing! Thank you!

PR #438, opened 2023-10-04
Standardize MSDF

cthoyt:
The functionality for standardizing a MSDF was already built in to the `reconcile_prefix_and_data()` function. This PR does the following:

1. Makes this a re-usable instance-level function of an MSDF
2. Adds similar functionality for standardizing the metadata of the MSDF
3. Add explicit unit tests!

### DIFF
--- a/src/sssom/util.py
+++ b/src/sssom/util.py
@@ -109,6 +109,39 @@ class MappingSetDataFrame:
         """Clean up the context."""
         self.converter = curies.chain([_get_built_in_prefix_map(), self.converter])
 
+    def _standardize_curie_or_iri(self, curie_or_iri: str) -> str:
+        """Standardize a CURIE or IRI, returning the original if not possible.
+
+        :param curie_or_iri: Either a string representing a CURIE or an IRI
+        :returns:
+            - If the string represents an IRI, tries to standardize it. If not possible, returns the original value
+            - If the string represents a CURIE, tries to standardize it. If not possible, returns the original value
+            - Otherwise, return the original value
+        """
+        if is_iri(curie_or_iri):
+            return self.converter.standardize_uri(curie_or_iri) or curie_or_iri
+        if is_curie(curie_or_iri):
+            return self.converter.standardize_curie(curie_or_iri) or curie_or_iri
+        return curie_or_iri
+
+    def standardize(self) -> None:
+        """Standardize this MSDF's dataframe and metadata with respect to its converter."""
+        self._standardize_metadata()
+        self._standardize_df()
+
+    def _standardize_df(self) -> None:
+        """Standardize this MSDF's dataframe with respect to its converter."""
+        for column, values in _get_sssom_schema_object().dict["slots"].items():
+            if values["range"] != "EntityReference":
+                continue
+            if column not in self.df.columns:
+                continue
+            self.df[column] = self.df[column].map(self._standardize_curie_or_iri)
+
+    def _standardize_metadata(self) -> None:
+        """Standardize this MSDF's metadata with respect to its converter."""
+        raise NotImplementedError
+
     def merge(self, *msdfs: "MappingSetDataFrame", inplace: bool = True) -> "MappingSetDataFrame":
         """Merge two MappingSetDataframes.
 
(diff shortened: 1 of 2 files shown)


matentzn (commented):
There are many ways a mapping set can be standardised - can we refactor this code to something more explicit like `standardise_entity_references` or something along these lines?

cthoyt:
@matentzn done in 54f197d

src/sssom/util.py
```diff
@@ -109,6 +109,25 @@ def clean_context(self) -> None:
...
         self.converter = curies.chain([_get_built_in_prefix_map(), self.converter])
 
+    def standardize_references(self) -> None:
+        """Standardize this MSDF's dataframe and metadata with respect to its converter."""
+        self._standardize_metadata_references()
+        self._standardize_df_references()
+
+    def _standardize_df_references(self) -> None:
+        """Standardize this MSDF's dataframe with respect to its converter."""
+        func = partial(_standardize_curie_or_iri, converter=self.converter)
```
matentzn, line 119: What on earth is a `partial()` is this like black magic?

src/sssom/util.py
```diff
@@ -194,6 +213,54 @@ def remove_mappings(self, msdf: "MappingSetDataFrame") -> None:
...
+        - Otherwise, return the original value
+    """
+    if is_iri(curie_or_iri):
+        return converter.standardize_uri(curie_or_iri) or curie_or_iri
+    if is_curie(curie_or_iri):
+        return converter.standardize_curie(curie_or_iri) or curie_or_iri
+    return curie_or_iri
+
+
+def _standardize_metadata(converter, metadata) -> None:
```
matentzn, line 232: this method does much more than standardising. Its more like a repair! For example, while you cant use a | separate in a URL, I _think_ you can use it in an IRI. But well, mooooot, I guess if you start promoting IRI grammar rules we will get into hells kitchen. So yea, probably ok.

matentzn: (You can ignore this comment, nothing actionable)

tests/test_utils.py
```diff
@@ -194,3 +195,91 @@ def test_get_prefixes(self):
...
+    def test_standardize_coerce_list(self):
+        """Test that a metadata field that should not be a list gets coerced to a single element."""
+        metadata = {"mapping_registry_id": ["https://example.org/r1"]}
+        msdf = MappingSetDataFrame(df=pd.DataFrame(), converter=Converter([]), metadata=metadata)
+        msdf._standardize_metadata_references()
+        self.assertEqual({"mapping_registry_id": "https://example.org/r1"}, msdf.metadata)
+
+    def test_standardize_metdata_delete_empty(self):
+        """Test that an element that should not be a list but is empty just gets deleted."""
+        metadata = {"mapping_registry_id": []}
```
matentzn, line 275: `mapping_registry_id` is not a mapping set element. Use `mapping_set_id` instead.

tests/test_utils.py
```diff
@@ -194,3 +195,91 @@ def test_get_prefixes(self):
...
+    def test_standardize_metadata_multivalued_type_error(self):
+        """Test raising on a  non-string, non-list object given to a multivalued slot."""
+        metadata = {"creator_id": object()}
+        msdf = MappingSetDataFrame(df=pd.DataFrame(), converter=Converter([]), metadata=metadata)
+        with self.assertRaises(TypeError):
+            msdf._standardize_metadata_references()
+
+    def test_standardize_error_non_list(self):
+        """Test that a type error is raised when metadata is presented that should not be a list."""
+        metadata = {"mapping_registry_id": ["https://example.org/r1", "https://example.org/r2"]}
```
matentzn, line 261: `mapping_registry_id` is not a mapping set element. Use `mapping_set_id` instead.

tests/test_utils.py
matentzn:
```suggestion
        metadata = {"mapping_set_id": []}
```

matentzn: (you do this multiple times so I let you fix this on your branch)

src/sssom/util.py
cthoyt: no way, `partial` is like one of the best tools in Python. It makes working with functions where you want to keep something constant much easier. This is similar to `func = lambda curie_or_iri: _standardize_curie_or_iri(curie_or_iri, converter=self.converter)`

tests/test_utils.py
cthoyt: oof, changing this breaks all of the tests since `mapping_set_id` has different properties. Is this necessary?

src/sssom/util.py
matentzn: Nice!

tests/test_utils.py
matentzn: Well, it is not at all in the schema -  don't you want at least to test an element that is in the schema?

https://mapping-commons.github.io/sssom/MappingSet/

cthoyt: oh, I see.

cthoyt: @matentzn great feedback, thank you. I updated the implementation to take this into account in e16db81 (also with tests)

src/sssom/util.py
```diff
@@ -194,6 +215,69 @@ def remove_mappings(self, msdf: "MappingSetDataFrame") -> None:
...
+    if is_curie(curie_or_iri):
+        return converter.standardize_curie(curie_or_iri) or curie_or_iri
+    return curie_or_iri
+
+
+def _standardize_metadata(
+    converter: Converter, metadata: Dict[str, Any], *, raise_on_invalid: bool = False
+) -> None:
+    schema_object = _get_sssom_schema_object()
+    slots_dict = schema_object.dict["slots"]
```
matentzn, line 238: Not that it matters here for this PR, but "get_slot_list()" should probably be refactored (schema_object is not needed again)

matentzn (approved):
LGTM!

src/sssom/util.py
cthoyt: Yes, lots of stuff related to `_get_sssom_schema_object()` is begging for refactoring

matentzn (approved):

PR #440, opened 2023-10-04
Get MSDF from LinkML objects

### LINKED ISSUE #439: What to do when I have a list of `Mapping` objects?
I can't find anywhere in sssom-py a utility function that will take a list of `Mapping` objects and turn them into either a MSDF, MappingSet, or MappingSetDoc.

OAK implemented its own janky solution for this, but we should provide something first-party

https://github.com/INCATools/ontology-access-kit/blob/75940bfa883001afb0e4aebb339fd62583c13844/src/oaklib/utilities/lexical/lexical_indexer.py#L313-L325

cthoyt:
Closes #439

This PR adds three methods to the `MappingSetDataFrame` class that allow instantiation from LinkML objects such as `MappingSetDocument`, `MappingSet`, or a list of `Mapping`.

It has the benefit of pre-baking in logic for dealing with default metadata and converters, meaning that nobody has to roll this themselves downstream

This is going to help us address https://github.com/INCATools/ontology-access-kit/blob/75940bfa883001afb0e4aebb339fd62583c13844/src/oaklib/utilities/lexical/lexical_indexer.py#L313-L325, specifically in https://github.com/INCATools/ontology-access-kit/issues/664


src/sssom/util.py
```diff
@@ -105,6 +105,59 @@ def with_converter(
...
         )
 
+    @classmethod
+    def from_mappings(
+        cls,
+        mappings: List[SSSOM_Mapping],
+        *,
+        converter: HINT = None,
+        metadata: Optional[MetadataType] = None,
+    ) -> "MappingSetDataFrame":
```
matentzn, line 115: Is this a trick so you can specify the return type (for use in code quality checking tools) without importing it?

src/sssom/util.py
```diff
@@ -105,6 +105,59 @@ def with_converter(
...
+        """Instantiate from a list of mappings, mapping set metadata, and an optional converter."""
+        metadata = dict(ChainMap(metadata or {}, get_default_metadata()))
```
matentzn, line 117: Because this statement is a bit hard to read (I am used to it now), can we factor it into a private `_prepare_metadata()` method?

src/sssom/util.py
```diff
@@ -105,6 +105,59 @@ def with_converter(
...
+        :returns: A mapping set dataframe
+        """
+        doc = MappingSetDocument(converter=ensure_converter(converter), mapping_set=mapping_set)
+        return cls.from_mapping_set_document(doc)
+
+    @classmethod
+    def from_mapping_set_document(cls, doc: MappingSetDocument) -> "MappingSetDataFrame":
+        """Instantiate from a mapping set document."""
+        if doc.mapping_set.mappings is None:
+            return cls(df=pd.DataFrame(), converter=doc.converter)
```
matentzn, line 139: what is this cls?

matentzn: Maybe some strange magic that refers to the parent class.

matentzn (commented):
I will have to rely on your own self-critical assessment of any risks involved here.. I checked it line by line, and it looks good to me, but I cannot mentally parse all of the details.. 

Are there any risks involved?

src/sssom/util.py
cthoyt: yes, since the name of the class isn't avaliable in the global namespace until after it's parsed. You can also do `from __future__ import annotations` to get around this too

src/sssom/util.py
cthoyt: Class methods allow access to the class. So if you want to make an instance of the class from a classmethod, you can refer to `cls` instead of writing the name of the class. This makes it more impervious to tricky inheritance.

cthoyt:
@matentzn only risks is that there are more convoluted ways people will think of on how they want to provide metadata, besides that, this is pretty straightforward.

cthoyt:
@matentzn I added a round-trip test that demonstrates this works

src/sssom/util.py
cthoyt: Addressed in 28e67e2

src/sssom/util.py
matentzn: Cool!

matentzn (approved):
Everything looks fine to me!

PR #442, opened 2023-10-04
Remove "None" checks for df and metadata

cthoyt:
Both of these fields are now non-optional, so this reduces a lot of unnecessary code for checking that


matentzn (approved):
LGTM

PR #444, opened 2023-10-04
Make MSDF with default converter

cthoyt:
This makes it possible to create a MSDF with only a pandas dataframe

```python
from sssom import MappingSetDataFrame

df = ... # some pandas dataframe
msdf = MappingSetDataFrame(df=df)

# align to default converter
msdf.standardize_references() 
```


matentzn (approved):
fantastic, thank you!

PR #446, opened 2023-10-05
Remove duplicate definition of schema constant

cthoyt:
`sssom.constants.CURIE_MAP` was redefined as `sssom.util.PREFIX_MAP`. This PR switches all of the second to use the first.

### DIFF
--- a/src/sssom/util.py
+++ b/src/sssom/util.py
@@ -27,6 +27,7 @@ from .constants import (
     COLUMN_INVERT_DICTIONARY,
     COMMENT,
     CONFIDENCE,
+    CURIE_MAP,
     MAPPING_JUSTIFICATION,
     MAPPING_SET_ID,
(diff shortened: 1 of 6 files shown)


src/sssom/util.py
```diff
@@ -71,9 +72,6 @@
 
 logging = _logging.getLogger(__name__)
 
-#: The key that's used in the YAML section of an SSSOM file
-PREFIX_MAP_KEY = "curie_map"
```
matentzn, line 75: This line here is my only "concern". Could this be changed to PREFIX_MAP_KEY = CURIE_MAP for backwards compatibility, or not worth it? I leave the judgement to you.

cthoyt: The issue is the proliferation of constants, so I would rather not

matentzn (approved):
Great

PR #447, opened 2023-10-05
Cleanup Metadata usages

cthoyt:
- Standardizes additional handling of prefix maps and metadata
- Deletes the no longer used `Metadata` class
- Deletes `sssom.typehints` module (which was not named aptly for how it was used) and moves its remaining functionality into `sssom.constants`


src/sssom/constants.py
```diff
@@ -271,3 +272,41 @@ def _get_sssom_schema_object() -> SSSOMSchemaView:
 
 SSSOM_URI_PREFIX = "https://w3id.org/sssom/"
 DEFAULT_LICENSE = f"{SSSOM_URI_PREFIX}license/unspecified"
+
+#: The type for metadata that gets passed around in many places
+MetadataType = Dict[str, Any]
+
+
+def generate_mapping_set_id() -> str:
```
matentzn, line 280: Is it ok to have methods in the `constants` file?

src/sssom/io.py
```diff
@@ -134,34 +135,25 @@ def split_file(input_path: str, output_directory: Union[str, Path]) -> None:
...
-        return Metadata.default()
+        return get_converter(), get_default_metadata()
 
     with Path(metadata_path).resolve().open() as file:
         metadata = yaml.safe_load(file)
-    if not metadata.get("mapping_set_id"):
-        metadata["mapping_set_id"] = generate_mapping_set_id()
-    if not metadata.get("license"):
-        metadata["license"] = DEFAULT_LICENSE
-        logging.warning(f"No License provided, using {DEFAULT_LICENSE}")
```
matentzn, line 155: While your ChainMap is beautiful in terms of code readability, it does do silent repairs. I am not saying no, just pointing out that the way you have it, `sssom-py` will add certain prefixes and metadata without even informing the user that it happened.. Is there a way we can make ChainMap a bit more chatty somehow, or is it out of our control? Basically some way to report back if a piece of metadata or a prefix was added that was not previously there?

matentzn (commented):
No larger concern, just 1 or 2 questions.

src/sssom/constants.py
cthoyt: I see the inconsistency here, but I think it outweights the maze of imports you have to go through to find things to keep these closely related parts together.

src/sssom/io.py
cthoyt: ChainMap is implemented in C, so there's nothing we can do directly.

if desired, I can implement a chatty chain map. But, at the same time, we should reconsider why we're adding dummy values to a required field. Shouldn't we require users to for real think about this and supply something themselves? Or if we don't think so, why not just allow it to be none?

src/sssom/constants.py
matentzn: Ok!

src/sssom/io.py
matentzn: I am trying to think of an adequate precedence. For example, if you save an ontology in protege and you don't give it an IRI, it will just add one. Maybe it is ok, for these two values. If someone complains we can reconsider this! Thanks for your thought!

matentzn (approved):

PR #452, opened 2023-10-07
Refactor `_extract_global_metadata()`

cthoyt:
This PR does 3 things:

1. Makes `_extract_global_metadata()` private
2. Removes the part in the definition where it adds a `CURIE_MAP` since this is subsequently popped off
3. Updates the return type to be more accurate


matentzn (approved):
LGTM

PR #453, opened 2023-10-07
Directly unpack metadata tuple

cthoyt:
This PR makes it much more direct how the Metadata tuple object is unpacked. Additionally, the need to expose a "prefix_map" is no longer there, so this is removed.


matentzn (approved):
LGTM!

PR #454, opened 2023-10-07
Update type hints for prefix map merging

cthoyt:
This PR does the following:

1. Creates a literal type for the 3 different prefix map merge modes
2. Updates type hints based on that


src/sssom/constants.py
```diff
@@ -37,14 +37,10 @@
...
-PREFIX_MAP_MODE_METADATA_ONLY = "metadata_only"
-PREFIX_MAP_MODE_SSSOM_DEFAULT_ONLY = "sssom_default_only"
-PREFIX_MAP_MODE_MERGED = "merged"
-PREFIX_MAP_MODES = [
-    PREFIX_MAP_MODE_METADATA_ONLY,
-    PREFIX_MAP_MODE_SSSOM_DEFAULT_ONLY,
-    PREFIX_MAP_MODE_MERGED,
-]
+MergeMode = Literal["metadata_only", "sssom_default_only", "merged"]
+PREFIX_MAP_MODE_METADATA_ONLY: MergeMode = "metadata_only"
```
matentzn, line 41: Never seen this syntax. What does this do, when considering line 40? Looks redundant at first glance (twice the same assignment)

src/sssom/io.py
```diff
@@ -59,7 +60,7 @@ def parse_file(
     output: TextIO,
     input_format: Optional[str] = None,
     metadata_path: Optional[str] = None,
-    prefix_map_mode: Optional[str] = None,
+    merge_mode: Optional[MergeMode] = None,
```
matentzn, line 63: All borders have fallen now, but renaming parameters to central methods like this is really annoying for users, but the change makes sense, so lets just break and mend it all now.

src/sssom/io.py
```diff
@@ -159,20 +160,20 @@ def get_metadata_and_prefix_map(
     else:
         prefix_map = {}
     converter = Converter.from_prefix_map(prefix_map)
-    converter = _merge_converter(converter, prefix_map_mode=prefix_map_mode)
+    converter = _merge_converter(converter, mode=merge_mode)
```
matentzn, line 163: Why is `merge_mode` called `mode` in _merge_converter? Because the context is clear?

matentzn (commented):
Just a few questions, nothing actionable I think

src/sssom/constants.py
cthoyt: This makes checking that one of the constants is used properly is checked by type checking. It further builds the string enumeration into the automatic documentation generation

src/sssom/io.py
cthoyt: I have a feeling this one is unused - it is only part of a method that should itself be private. In fact, I would rather remove this option completely and have it automatically chain on all instances.

src/sssom/io.py
cthoyt: Oversight. I was planning on deleting this method next Pr anyway since it’s only used once inside another private method

cthoyt:
@matentzn actually, I would much rather delete this functionality entirely and only support the `PREFIX_MAP_MODE_MERGED` mode. Given we spent lots of effort to make the `curies` integration, this option is much more powerful than before and I can't see why one would choose either of the others.

matentzn:
We should be able at CLI time to allow an option to "fail if a prefix is not declared", like a "strict parsing" mode. This prevents cases like completely automated systems that accidentally infer a completely wrong prefix map for a dataset. I'd prefer to continue supporting that functionality, but I would be ok to move MERGED mode as default..

src/sssom/io.py
cthoyt: I reverted this and removed the renames

cthoyt:
@matentzn I simplified this PR and took all of the questionable parts out. Now it's just about using type hints better.

PR #455, opened 2023-10-07
Consolidate mapping metadata setting

cthoyt:
This PR consolidates metadata handling functionality. Mainly it does two things:

1. Use a chainmap to ensure default metadata in `_init_mapping_set()`
2. Calls to `_set_metadata_in_mapping_set` to be directly inside `_init_mapping_set()`

Unfortunately, `_set_metadata_in_mapping_set` can't be completely deleted since it's used in one other place, but this is still a good reduction of complexity.


matentzn (approved):
LGTM

PR #459, opened 2023-11-04
Improve CURIE and IRI validation

cthoyt:
Now that `curies` implements CURIE and IRI validation checks in https://github.com/cthoyt/curies/pull/90, this PR replaces the SSSOM functionality.

Note, this requires passing a Converter to a few functions that previously didn't need one, but it seems like a small price to pay for data integrity

This PR also adds deprecation warnings for `sssom.util.is_curie` and `sssom.util.is_iri` since other packages might be using these.


src/sssom/util.py
```diff
@@ -294,10 +295,11 @@ def _standardize_curie_or_iri(curie_or_iri: str, *, converter: Converter) -> str
         - If the string represents a CURIE, tries to standardize it. If not possible, returns the original value
         - Otherwise, return the original value
     """
-    if is_iri(curie_or_iri):
-        return converter.standardize_uri(curie_or_iri) or curie_or_iri
-    if is_curie(curie_or_iri):
-        return converter.standardize_curie(curie_or_iri) or curie_or_iri
+    if converter.is_uri(curie_or_iri):
+        # TODO switch to compress
+        return converter.standardize_uri(curie_or_iri, strict=True)
```
matentzn, line 300: Shouldn't it be strict=False to replicate the original functionality?

src/sssom/util.py
```diff
@@ -1101,6 +1103,11 @@ def is_curie(string: str) -> bool:
 
 def is_iri(string: str) -> bool:
```
matentzn, line 1104: Missing `@deprecated` annotator?

src/sssom/util.py
```diff
@@ -1101,6 +1103,11 @@ def is_curie(string: str) -> bool:
...
     """Check if the string is an IRI."""
+    warnings.warn(
+        "Use :meth:`curies.Converter.is_uri` via a well-defined Converter instance",
+        DeprecationWarning,
+        stacklevel=2,
+    )
     return validators.url(string)
```
matentzn, line 1111: Maybe replace this by `curies.converter.is_uri()` to ensure consistent behaviour (not removing the method, wrapping the converter method)?

matentzn:
Aweeeesome! Thanks!

src/sssom/util.py
cthoyt: if it already passes the `converter.is_uri` check, then it can indeed be standardized. The strict flag makes the typing work better.

matentzn: Duuuuuh. Sorry you are right!

src/sssom/util.py
cthoyt: it's actually only used in one place, where we need global checking for URI-ness. I just made an update to make this all more clear

src/sssom/util.py
cthoyt: I don't think there are any remaining instances of using `@deprecated` in this package. Is there a benefit to using that instead of just giving a clear warning?

matentzn: Yes, documentation generation! Also, technically, `@deprecated` should automatically produce the warning you added manually, leading to less code clutter

src/sssom/util.py
matentzn: Ok, thank you!

src/sssom/util.py
cthoyt: Done in 1654b4a

PR #467, opened 2023-11-20
Reduce three dependencies

cthoyt:
This PR reduces three dependencies:

1. Removes the backport of `importlib.metadata`. Since SSSOM-py is now py38+, this is part of the standard library
2. Hides the import for networkx inside the only function that uses it
3. Hides the import for pansql inside the only function that uses it

Both of the hides are for seemingly specific functionality, so I don't think this will bother anyone.


src/sssom/cliques.py
```diff
@@ -27,9 +27,13 @@
 from .sssom_document import MappingSetDocument
 from .util import MappingSetDataFrame
 
+if TYPE_CHECKING:
+    import networkx
 
-def to_digraph(msdf: MappingSetDataFrame) -> nx.DiGraph:
+def to_digraph(msdf: MappingSetDataFrame) -> "networkx.DiGraph":
```
matentzn, line 33: We really do not need this method at all, but its a good first step to remove the networkx dependency

matentzn (approved):

src/sssom/cliques.py
cthoyt: I'm not against having a first-party export function to convert SSSOM mappings into a graph I think this is valuable, but having it mixed up with the cliques.py code is not great

matentzn:
@hrshdhgd do you see anything problematic here?

hrshdhgd (approved):
Looks good to me.

matentzn (approved):

PR #468, opened 2023-11-20
Fix test after EPM update

matentzn:
The reason for the increase in the count here is that the EPM caught at least 1 more URI prefix during the parsing process


cthoyt (changes requested):
These numeric tests are truly awful - prone to issues and also completely opaque as to what's actually being tested. It would be great if we could make this more explicit

matentzn:
Not a priority. Harshads backlog is 20 tickets; mine is about 50 pull request reviews and emails.. and we need to fix the master branch. Sorry :/

cthoyt:
Something to keep in mind for next project starting from scratch

PR #472, opened 2023-11-28
Use converter during obographs parsing instead of hardcode oboInOwl:hasDbXref

hrshdhgd:


matentzn (approved):

src/sssom/parsers.py
```diff
@@ -607,7 +607,9 @@ def from_obographs(
                                     mdict[SUBJECT_ID] = safe_compress(nid, converter)
                                     mdict[OBJECT_ID] = safe_compress(xref_id, converter)
                                     mdict[SUBJECT_LABEL] = label
-                                    mdict[PREDICATE_ID] = "oboInOwl:hasDbXref"
+                                    mdict[PREDICATE_ID] = converter.compress(
```
cthoyt, line 610: might save a bit of time to store this as a variable at the beginning of the function rather than running compress over and over

matentzn:
Thanks Charlie you are absolutely right, @hrshdhgd  would this be easy to do? Don't sweat it but if it's two lines of code I would kinda like the unnecessary duplication to go..

PR #473, opened 2023-11-28
What to do when having a uri-prefix clash?

matentzn:
Note: If you remove either one of the two prefixes, it works

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -5,7 +5,7 @@ import unittest
 import numpy as np
 import pandas as pd
 import yaml
-from curies import Converter, Record
+from curies import Converter, Record, chain
 from sssom_schema import Mapping as SSSOM_Mapping
 
 from sssom.constants import (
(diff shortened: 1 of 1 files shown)


cthoyt:
Okay this is a good demo. Do you think it will be possible to come up with a consistent set of rules for this kind of merge where in your higher priority EPM, you have two disparate records that a lower priority EPM in the chain would merge together?

matentzn:
@hrshdhgd lets make sure that the test fails (assertRaises) and maybe wrap the exception in something more progressive that helps users actually dealing with the clash.

Also we should make sure that the raised exception mentions the clash. I don't know how hard/easy that is, but for now I would just make an issue in the curies repo asking for a messages that "involves all prefixes that are part of the clash", not just the last one.

cthoyt:
sorry all I have not had time to write an excellent tutorial on this yet. I would like you to let `curies` do the hard work of documenting this to make sure that it's absolutely excellent. Contributions welcome if you want to get started solving this issue upstream in a PR there.

hrshdhgd:
@matentzn : Test should pass now.

matentzn (commented):
Approved!

Thanks!

PR #474, opened 2023-11-29
Refactor OBO Graph JSON parser

cthoyt:
This PR does the following:

1. massively reduces indentation
2. renames variables in a nice way
3. refactors out shared code into a helper function
4. It also changes so it doesn't add an empty string for subject or object label, and instead doesn't add anything

### DIFF
--- a/src/sssom/parsers.py
+++ b/src/sssom/parsers.py
@@ -641,21 +641,21 @@ def _make_mdict(
     try:
         subject_curie = safe_compress(subject_id, converter)
     except ValueError as e:
-        logging.debug("could not parse subject: %s", subject_id)
+        logging.debug("could not parse subject %s - %s", subject_id, e)
     else:
         mdict[SUBJECT_ID] = subject_curie
 
     try:
         predicate_curie = safe_compress(predicate_id, converter)
     except ValueError as e:
-        logging.debug("could not parse object: %s", object_id)
+        logging.debug("could not parse object %s - %s", object_id, e)
     else:
         mdict[PREDICATE_ID] = predicate_curie
 
     try:
         object_curie = safe_compress(object_id, converter)
     except ValueError as e:
-        logging.debug("could not parse object: %s", object_id)
+        logging.debug("could not parse object %s - %s", object_id, e)
     else:
         mdict[OBJECT_ID] = object_curie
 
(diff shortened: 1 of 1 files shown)


src/sssom/parsers.py
```diff
@@ -68,6 +68,8 @@
 
 logging = _logging.getLogger(__name__)
 
+DBXREF_URI = "http://www.geneontology.org/formats/oboInOwl#hasDbXref"
```
matentzn, line 71: This should be imported from constants.py (probably its already there)

src/sssom/parsers.py
```diff
@@ -566,118 +568,104 @@ def from_obographs(
...
+                continue
+            mdict = _make_mdict(edge["sub"], predicate_uri, edge["obj"], converter, labels)
+            _add_valid_mapping_to_list(mdict, mlist)
+
+        if OWL_EQUIV_CLASS_URI in mapping_predicates:
+            for equivalents in graph.get("equivalentNodesSets", []):
+                node_ids = equivalents.get("nodeIds")
+                if not node_ids:
+                    continue
+                for subject_uri, object_uri in itt.product(node_ids, repeat=2):
```
matentzn, line 618: What is itt?

matentzn (commented):
Beautiful refactoring of my terrible code 🗡️ 

I hope the tests are strong enough to ensure nothing is lost. 

@hrshdhgd can you check that when using this branch, mondo.sssom.tsv is still generated correctly (i.e. no rows are lost compared to normal sssom-py)

src/sssom/parsers.py
cthoyt: itertools. Importing as `itt` is very common

src/sssom/parsers.py
cthoyt: Done in 6c09ed3

src/sssom/parsers.py
matentzn: Thanks

hrshdhgd:
[Just ran it](https://github.com/monarch-initiative/mondo/pull/6872/commits/2bfcad003e18bb12203c89d7ccc3e8008946d600) and the differences aren't major.

matentzn (approved):
Thank you @cthoyt for this PR, and @hrshdhgd for checking!

PR #480, opened 2023-12-14
Improve testing of round trip serialization

hrshdhgd:
This fixes: https://github.com/mapping-commons/sssom/issues/321

So the issue was when a tsv file is `sssom convert`ed to a JSON file and then the same JSON file is `sssom parse`d back, there was an error thrown : 
```
ValueError: ['Some prefixes'] are used in the SSSOM mapping set but it does not exist in the prefix map
``` 
This happened because the `@context` block of the JSON file which contained the prefix map was never considered while forming the `prefix_map` and the metadata items (`mapping_set_id`, `license` etc) were also ignored. In addition to this, an absence of the external metadata file path param made it get the default metadata and prefix map internally which is a clear bug in the system.

This PR resolves that issue.


src/sssom/parsers.py
```diff
@@ -27,8 +27,11 @@
     CONFIDENCE,
     CURIE_MAP,
     DEFAULT_MAPPING_PROPERTIES,
+    JSON_CONTEXT_KEY,
```
matentzn, line 30: A key like this which is part of a format (JSON-LD, rather than a specific data model like SSSOM) does IMO not necessarily have to be places in a variable; I don't think this is an issue, just saying that I assume there will be some rule against it.

src/sssom/parsers.py
```diff
@@ -247,22 +250,35 @@ def parse_sssom_rdf(
...
+    # Initialize meta if it's None
+    if meta is None:
+        meta = {}
+
+    # Update metadata with values from JSON document
+    meta_keys_to_update = [MAPPING_SET_ID, LICENSE]
+    meta.update({key: jsondoc[key] for key in meta_keys_to_update if key in jsondoc})
+
+    # Update CURIE_MAP if context_from_jsondoc exists
+    if context_from_jsondoc:
```
matentzn, line 271: can `context_from_jsondoc` theoretically be "null"? I think so, just double checking

src/sssom/parsers.py
```diff
@@ -247,22 +250,35 @@ def parse_sssom_rdf(
...
+        prefixes_from_json = {
+            key: value
+            for key, value in context_from_jsondoc.items()
+            if isinstance(value, str) and not key.startswith("@")
+        }
```
matentzn, line 276: Could the above be shortened by using `curies.Converter.from_json().bimap`?

https://curies.readthedocs.io/en/latest/api/curies.Converter.html#curies.Converter.from_jsonld

tests/data/basic_subset.tsv
```diff
@@ -13,6 +13,7 @@
 #  b: "http://example.org/b/"
 #  c: "http://example.org/c/"
 #  d: "http://example.org/d/"
+#  orcid: "https://orcid.org/"
```
matentzn, line 16: Why did we only now realise this was missing? Weird.

tests/test_parsers.py
```diff
@@ -410,3 +410,26 @@ def test_round_trip(self):
...
+        input_msdf = parse_sssom_table(input_tsv)
+        input_msdf.clean_prefix_map()
+        with TemporaryDirectory() as directory:
+            directory = Path(directory)
+            json_path = directory.joinpath("basic_subset.json")
+            tsv_path = directory.joinpath("json_to_tsv.tsv")
+            with json_path.open("w") as jfile:
+                convert_file(input_path=input_tsv, output_format="json", output=jfile)
+            with tsv_path.open("w") as tfile:
+                parse_file(
```
matentzn, line 426: Why use the wrapper method rather than parse_sssom_json?

tests/test_parsers.py
```diff
@@ -410,3 +410,26 @@ def test_round_trip(self):
...
+                    input_path=str(json_path),
+                    input_format="json",
+                    prefix_map_mode="merged",
+                    output=tfile,
+                )
+            msdf = parse_sssom_table(tsv_path)
+            pd.testing.assert_frame_equal(input_msdf.df, msdf.df)
```
matentzn, line 433: cool. Never ever heard of that before! Nice!

src/sssom/parsers.py
hrshdhgd: I can revert it back if there's some rule against it. I had no idea.

src/sssom/parsers.py
hrshdhgd: I don't know for sure. Just taking precautionary measures.

src/sssom/parsers.py
hrshdhgd: You're absolutely right! Great catch, thanks!

src/sssom/parsers.py
matentzn: I dont really now; You would have to google or ask chatgpt! I would be interested in that, because it comes up in so many projects. Basically: When is it better to use strings as keys vs static variables.

tests/data/basic_subset.tsv
hrshdhgd: Yeah weird. Because this test tsv never went through `clean_prefixes()` until this time.

tests/test_parsers.py
hrshdhgd: Thanks!

tests/test_parsers.py
matentzn: I dont see a big problem with this, but it seems like it is adding a variable into this test (which is "the mechanism by which `parse_file` determines which parsing function to call"), which should ideally be tested in a another specialised test, but I guess you could argue that this test "more closely approximates the behaviour of CLI". So I will leave it at that.

hrshdhgd: Exactly.

src/sssom/parsers.py
```diff
@@ -247,22 +249,27 @@ def parse_sssom_rdf(
...
     with open(file_path) as json_file:
         jsondoc = json.load(json_file)
+
+    # Initialize meta if it's None
+    if meta is None:
+        meta = {}
+
+    # Update metadata with values from JSON document
+    meta_keys_to_update = [MAPPING_SET_ID, LICENSE]
+    meta.update({key: jsondoc[key] for key in meta_keys_to_update if key in jsondoc})
```
matentzn, line 266: Why is this step necessary? There are dozens of `meta` level metadata elements in the sssom file, like descriptions and comments. Shouldnt this line load _all_ the mapping set level metadata? https://github.com/mapping-commons/sssom-py/blob/4c480a5d4bc419ba81f496ee3aa1152ddb0b5589/src/sssom/parsers.py#L491C5-L491C16

hrshdhgd: The `meta` returned without the extra lines in question does not take into account the metadata in the JSON file.

In this line:
```
converter_via_metadata, meta = _get_prefix_map_and_metadata(prefix_map=prefix_map, meta=meta)
```
The `_get_prefix_map_and_metadata()` returns just the default metadata. In spite of taking the `jsondoc` into consideration, the metadata from JSON is not really considered.

Proof: I commented out these lines and ran the test. Gave me this:
```
AssertionError: {'cre[41 chars]ps://creativecommons.org/publicdomain/zero/1.0[160 chars]her'} != {'cre[41 chars]ps://w3id.org/sssom/license/unspecified', 'map[169 chars]her'}
E             {'creator_id': ['orcid:1234'],
E           -  'license': 'https://creativecommons.org/publicdomain/zero/1.0/',
E           +  'license': 'https://w3id.org/sssom/license/unspecified',
E              'mapping_date': '2020-05-30',
E           -  'mapping_set_id': 'https://w3id.org/sssom/mapping/tests/data/basic.tsv',
E           +  'mapping_set_id': 'https://w3id.org/sssom/mappings/cd9c5c7c-fe7a-49bc-9cb6-91d89de30867',
E              'mapping_tool': 'https://github.com/cmungall/rdf_matcher'}
```

hrshdhgd: It's getting lost somewhere. I'll make this change in the correct location and remove it from the current location.

hrshdhgd: Ok dug deeper:
Below the line linked by you above, there is this:
```
_set_metadata_in_mapping_set(mapping_set, metadata=meta)
``` 

https://github.com/mapping-commons/sssom-py/blob/f4256047f75f8d0a286b7255f76ab884d16e049d/src/sssom/parsers.py#L471-L486

This line  causes the issue. As a rule in this function `metadata` trumps `mapping_set` if key exists in both objects.

I'll try to explain this here so please bear with me.

In the absence of a metadata path in the param list while running the CLI for `sssom parse` (JSON=>TSV step) , a default metadata is adopted. 

So when we run:
```
with tsv_path.open("w") as tfile:
                parse_file(
                    input_path=str(json_path),
                    input_format="json",
                    prefix_map_mode="merged",
                    output=tfile,
                )
```
the param `metadata_path` is None (absent). This eventually calls `parse_sssom_json()` which looks like this:

https://github.com/mapping-commons/sssom-py/blob/f4256047f75f8d0a286b7255f76ab884d16e049d/src/sssom/parsers.py#L249-L258


Note `metadata_path` == `meta` which will be `None`. This is part triggers: 

https://github.com/mapping-commons/sssom-py/blob/f4256047f75f8d0a286b7255f76ab884d16e049d/src/sssom/parsers.py#L304-L305

Now the default metadata is: 
```
meta = {'mapping_set_id': 'https://w3id.org/sssom/mappings/d105f78d-a95a-445c-879f-e2ed4dc86aa7', 'license': 'https://w3id.org/sssom/license/unspecified'}
```
and hence the ones in the `mapping_set` (which are correct) get lost.

hrshdhgd: And to answer the original question, this code seems appropriate but instead of just 2 attributes from `MappingSet` I have included all 
```
[attr for attr in dir(MappingSet) if not attr.startswith("_")]
```

matentzn: Wow thanks for the detailed explanation. Can we solve this problem in our next 1:1? There is a lot to unpack here, and not enough time for me digg into everything. But your `meta` analysis seems spot on, and this suggests there is a mistake in the design `meta`. 

Without going into details, the fact that we want `meta` to trump whatever is in the JSON (or TSV, or any) file is in conflict with the fact that providing _no_ meta leads to adding default metadata. Default metadata should only be added _after_ we are sure the sssom file does not have that information in it.

Thanks for this thorough analysis, was a pleasure to read!

hrshdhgd: Sure! Just to add here , I tried rectifying the rule: `metadata` Trump's `mapping_set` by only adding key-value pairs from `metadata` into `mapping_set` but that breaks other tests . So I think there is a deeper problem we need to deal with in this case.

cthoyt:
I am pretty skeptical of this. Can you make a 100% explicit test that shows what is going on here? Loading these opaque files makes it difficult to review this PR

src/sssom/parsers.py
matentzn: I will close this comment as I understand the problem now; will work it out now.

matentzn:
@cthoyt I have now updated the code to reflect your design in the parse_sssom_table method. As we wont have Harshad for a while, and I will have to deal with maintaining SSSOM for a bit, if you can find a minute to see if what I am doing makes sense, that would be much appreciated.

src/sssom/parsers.py
```diff
@@ -235,10 +237,27 @@ def parse_sssom_rdf(
...
+
+    # The priority order for combining prefix maps are:
+    #  1. Built-in prefix map
+    #  2. Internal prefix map inside the document
+    #  3. Prefix map passed through this function inside the ``meta``
+    #  4. Prefix map passed through this function to ``prefix_map`` (handled with ensure_converter)
+    converter = curies.chain(
+        [
+            _get_built_in_prefix_map(),
+            Converter.from_rdflib(g),
```
cthoyt, line 256: can you add a test for this round trip?

tests/test_parsers.py
```diff
@@ -410,3 +414,99 @@ def test_round_trip(self):
...
+            reconsitited_msdf_with_meta.metadata,
+        )
+
+    def test_round_trip_json_tsv(self):
+        """Test JSON => TSV using parse()."""
+        self._basic_round_trip(parse_sssom_json)
+
+    def test_round_trip_rdf_tsv(self):
+        """Test RDF => TSV using parse()."""
+        self._basic_round_trip(parse_sssom_rdf)
```
matentzn, line 512: @cthoyt I added the test - as it is identical with the json one, I decided to not duplicate the code. Do you consider this an anti-pattern?

src/sssom/parsers.py
matentzn: Done!

tests/test_parsers.py
cthoyt: This is great. It means the test can be arbitrarily extended for all formats using the _registry_ pattern. I updated the test in [2081f47](https://github.com/mapping-commons/sssom-py/pull/480/commits/2081f47cfec98e6ca3a3f3acf13382b586a1e812) to do exactly that (though now it identifies more errors to fix ;)

tests/test_parsers.py
```diff
@@ -403,10 +393,57 @@ def test_round_trip(self):
...
+        # a chain of the original metadata with the test metadata.
+        # In particular, this ensures that fields in the test metadata provided are added
+        # to the MappingSet if they are not present, but not updated if they are already present.
+        self.assertEqual(combine_meta, reconstituted_msdf_with_meta.metadata)
+
+    def test_round_trip_json(self):
+        """Test writing then reading JSON."""
+        self._basic_round_trip("json")
+
+    def test_round_trip_rdf(self):
```
cthoyt, line 443: currently failing because no ORCID prefix is preserved during round-trip

cthoyt:
@matentzn current issues are now due to data quality with the test data. I can't address further since it's going to require debugging some mismatch between the RDF and the LinkML schema


```python-traceback
INFO:root:Unmapped predicated: {rdflib.term.URIRef('https://w3id.org/sssom/comment'), rdflib.term.URIRef('https://w3id.org/sssom/creator_id'), rdflib.term.URIRef('https://w3id.org/sssom/license'), rdflib.term.URIRef('https://w3id.org/sssom/mapping_date')}
INFO:root:Triple processed = 2459, unprocessed = 86


Ran 1 test in 6.724s

FAILED (errors=1)

Error
  File "/Users/cthoyt/dev/sssom-py/tests/test_parsers.py", line 250, in test_parse_sssom_rdf
    msdf = from_sssom_rdf(g=self.rdf_graph, prefix_map=self.df_converter, meta=self.metadata)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/cthoyt/dev/sssom-py/src/sssom/parsers.py", line 451, in from_sssom_rdf
    RDFLibLoader().load(
  File "/Users/cthoyt/.virtualenvs/indra/lib/python3.11/site-packages/linkml_runtime/loaders/rdflib_loader.py", line 256, in load
    objs = self.from_rdf_graph(g, schemaview=schemaview, target_class=target_class, prefix_map=prefix_map, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/cthoyt/.virtualenvs/indra/lib/python3.11/site-packages/linkml_runtime/loaders/rdflib_loader.py", line 205, in from_rdf_graph
    return [target_class(**x) for x in root_dicts]
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/cthoyt/.virtualenvs/indra/lib/python3.11/site-packages/linkml_runtime/loaders/rdflib_loader.py", line 205, in <listcomp>
    return [target_class(**x) for x in root_dicts]
            ^^^^^^^^^^^^^^^^^
  File "<string>", line 31, in __init__
  File "/Users/cthoyt/.virtualenvs/indra/lib/python3.11/site-packages/sssom_schema/datamodel/sssom_schema.py", line 110, in __post_init__
    self.MissingRequiredField("license")
  File "/Users/cthoyt/.virtualenvs/indra/lib/python3.11/site-packages/linkml_runtime/utils/yamlutils.py", line 273, in MissingRequiredField
    raise ValueError(f"{field_name} must be supplied")
ValueError: license must be supplied
```

tests/test_parsers.py
matentzn: Fixed

matentzn (approved):

matentzn:
@cthoyt thank you for your help on this; I fixed the remaining test and now everything passes. If you don't see any further issue, I think this is ready to go!

PR #482, opened 2023-12-21
Laziload global MAPPING_SLOTS variable

matentzn:


cmungall (dismissed):
better, but I still don't understand why a global is needed here. Why not just use lru_cache?

matentzn:
I am not a programmer and I dont know the cost of "set()". If this is negligible on the scale of of 1 million operations, I would use lru cache of the schema object. What do you think?

matentzn:
@cmungall I am now simply passing the `mapping_slots` variable into the function and declare the variable in the calling function. globals are gone.

cthoyt (approved):

matentzn:
Thank you @cthoyt!

PR #485, opened 2024-01-13
Bugfix: `clean_prefix_map()` not detecting prefix synonyms

joeflack4:
## Overview
This PR originally had more changes in order to provide fixes for https://github.com/monarch-initiative/mondo-ingest/pull/394. However it is no longer necessary to merge this in order for that PR to be merged.

This PR is now just a bugfix in `clean_prefix_map()` in which the converter object contains prefix synonyms.

---

@twhetzel FYI for your review!
@souzadevinicius A lot going on here, but FYI. This is related to my [mondo ingest pr](https://github.com/monarch-initiative/mondo-ingest/pull/394).

### DIFF
--- a/src/sssom/io.py
+++ b/src/sssom/io.py
@@ -132,19 +132,8 @@ def split_file(input_path: str, output_directory: Union[str, Path]) -> None:
 
 def convert_plain_prefix_map_to_extended(prefix_map):
     """Convert a standard key/val prefix map to extended prefix map format."""
-    by_uri_prefix = {}
-    for prefix, uri_prefix in prefix_map.items():
-        if uri_prefix in by_uri_prefix:
-            by_uri_prefix[uri_prefix]["prefix_synonyms"].append(prefix)
-            continue
-        by_uri_prefix[uri_prefix] = {
-            "prefix": prefix,
-            "prefix_synonyms": [],
-            "uri_prefix": uri_prefix,
-            "uri_prefix_synonyms": [],
-        }
-    epm = list(by_uri_prefix.values())
-    return epm
+    # TODO delete
+    return curies.upgrade_prefix_map(prefix_map)
 
 
 def get_metadata_and_prefix_map(
@@ -165,8 +154,8 @@ def get_metadata_and_prefix_map(
 
     metadata = dict(ChainMap(metadata, get_default_metadata()))
     prefix_map = metadata.pop(CURIE_MAP, {})
-    epm = convert_plain_prefix_map_to_extended(prefix_map)
-    converter = Converter.from_extended_prefix_map(epm)
+    records = curies.upgrade_prefix_map(prefix_map)
+    converter = Converter(records)
     converter = _merge_converter(converter, prefix_map_mode=prefix_map_mode)
     return converter, metadata
 
--- a/src/sssom/io.py
+++ b/src/sssom/io.py
@@ -130,12 +130,6 @@ def split_file(input_path: str, output_directory: Union[str, Path]) -> None:
     write_tables(splitted, output_directory)
 
 
-def convert_plain_prefix_map_to_extended(prefix_map):
-    """Convert a standard key/val prefix map to extended prefix map format."""
-    # TODO delete
-    return curies.upgrade_prefix_map(prefix_map)
-
-
 def get_metadata_and_prefix_map(
     metadata_path: Union[None, str, Path] = None, *, prefix_map_mode: Optional[MergeMode] = None
 ) -> Tuple[Converter, MetadataType]:
@@ -153,9 +147,7 @@ def get_metadata_and_prefix_map(
         metadata = yaml.safe_load(file)
 
     metadata = dict(ChainMap(metadata, get_default_metadata()))
-    prefix_map = metadata.pop(CURIE_MAP, {})
-    records = curies.upgrade_prefix_map(prefix_map)
-    converter = Converter(records)
+    converter = Converter.from_prefix_map(metadata.pop(CURIE_MAP, {}))
     converter = _merge_converter(converter, prefix_map_mode=prefix_map_mode)
     return converter, metadata
 
--- a/src/sssom/io.py
+++ b/src/sssom/io.py
@@ -130,6 +130,23 @@ def split_file(input_path: str, output_directory: Union[str, Path]) -> None:
     write_tables(splitted, output_directory)
 
 
+def convert_plain_prefix_map_to_extended(prefix_map):
+    """Convert a standard key/val prefix map to extended prefix map format."""
+    by_uri_prefix = {}
+    for prefix, uri_prefix in prefix_map.items():
+        if uri_prefix in by_uri_prefix:
+            by_uri_prefix[uri_prefix]["prefix_synonyms"].append(prefix)
+            continue
+        by_uri_prefix[uri_prefix] = {
+            "prefix": prefix,
+            "prefix_synonyms": [],
+            "uri_prefix": uri_prefix,
+            "uri_prefix_synonyms": [],
+        }
+    epm = list(by_uri_prefix.values())
+    return epm
+
+
 def get_metadata_and_prefix_map(
     metadata_path: Union[None, str, Path] = None, *, prefix_map_mode: Optional[MergeMode] = None
 ) -> Tuple[Converter, MetadataType]:
@@ -147,7 +164,9 @@ def get_metadata_and_prefix_map(
         metadata = yaml.safe_load(file)
 
     metadata = dict(ChainMap(metadata, get_default_metadata()))
-    converter = Converter.from_prefix_map(metadata.pop(CURIE_MAP, {}))
+    prefix_map = metadata.pop(CURIE_MAP, {})
+    epm = convert_plain_prefix_map_to_extended(prefix_map)
+    converter = Converter.from_extended_prefix_map(epm)
     converter = _merge_converter(converter, prefix_map_mode=prefix_map_mode)
     return converter, metadata
 
--- a/src/sssom/util.py
+++ b/src/sssom/util.py
@@ -246,7 +246,15 @@ class MappingSetDataFrame:
         if self.metadata:
             prefixes_in_table.update(get_prefixes_used_in_metadata(self.metadata))
 
-        missing_prefixes = prefixes_in_table - self.converter.get_prefixes()
+        prefixes = {record.prefix for record in self.converter.records}
+        aliases = {
+            p
+            for p in itt.chain(*[record.prefix_synonyms for record in self.converter.records])
+            if p
+        }
+        converter_prefixes = prefixes | aliases
+        missing_prefixes = prefixes_in_table - converter_prefixes
+
         if missing_prefixes and strict:
             raise ValueError(
                 f"{missing_prefixes} are used in the SSSOM mapping set but it does not exist in the prefix map"
--- a/src/sssom/util.py
+++ b/src/sssom/util.py
@@ -246,15 +246,7 @@ class MappingSetDataFrame:
         if self.metadata:
             prefixes_in_table.update(get_prefixes_used_in_metadata(self.metadata))
 
-        prefixes = {record.prefix for record in self.converter.records}
-        aliases = {
-            p
-            for p in itt.chain(*[record.prefix_synonyms for record in self.converter.records])
-            if p
-        }
-        converter_prefixes = prefixes | aliases
-        missing_prefixes = prefixes_in_table - converter_prefixes
-
+        missing_prefixes = prefixes_in_table - self.converter.get_prefixes(include_synonyms=True)
         if missing_prefixes and strict:
             raise ValueError(
                 f"{missing_prefixes} are used in the SSSOM mapping set but it does not exist in the prefix map"
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -431,6 +431,29 @@ class TestUtils(unittest.TestCase):
                 self.assertEqual(value, result_with_mapping_object[key])
                 self.assertEqual(value, result_with_dict[key])
 
+    def test_converter_get_prefixes_including_synonyms_and_get_subconverter(self):
+        """Test if get_subconverter() works with prefix synonyms."""
+        EPM = [
+            {
+                "prefix": "SCTID",
+                "prefix_synonyms": ["snomed"],
+                "uri_prefix": "http://snomed.info/id/",
+            },
+        ]
+
+        converter = Converter.from_extended_prefix_map(EPM)
+        self.assertNotIn("snomed", converter.get_prefixes(include_synonyms=False))
+        self.assertIn("snomed", converter.get_prefixes(include_synonyms=True))
+        subconverter = converter.get_subconverter(["snomed"])
+        self.assertIn(
+            "snomed",
+            subconverter.bimap.keys(),
+            "The subconverter which is used to generate "
+            "the prefixmap for the serialisation does not "
+            "contain the 'snomed' prefix, but it is used in "
+            "the raw data",
+        )
+
     def test_curiechain_with_conflicts(self):
         """Test curie map with CURIE/URI clashes."""
         PREFIXMAP_BOTH = {
--- a/poetry.lock
+++ b/poetry.lock
@@ -238,13 +238,13 @@ files = [
 
 [[package]]
 name = "curies"
-version = "0.7.5"
+version = "0.7.6"
 description = "Idiomatic conversion between URIs and compact URIs (CURIEs)."
 optional = false
 python-versions = ">=3.8"
 files = [
-    {file = "curies-0.7.5-py3-none-any.whl", hash = "sha256:56b6a980b0b6a1e4ebe59c42c528ab3ce365d521d8382f37601a4c030c66c28d"},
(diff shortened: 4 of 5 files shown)


src/sssom/io.py
```diff
@@ -130,6 +130,23 @@ def split_file(input_path: str, output_directory: Union[str, Path]) -> None:
     write_tables(splitted, output_directory)
 
 
+def convert_plain_prefix_map_to_extended(prefix_map):
```
joeflack4, line 133: ## `convert_plain_prefix_map_to_extended()`
<details><summary>Details</summary>
<p>

### About
Converts a plain prefix_map to an extended prefix map (EPM).

### Rationale
Why did I make this? I can't say 100% for certain because I'm new to a lot of this code, but from what I can see, there are parts of the code which seem to expect prefix maps in the plain (Dict of PREFIX -> URI_PREFIX) format. And I can't on my own (and don't want to) redesign all of `sssom-py` to use EPMs instead. Therefore I have created this for flexibility, to handle situations where I must deal with plain prefix maps some of the time, but I need functionality of the extended prefix maps (in my case, prefix synonyms).

### Other changes to consider
I wanted to add some typing to this, like one of the following...
```py
def convert_plain_prefix_map_to_extended(prefix_map: Dict[PREFIX, URI]) -> List[Dict]:

def convert_plain_prefix_map_to_extended(prefix_map: PREFIX_MAP) -> EXTENDED_PREFIX_MAP:
```
...but I didn't want to be bogged down with deciding/discussing this, as it's important to merge this soon to fix a bug that's preventing the Mondo pipeline from being run.

</p>
</details>

src/sssom/util.py
```diff
@@ -246,7 +246,11 @@ def clean_prefix_map(self, strict: bool = True) -> None:
         if self.metadata:
             prefixes_in_table.update(get_prefixes_used_in_metadata(self.metadata))
 
-        missing_prefixes = prefixes_in_table - self.converter.get_prefixes()
+        prefixes = {record.prefix for record in self.converter.records}
+        aliases = {p for p in itt.chain(*[record.prefix_synonyms for record in self.converter.records]) if p}
+        converter_prefixes = prefixes | aliases
```
joeflack4, line 251: ## Updates: `clean_prefix_map()`
<details><summary>Details</summary>
<p>

## Overview
I was getting this error thrown near the end of `clean_prefix_map()`:
```
ValueError: {'oio'} are used in the SSSOM mapping set but it does not exist in the prefix map
```
Basically, there were some rows in the mapping set which used `oio`, but `oio` was not detected in the prefix map.

The cause of this bug was that when it checked for prefixes in the prefix map, it was only checking `prefix` on each `Record` and ignoring `prefix_synonyms`.

## Alternative solutions
I was considering that instead of doing this fix in `sssom-py`, a fix could also be implemented in `curies`. The bug I described above is actually exhibited in `curies` in `Converter.get_prefixes()`, so the same fix could be implemented there. Perhaps it could also be parameterized like, `Converter.get_prefixes(include_synonyms=True)`.

</p>
</details>

src/sssom/io.py
```diff
@@ -147,7 +164,9 @@ def get_metadata_and_prefix_map(
         metadata = yaml.safe_load(file)
 
     metadata = dict(ChainMap(metadata, get_default_metadata()))
-    converter = Converter.from_prefix_map(metadata.pop(CURIE_MAP, {}))
+    prefix_map = metadata.pop(CURIE_MAP, {})
+    epm = convert_plain_prefix_map_to_extended(prefix_map)
+    converter = Converter.from_extended_prefix_map(epm)
```
joeflack4, line 169: ## Update: `get_metadata_and_prefix_map()`
<details><summary>Details</summary>
<p>

### Overview
Now instantiates a `Converter` from extended prefix map (EPM) instead of plain prefix map.

To see how this works, look at my comment for `convert_plain_prefix_map_to_extended()`.

It does this in a funky way which I think needs rethinking and refactoring. But that could take a fair amount of discussion. For now, since the Mondo pipeline is completely broken by this, I chose to make this quicker change.

## Other considerations & blockers


</p>
</details>

src/sssom/io.py
cthoyt: I'm not sure these code changes appropriately address the issue where you have a prefix map that isn't bijective. Please add some unit tests that reproduce the issue you're having then we can take a second pass at the implementation

joeflack4: I should use that term, bijective prefix map, if I can figure out exactly one that means. Non-duplicative entries I guess.

I know it addresses my issues because I made the changes locally and they work. But yeah I can add a test.

cthoyt: The issue here is this function is non-deterministic in how it picks which prefix becomes the "canonical" one, based on the order of iteration through the prefix map. Some alternatives:

1. Preferred: enforce the MONDO prefix map is bijective
2. Help me push through changes to the SSSOM spec to allow the `curie_map` field to contain an EPM, then we can more carefully define the prefix map
3. Least preferred: add an additional argument to this `convert_plain_prefix_map_to_extended` function that is the list of preferred prefixes, which you have to specify ahead of time, therefore making this function deterministic

joeflack4: 1. _Bijective enforcement_: An alternative would be something like a `strict` param for bijectiveness, and if `strict=False`, we can do this behavior to solve this problem via inferred prefix synonymity. Else, throw an error.
2. _Spec_: I probably don't have much time to work on the spec. At least for this PR I think @matentzn will probably agree the bugfix should come before official spec change. I would suggest though that rather than (a), `curie_map` allowing for both plain and extended prefix maps, (b) SSSOM should introduce a new field `extended_prefix_map`.
3. _Determinism_: For making this function deterministic, there are other possibilities. Here's another idea, which I prefer: For all cases where the same `uri_prefix` has more than 1 `prefix`, the prefix that comes first alphabetically is inferred as canonical and the others are synonyms. We can also show a warning message when this happens. I think this goes well with my suggestion in (1) for a `strict` param. I prefer this is type of way of handling this because I think we should solve this problem primarily by encouraging extended prefix maps over plain prefix maps whenever possible, but for those who wish to use plain prefix maps, give them the least amount up front work to do; make it an easy user experience.

matentzn: Thank you @joeflack4 for attacking this problem! It is very important you make issues before making PRs, as there are always dozens of unspoken (undocumented) assumptions about the way we designed things - for example in sssom-py. 

Can your problem be solved by making sure that the prefix maps passed to Mondo are bijective? What exactly is the problem in the Mondo pipeline, i.e. why are we even passing non-bijective prefixmaps into the system? Only if we can establish that there is an undeniable use case for non-bijective prefixmaps should we consider interpreting those as EPMs (I am not generally opposed to that, this PR may well have the solution). 

To make an informed call, which may get me to reconsider this PR here, I need to know exactly why the problem occurred in the first place.

If we decide that the use case is compelling, and your solution is the way to go, then we can create a test (an explicit one, see `test_parsers.py` towards the end that illustrates the compelling issue and its solution.

src/sssom/util.py
matentzn: Why are there two seperate prefixes for the same thing your mapping set?

src/sssom/io.py
matentzn: > Can we rename this? Because that's what it actually does: return converter, metadata

I prefer to only rename things if they are either _private_ methods, or there is no other way to not create churn for downstream users.

cthoyt: Added deprecation PR for this in https://github.com/mapping-commons/sssom-py/pull/486

src/sssom/util.py
cthoyt: also, this is not related to the other changes so it would make much more sense to break this up into 2 PRs

src/sssom/io.py
joeflack4, line 133: ## Refactoring from Charlie's `curies` update
I believe that https://github.com/cthoyt/curies/pull/99 will address some/all of the same changes here, so, assuming we end up merging this PR, will probably need to refactor afterwards to utilize the `curies` solution instead.

src/sssom/io.py
joeflack4: That's what I was gonna suggest. Rename it to `_get_converter_and_metadata()`, have the old function reroute to this one, and deprecate the old function.

src/sssom/util.py
joeflack4: @matentzn I'm not sure how `oio` and `oboInOwl` got into `mondo-ingest`'s `mondo.sssom.config.yaml`; I mostly don't use that. They have duplicative URI prefixes.
However even when removing 1 of them from `mondo.sssom.config.yaml`, the problem arises again. Basically the lexmatch pipeline gets and merges 2 prefix maps. One of them comes from `mondo.sssom.config.yaml`. The second is dynamically generated by reading the contents of Mondo.

cthoyt: I understand that they're both motivated by the same work that you're doing, but they are not related code changes, which makes working through this PR not easy nor enjoyable. This will also slow down the speed we're able to work through this

joeflack4: There's not a lot of code changes here. I'm surprised if this is a difficult PR to work through.

@matentzn @cthoyt But yes there are 2 different issues here, so I can split into 2 different PRs if you guys want?
1. PR for dealing with bijective prefix maps (`convert_plain_prefix_map_to_extended()` and `get_metadata_and_prefix_map()`
2. PR for what I think is a bug in `clean_prefix_map()` where it does not detect prefix synonyms.

src/sssom/io.py
cthoyt: Even though I've provided a utility function in `curies` to do this, I still think SSSOM-py should require bijective prefix maps (or if not, require them to be formulated as EPMs)

cthoyt: done in 57dd171

matentzn: I agree with @cthoyt, for different reasons than him, but the outcome is the same - I think non-bijective prefix maps are, in 85% of the cases, unintentional and should be fixed by the provider rather than processed as EPMs. I fear that interpreting prefixmaps as EPMs in sssom-py will lead to unintended, like cases where the user simply made a typo which is now leading to a wrong reconciliation.

joeflack4: Sounds good; I suggested some ways of doing this earlier--passing EPMs instead. The choices:
a. Work w/ the EPMs in `mondo-ingest`; don't need to call `sssom-py` as Harshad initially set up
b. Keep Harhad's original lexmatch scripting pattern, and add a new function to `sssom-py`: `_get_converter_and_metadata_from_epm()`
c. keep Harhad's original lexmatch scripting pattern, and update `_get_converter_and_metadata()` for a new param for EPMs

Since this is now a private function I'm guessing I should try for 'a'.

matentzn: I dont know the concrete thoughts you are having about (a), but I think this should be the right place for the solution.

src/sssom/util.py
```diff
@@ -246,7 +246,7 @@ def clean_prefix_map(self, strict: bool = True) -> None:
         if self.metadata:
             prefixes_in_table.update(get_prefixes_used_in_metadata(self.metadata))
 
-        missing_prefixes = prefixes_in_table - self.converter.get_prefixes()
+        missing_prefixes = prefixes_in_table - self.converter.get_prefixes(include_synonyms=True)
```
matentzn, line 249: Why this change?

cthoyt: This overwrote the changes Joe made with an equivalent upstream improvement to the `curies` package. I still don't think we should add this here, but this is for posterity. Ultimately I hope this PR will get rejected in favor of solving the data problems upstream.

joeflack4: Big thanks @cthoyt for contributing your updates to the PR! I thought I was going to have to do that.

However the changes in this function `clean_prefix_map()` don't seem like they'd involve an upstream data fix. This is just a bugfix. Sorry, as mentioned there are 2 things in this PR and maybe it would've been best to fix: (1) for handling bijective maps and converting plain prefix maps to EPMs, and (2) fixing `clean_prefix_map()` so that it detects prefix synonyms. Otherwise it throws an error because it thinks prefixes are missing.

cthoyt: @joeflack4 I'll revert this change and leave it for you to open a new PR that includes the update and a unit test that demonstrates the bug

joeflack4: Hmm, I think one or both of us is confused. You don't need to revert anything. Your change to move this logic from `sssom-py`'s `clean_prefix_map()` to `curies` is good. I wanted this logic in `curies` originally, and I think the `get_prefixes(include_synonyms=True)` solution is good.

I could make a test but it's probably redundant if you've already tested the `include_synonyms` param in `curies`.

I think this one line update adding `include_synonyms=True` is probably the only change needed in `sssom-py` now to address my issues. I'm likely going to close this PR and open a new one with just those changes (as well as the pyproject.tom / poetry.lock updates).

src/sssom/io.py
joeflack4: This functionality has now been removed. It is not necessary to solve any problems in https://github.com/monarch-initiative/mondo-ingest/pull/394. It can be re-handled in some variation in future PR for EPM support.

src/sssom/io.py
joeflack4: Charlie has addressed this.

Other functionality I added has now been removed. It is not necessary to solve any problems in https://github.com/monarch-initiative/mondo-ingest/pull/394. It can be re-handled in some variation in future PR for EPM support.``

src/sssom/util.py
joeflack4: This fix has been superseded with `curies` solution which does the same things. See [this comment](https://github.com/mapping-commons/sssom-py/pull/485#discussion_r1455360522).

src/sssom/util.py
joeflack4: This PR is essentially a 1-liner now along with updated `pyproject.toml` and `poetry.lock` for `curies` upgrade.

joeflack4: This PR is essentially a 1-liner now along with updated `pyproject.toml` and `poetry.lock` for `curies` upgrade.

matentzn, line 249: Maybe i am wrong. I added a test below that illustrates the issue. 

Lets say the raw data (prefixes in table) contains `snomed`. The converter has `SCTID` as a prefix. `snomed` as a synonym. Now this line of code here will not include `snomed` in missing prefixes, and move on. 

Now you go a few lines below to extract a subconverter over all the prefixes in the table, including `snomed`, which will now become the main converter (this part is not very relevant for the problem but bear with me). Now, when the prefixmap gets serialsed to a valid SSSOM file, which uses `msdf.converter.bimap` (see write_table() method). This map does _not_ contain `snomed`. so it will write SCTID, while the data (the mappings) still contains `snomed`. 

If you still think this PR is right, you need to convince me that giving a mapping set with a value containing `snomed`, and `snomed` being a synonym in the EPM, the serialised SSSOM files contains `snomed` prefix..

tests/test_utils.py
```diff
@@ -431,6 +431,29 @@ def test_get_dict_from_mapping(self):
...
+            {
+                "prefix": "SCTID",
+                "prefix_synonyms": ["snomed"],
+                "uri_prefix": "http://snomed.info/id/",
+            },
+        ]
+
+        converter = Converter.from_extended_prefix_map(EPM)
+        self.assertNotIn("snomed", converter.get_prefixes(include_synonyms=False))
+        self.assertIn("snomed", converter.get_prefixes(include_synonyms=True))
```
cthoyt, line 446: it's not SSSOM-py's job to test `curies` functionality. The test should show how the change in this PR actually effects input and output to SSSOM-py functions

matentzn: This is not a test, it's some code to show why the pr the way it is should not be merged.. I don't intend to add this as a test here

src/sssom/util.py
joeflack4: Mm, I see the problem. I think you touched on this at the last meeting. How do you feel about serializing a non-bijective prefix map which would include the synonyms, then?

When it comes to working with synonyms in SSSOM, do you think this we've covered all of the cases? Or are there likely to be more places where synonym coverage is needed?

matentzn:
@joeflack4 - I think this PR should be closed, unless you think there is something we should salvage.

joeflack4:
@matentzn This PR is now a 1 line change:
`missing_prefixes = prefixes_in_table - self.converter.get_prefixes(include_synonyms=True)`

And it would seem it is a bugfix. It seems like this is something that should be merged.

I suppose I can pick up where Charlie left off and make sure that the test passes. I haven't looked at it yet.

matentzn:
I thought I had shown that there is a problem with this approach -  check my comments and the "test" that illustrates the issue. Maybe my assessment is wrong, but then you have to proof that my test is wrong 😁

joeflack4:
@matentzn Ah I'm sorry you're right, I see your last discussion.

joeflack4:
@matentzn I closed it but we have something unresolved that I mentioned [the last time](https://github.com/mapping-commons/sssom-py/pull/485#discussion_r1460684961):
> I don't see an issue on the topic of "EPM" / "extended prefix map" or "synonyms". Should we close this for now and move the discussion to a new issue?

PR #486, opened 2024-01-15
Deprecate `get_metadata_and_prefix_map`

cthoyt:
This function used to be used in OAK's lexical matcher pipeline, but has since been removed. Now, the only place it's (potentially mis)used is in the MONDO ingest pipeline (see https://github.com/search?q=get_metadata_and_prefix_map+-is%3Afork+-user%3Amapping-commons&type=code)

This PR deprecates it, and we can remove it in a future version.

### DIFF
--- a/src/sssom/io.py
+++ b/src/sssom/io.py
@@ -12,6 +12,7 @@ import curies
 import pandas as pd
 import yaml
 from curies import Converter
+from deprecation import deprecated
 
 from sssom.validators import validate
 
--- a/src/sssom/io.py
+++ b/src/sssom/io.py
@@ -139,9 +139,7 @@ def get_metadata_and_prefix_map(
     metadata_path: Union[None, str, Path] = None, *, prefix_map_mode: Optional[MergeMode] = None
 ) -> Tuple[Converter, MetadataType]:
     """Load metadata and a prefix map in a deprecated way."""
-    return _get_metadata_and_prefix_map(
-        metadata_path=metadata_path, prefix_map_mode=prefix_map_mode
-    )
+    return _get_metadata_and_prefix_map(metadata_path=metadata_path,prefix_map_mode=prefix_map_mode)
 
 
 def _get_metadata_and_prefix_map(
--- a/src/sssom/io.py
+++ b/src/sssom/io.py
@@ -139,7 +139,9 @@ def get_metadata_and_prefix_map(
     metadata_path: Union[None, str, Path] = None, *, prefix_map_mode: Optional[MergeMode] = None
 ) -> Tuple[Converter, MetadataType]:
     """Load metadata and a prefix map in a deprecated way."""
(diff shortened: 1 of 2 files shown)


matentzn (commented):
I think we need to discuss things like this on issues before we enact them, but lets leave that for the future;

Here my question is: if we make this method private, do we still have a utility method to load a externally defined sssom metadata file? Like 

https://github.com/mapping-commons/sssom/blob/master/examples/external/example1.sssom.yml

In the case of 
https://mapping-commons.github.io/sssom/spec/#external-mode

In other words, I am missing a bit the motivation behind the deprecation.

src/sssom/io.py
```diff
@@ -130,8 +130,22 @@ def split_file(input_path: str, output_directory: Union[str, Path]) -> None:
...
 def get_metadata_and_prefix_map(
     metadata_path: Union[None, str, Path] = None, *, prefix_map_mode: Optional[MergeMode] = None
+) -> Tuple[Converter, MetadataType]:
+    """Load metadata and a prefix map in a deprecated way."""
+    return _get_metadata_and_prefix_map(
+        metadata_path=metadata_path, prefix_map_mode=prefix_map_mode
+    )
+
+
+def _get_metadata_and_prefix_map(
```
joeflack4, line 147: I don't know if it's too much to handle in this PR as well, but per this [discussion thread](https://github.com/mapping-commons/sssom-py/pull/485#discussion_r1452770502), is this also a good time to rename this function to `_get_converter_and_metadata()`?

cthoyt:
> Here my question is: if we make this method private, do we still have a utility method to load a externally defined sssom metadata file?

No, we don't. Since external metadata is a YAML file, I don't think we even need one, since you can do one of the following:

1. Use the SSSOM-py parsers that handle these files implicitly
2. import a standard YAML processor if you want to work with it directly

The issue with having a built in function is it has super confusing semantics about how it combines metadata already inside the YAML file with some other metadata given _ad hoc_. I don't think we should have a function that does this and ask downstream users to implement their own merging, since it will always have different needs.

joeflack4:
@matentzn Haven't put much thought in it yet, but I think it's possible that @cthoyt may be right. I can probably refactor `mondo-ingest`  to do what it needs to do without relying on `sssom-py` for this.

matentzn:
>  I don't think we should have a function that does this and ask downstream users to implement their own merging, since it will always have different needs.

Fair enough. We can provide a wrapper of sssom parse later when a user requests it.

src/sssom/io.py
matentzn: Yes I agree @joeflack4. Now or never! Can we rename?

cthoyt: Done in be8d768

matentzn (approved):
Thank you @cthoyt for your infinite energy!

cthoyt:
@matentzn I forgot an import, can you please approve this again

PR #487, opened 2024-01-15
Winter cleaning

cthoyt:
1. Remove deprecated functions past their removal version
2. Fix some type annotations
3. Delete irrelevant/old comments

### DIFF
--- a/src/sssom/util.py
+++ b/src/sssom/util.py
@@ -1212,9 +1212,6 @@ def is_multivalued_slot(slot: str) -> bool:
     :param slot: Slot name
     :return: Slot is multivalued or no
     """
-    # Ideally:
-    # view = SchemaView('schema/sssom.yaml')
-    # return view.get_slot(slot).multivalued
     return slot in _get_sssom_schema_object().multivalued_slots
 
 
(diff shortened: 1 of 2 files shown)


src/sssom/util.py
```diff
@@ -515,11 +514,13 @@ def get_row_based_on_hierarchy(df: pd.DataFrame):
 
     :param df: Dataframe containing multiple predicates for same subject and object.
     :return: Dataframe with a single row which ranks higher in the hierarchy.
+    :raises KeyError: if no rows are available
     """
     for pred in PREDICATE_LIST:
         hierarchical_df = df[df[PREDICATE_ID] == pred]
         if not hierarchical_df.empty:
             return hierarchical_df
+    raise KeyError
```
matentzn, line 523: What is the motivation behind this change?

cthoyt: The code that uses this function always assumes it returns something. before, there was an implicit return of None at the end. If for whatever reason none of the predicates in the predicate list are available, this code should explicitly break, rather than ending up with a type error when trying to concat a None onto the dataframe that gets returned. This is why strict typing is helpful!

matentzn: Ok, thank you!

matentzn: I don't really understand what this method is doing. I guess your change is independent of that, and makes sense

matentzn (approved):

PR #492, opened 2024-02-02
Updated using `importlib_resource` instead of `pkg_resources` and prepare for later versions of `pandas`

### LINKED ISSUE #491: Get rid of warning "Downcasting behavior in `replace` is deprecated and will be removed in a future version."
```
/usr/local/lib/python3.10/dist-packages/sssom/util.py:156: FutureWarning: Downcasting behavior in `replace` is deprecated and will be removed in a future version. To retain the old behavior, explicitly call `result.infer_objects(copy=False)`. To opt-in to the future behavior, set `pd.se
t_option('future.no_silent_downcasting', True)`
```

clutters all log files massively now; Would be good to adress this correctly (not just suppressing the warning)

hrshdhgd:
Fixes #491  

- [x] Updated using `importlib_resource` instead of `pkg_resources`
   - Reason: `pkg_resources` is going to be deprecated.

- [x] Refactor `pandas` related code to smoothly transition to future versions and handle deprecation warnings.
  - [Pandas PR](https://github.com/pandas-dev/pandas/pull/54710/files#diff-55001624a0932c1b6cee2e6ddb65dea85c1faf0dee84812c0ca0c32916a71438):
  ```
                             "Downcasting behavior in `replace` is deprecated and "
                             "will be removed in a future version. To retain the old "
                             "behavior, explicitly call "
                             "`result.infer_objects(copy=False)`. "
                             "To opt-in to the future "
                             "behavior, set "
                             "`pd.set_option('future.no_silent_downcasting', True)`",
  ```
  - `A value is trying to be set on a copy of a slice from a DataFrame`
  - `.apply(max)` => `.apply(np.maximum.reduce)`
  - `UserWarning: Boolean Series key will be reindexed to match DataFrame index.`


hrshdhgd:
@matentzn : The only warning that shows now is 
```
python3.10/site-packages/curies/api.py:261
  /opt/anaconda3/envs/sssom-py/lib/python3.10/site-packages/curies/api.py:261: PydanticDeprecatedSince20: Pydantic V1 style `@validator` validators are deprecated. You should migrate to Pydantic V2 style `@field_validator` validators, see the migration guide for more details. Deprecated in Pydantic V2.0 to be removed in V3.0. See Pydantic V2 Migration Guide at https://errors.pydantic.dev/2.5/migration/
    @validator("uri_prefix_synonyms")  # type:ignore
```

which is a `curies` related.

src/sssom/constants.py
```diff
@@ -6,14 +6,14 @@
 from functools import cached_property, lru_cache
 from typing import Any, Dict, List, Literal, Set
 
-import pkg_resources
+import importlib_resources
```
matentzn, line 9: This is the context right: https://setuptools.pypa.io/en/latest/pkg_resources.html

image

src/sssom/constants.py
```diff
@@ -6,14 +6,14 @@
...
 import yaml
 from linkml_runtime.utils.schema_as_dict import schema_as_dict
 from linkml_runtime.utils.schemaview import SchemaView
 
 HERE = pathlib.Path(__file__).parent.resolve()
 
-SCHEMA_YAML = pkg_resources.resource_filename("sssom_schema", "schema/sssom_schema.yaml")
+SCHEMA_YAML = importlib_resources.files("sssom_schema").joinpath("schema/sssom_schema.yaml")
```
matentzn, line 16: Lets hope this works also after the wheel is built - but why not :D I always am scared of these low level tools.

src/sssom/util.py
```diff
@@ -80,6 +80,12 @@
 KEY_FEATURES = [SUBJECT_ID, PREDICATE_ID, OBJECT_ID, PREDICATE_MODIFIER]
 TRIPLES_IDS = [SUBJECT_ID, PREDICATE_ID, OBJECT_ID]
 
+# ! This will be unnecessary when pandas >= 3.0.0 is released
```
matentzn, line 83: Can you link an issue here that explains why it will be unnecessary? Just out of curiosity

src/sssom/util.py
```diff
@@ -151,6 +157,12 @@ def from_mapping_set_document(cls, doc: MappingSetDocument) -> "MappingSetDataFr
         df = pd.DataFrame(get_dict_from_mapping(mapping) for mapping in doc.mapping_set.mappings)
         meta = _extract_global_metadata(doc)
 
+        if pandas_version >= (2, 0, 0):
```
matentzn, line 160: Is there still push-back in the community (linkml etc) to simply depend on >=2.0.0 instead of doing a hack like this?

src/sssom/util.py
```diff
@@ -160,6 +172,14 @@ def from_mapping_set_document(cls, doc: MappingSetDocument) -> "MappingSetDataFr
             slot for slot, slot_metadata in slots.items() if slot_metadata["range"] == "double"
         }
         non_double_cols = df.loc[:, ~df.columns.isin(slots_with_double_as_range)]
+
+        if pandas_version >= (2, 0, 0):
+            # For pandas >= 2.0.0, use the 'copy' parameter
+            non_double_cols = non_double_cols.infer_objects(copy=False)
+        else:
+            # For pandas < 2.0.0, call 'infer_objects()' without any parameters
+            non_double_cols = non_double_cols.infer_objects()
```
matentzn, line 181: Same q as above: why not simply depend on pandas 2.0?

src/sssom/util.py
```diff
@@ -423,7 +443,7 @@ def filter_redundant_rows(df: pd.DataFrame, ignore_predicate: bool = False) -> p
     else:
         key = [SUBJECT_ID, OBJECT_ID, PREDICATE_ID]
     dfmax: pd.DataFrame
-    dfmax = df.groupby(key, as_index=False)[CONFIDENCE].apply(max).drop_duplicates()
+    dfmax = df.groupby(key, as_index=False)[CONFIDENCE].apply(np.maximum.reduce).drop_duplicates()
```
matentzn, line 446: Why this change? Is `max` (I assume a built-in like `None`, since it is not imported anywhere?) deprecated?

src/sssom/util.py
```diff
@@ -1397,18 +1417,26 @@ def invert_mappings(
...
         )
+
+        # Filter rows where 'SUBJECT_ID' does not start with the prefix but 'OBJECT_ID' does
         non_prefix_subjects_df = pd.DataFrame(
             non_predicate_modified_df[
-                (~subject_starts_with_prefix_condition & object_starts_with_prefix_condition)
+                (
+                    ~non_predicate_modified_df[SUBJECT_ID].str.startswith(subject_prefix + ":")
+                    & non_predicate_modified_df[OBJECT_ID].str.startswith(subject_prefix + ":")
+                )
```
matentzn, line 1436: Why these changes above? Are they related to warning?

matentzn (commented):
Wellllcome back! My god I missed you man! Hope you are well!

src/sssom/constants.py
hrshdhgd: Yes!

hrshdhgd:
> Wellllcome back! My god I missed you man! Hope you are well!

Haha! All well here sir!

src/sssom/constants.py
hrshdhgd: PyPA recommended this switch: https://setuptools.pypa.io/en/latest/pkg_resources.html

src/sssom/util.py
hrshdhgd: Not sure about the pushback. Since we had tests specific to older pandas versions in our workflow, I added this.

src/sssom/util.py
hrshdhgd: https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.infer_objects.html#

src/sssom/util.py
matentzn: hmmmm I would kinda like to consider to drop pandas 1.X support, especially since 3.0 is one the way. Do you know anyone who has an opinion about this?

src/sssom/util.py
hrshdhgd: Another warning from `pandas` I think

src/sssom/util.py
hrshdhgd: Yes: `UserWarning: Boolean Series key will be reindexed to match DataFrame index.`

src/sssom/util.py
hrshdhgd: Not sure who raised this issue in the first place. For some reason I feel it was either Charlie or Chris. Do not recollect it at all.

src/sssom/constants.py
matentzn: Thx

src/sssom/util.py
matentzn: Isn't oak already on pandas 2? Thats what Chris said on slack..

src/sssom/util.py
matentzn: Can you link the docs you used to inform the switch to np.reduce? I have not been able to find it googling..

src/sssom/util.py
matentzn: Cool if you could add that link to the comment in the script, that would be great

src/sssom/util.py
hrshdhgd: Oak was never the issue. I think it's #432

src/sssom/util.py
hrshdhgd: Weird ... I rolled back to `max` and the deprecation does not show up. Well, this change has been undone.

src/sssom/util.py
hrshdhgd: Done.

hrshdhgd:
Should I merge this?

src/sssom/util.py
matentzn: I am ok with merging this but next time we see ourselves introducing code to handle pandas 1.x I think we should refactor and get rid of the conditionals

hrshdhgd: I agree!

PR #494, opened 2024-02-04
Implementing separate methods for JSON and JSONLD

matentzn:
This PR adds methods

- parse_sssom_jsonld
- from_sssom_jsonld
- write_jsonld
- to_jsonld
- test_parse_sssom_jsonld
- test_write_sssom_jsonld

Which are exactly analogous to what was there before for JSON.

But its actual purpose is not so much to add those methods, but to **carefully review the format** (to make sure we are happy) so we can start making headway on https://github.com/mapping-commons/sssom/issues/321.

## Breaking changes

- `json` parameter now refers to `json`, but used to refer to `jsonld`. So anyone expecting `jsonld` will now be served with `json`.

## JSON Format

We need to make sure that the JSON format looks exactly as we envision it. Problems I see so far

- The biggest shortcoming of the JSON format at the moment is that json format does not have a `curie_map`. We will probably have to https://github.com/mapping-commons/sssom/issues/225

<details><summary>Here is an example JSON file</summary>

```
{
  "mapping_set_id": "https://w3id.org/sssom/mapping/tests/data/basic.tsv",
  "license": "https://creativecommons.org/publicdomain/zero/1.0/",
  "mappings": [
    {
      "subject_id": "a:something",
      "predicate_id": "rdfs:subClassOf",
      "object_id": "b:something",
      "mapping_justification": "semapv:LexicalMatching",
      "subject_label": "XXXXX",
      "subject_category": "biolink:AnatomicalEntity",
      "object_label": "xxxxxx",
      "object_category": "biolink:AnatomicalEntity",
      "subject_source": "a:example",
      "object_source": "b:example",
      "mapping_tool": "rdf_matcher",
      "confidence": 0.8,
      "subject_match_field": [
        "rdfs:label"
      ],
      "object_match_field": [
        "rdfs:label"
      ],
      "match_string": [
        "xxxxx"
      ],
      "comment": "mock data"
    },
    {
      "subject_id": "a:something",
      "predicate_id": "owl:equivalentClass",
      "object_id": "c:something",
      "mapping_justification": "semapv:LexicalMatching",
      "subject_label": "XYXYX",
      "subject_category": "biolink:AnatomicalEntity",
      "object_label": "xyxyxy",
      "object_category": "biolink:AnatomicalEntity",
      "subject_source": "a:example",
      "object_source": "c:example",
      "mapping_tool": "rdf_matcher",
      "confidence": 0.83,
      "subject_match_field": [
        "rdfs:label"
      ],
      "object_match_field": [
        "rdfs:label"
      ],
      "match_string": [
        "xxxxx"
      ],
      "comment": "mock data"
    }
  ],
  "creator_id": [
    "orcid:1234",
    "orcid:5678"
  ],
  "mapping_tool": "https://github.com/cmungall/rdf_matcher",
  "mapping_date": "2020-05-30"
}
```

</details>

### DIFF
--- a/tests/test_writers.py
+++ b/tests/test_writers.py
@@ -17,12 +17,13 @@ from sssom.constants import (
     SUBJECT_ID,
     SUBJECT_LABEL,
 )
-from sssom.parsers import parse_sssom_json, parse_sssom_rdf, parse_sssom_table
+from sssom.parsers import parse_sssom_json, parse_sssom_jsonld, parse_sssom_rdf, parse_sssom_table
 from sssom.writers import (
     _update_sssom_context_with_prefixmap,
     to_json,
     write_fhir_json,
     write_json,
+    write_jsonld,
     write_ontoportal_json,
     write_owl,
     write_rdf,
@@ -76,6 +77,18 @@ class TestWrite(unittest.TestCase):
         with open(path, "w") as file:
             write_json(self.msdf, file)
         msdf = parse_sssom_json(path)
+        self.assertEqual(
+            len(msdf.df),
+            self.mapping_count,
+            f"{path} has the wrong number of mappings.",
+        )
+
+    def test_write_sssom_jsonld(self):
+        """Test writing as JSON."""
+        path = os.path.join(test_out_dir, "test_write_sssom_json.json")
+        with open(path, "w") as file:
+            write_jsonld(self.msdf, file)
+        msdf = parse_sssom_jsonld(path)
         self.assertEqual(
             len(msdf.df),
             self.mapping_count,
--- a/poetry.lock
+++ b/poetry.lock
@@ -63,19 +63,22 @@ dev = ["freezegun (>=1.0,<2.0)", "pytest (>=6.0)", "pytest-cov"]
 
 [[package]]
 name = "beautifulsoup4"
-version = "4.12.2"
+version = "4.12.3"
 description = "Screen-scraping library"
 optional = false
 python-versions = ">=3.6.0"
 files = [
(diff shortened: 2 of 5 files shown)


tests/test_writers.py
```diff
@@ -82,6 +83,18 @@ def test_write_sssom_json(self):
...
+        path = os.path.join(test_out_dir, "test_write_sssom_json.json")
+        with open(path, "w") as file:
+            write_jsonld(self.msdf, file)
+        msdf = parse_sssom_jsonld(path)
+        self.assertEqual(
+            len(msdf.df),
+            self.mapping_count,
+            f"{path} has the wrong number of mappings.",
+        )
+
```
matentzn, line 97: Little @cthoyt in Nicos head:

> **AGAIN???**  Please add short explicit tests so I can understand what is going, in particular the difference between JSON and JSONLD serialisations.

matentzn: @matentzn: I will add tests after we have had some discussions in the nature of the JSON output.

gouttegd:
> We will probably have to https://github.com/mapping-commons/sssom/issues/225

The problem we might run into with that is that, as far as I know (and as I have noted in the discussion about the extension slots), LinkML does not have a `map` type. We’d want to declare a field that could be used like this:

```json
"curie_map": {
  "FBbt": "http://purl.obolibrary.org/obo/FBbt_"
}
```

but unless I missed something in LinkML’s docs, this is not possible. All we can do is to have a _list_ (i.e. a “multi-valued” field) of custom “dictionary entry“ types, like this:

```json
"curie_map":
  - "key": "Fbbt"
    "value": "http://purl.obolibrary.org/obo/FBbt_"
```

which of course would work but would be… weird, at the very least.

My own solution (that nobody will like, I know) to that is simple: decide that CURIEfied identifiers are only for the TSV format (which is what the spec currently says, incidentally), JSON should only contain full-length identifier. No CURIE map needed, problem solved.

PR #503, opened 2024-03-12
Fix CLI for invert

matentzn:


hrshdhgd (approved):
No major change in code here. Just a variable added.

PR #506, opened 2024-03-22
Add flip step for the built-in sssom:superClassOf predicate during OWL conversion

matentzn:
Depends on

- [ ] #505 

This PR adds a preprocessing step to the to_owl_graph() method which flips "sssom:superClassOf" edges. I should really document sssom:superClassOf as syntactic sugar, but right now I dont have the bandwidth. I am certain it will come up.


hrshdhgd:
I merged #505 to main and pulled down to this branch. I think `tox` should pass this one as well

hrshdhgd (dismissed):
If all tests pass, we're good to go.

PR #510, opened 2024-04-01
Update sssom-schema

hrshdhgd:
ran `poetry add sssom-schema@latest`


matentzn (dismissed):

matentzn:
Some stuff broke?

hrshdhgd:
Yes the test is failing because the [rdf test file](https://github.com/mapping-commons/sssom-py/blob/master/tests/data/basic.sssom.rdf) does not have subject_id. Did we use some other file to convert to rdf. I can recreate it?

matentzn:
The RDF file does have a subject_id, but it is _mapped_ to some other relation. Check Slack for details.

hrshdhgd:
This also passes now

matentzn (approved):
Thanks!!

PR #533, opened 2024-05-20
`TestWrite.test_missing_entries()`

joeflack4:
## Changes
- Add: TestWrite.test_missing_entries()


tests/test_writers.py
```diff
@@ -191,3 +192,69 @@ def test_write_sssom_ontoportal_json(self):
             self.mapping_count,
             f"{path} has the wrong number of mappings.",
         )
+
+    # TODO: Determine assertions that should pass
+    # TODO: Fix failing assertions that should pass
+    # TODO: before merging, remove pass/fail comments at end of assertion lines
+    def test_missing_entries(self):
```
joeflack4, line 199: ## Get assertions to pass
- [ ] 1. Determine assertions that should pass @matentzn 
- [ ] 2. Fix failing assertions that should pass
- [ ] 3. before merging, remove pass/fail comments at end of assertion lines

joeflack4, line 199: ## Which assertions are appropriate?
- [ ] @matentzn review

Can you look at the assertions I've made and determine which ones you think should be expected to pass, and which should not?

I've left temporary comments next to each assertion showing their current pass/fail status.

tests/test_writers.py
```diff
@@ -191,3 +192,69 @@ def test_write_sssom_ontoportal_json(self):
...
+                "rdfs": "http://www.w3.org/2000/01/rdf-schema#",
+                "semapv": "https://w3id.org/semapv/",
+                "skos": "http://www.w3.org/2004/02/skos/core#",
+                "sssom": "https://w3id.org/sssom/",
+            },
+            "license": "http://w3id.org/sssom/license/unspecified",
+            "mapping_provider": "https://www.orpha.net/",
+        }
+
+        # When passing Converter
```
joeflack4, line 232: ## FYI: 2 different setups
I have a set of assertions for 2 cases of `MappingSetDataFrame` instantiation:
- with inclusion of a `Converter`
- with no inclusion of a `Converter`

tests/test_writers.py
```diff
@@ -191,3 +192,69 @@ def test_write_sssom_ontoportal_json(self):
...
+        msdf: MappingSetDataFrame = MappingSetDataFrame(
+            converter=curies.Converter.from_prefix_map(metadata["curie_map"]),
+            df=df,
+            metadata=metadata,
+        )
+        self.assertIn("curie_map", msdf.metadata)  # pass
+        self.assertIn("icd11.foundation", msdf.metadata["curie_map"])  # pass
+        self.assertIn("icd11.foundation", msdf.prefix_map)  # pass
+        tmp_file = os.path.join(test_out_dir, "test_missing_entries.sssom.tsv")
```
matentzn, line 241: Is this file already there?

tests/test_writers.py
```diff
@@ -191,3 +192,69 @@ def test_write_sssom_ontoportal_json(self):
...
+        with open(tmp_file, "w") as f:
+            write_table(msdf, f)
+        msdf2 = parse_sssom_table(tmp_file)
+        self.assertIn("curie_map", msdf2.metadata)  # fail
```
matentzn, line 245: This behaviour is expected for now, because `curie_map` sort if lives outside the metadata. however, it will soon be solved: https://github.com/mapping-commons/sssom/pull/349

matentzn (commented):
This is a very good test! Thank you!

tests/test_writers.py
joeflack4: I think I named this output file `test_missing_entries` because that is the name of this test method. I think that's the convention that was used for other output files for other methods, which makes sense.

I did just notice that this `tmp_file` declaration appeared in 2 different places in the method. I just reduced it to 1 and pushed an amended commit.

There is no other test that references this file.

Maybe you're asking, "is this supposed to be a static file that gets read into the test?", and if so, the answer is no. This `tmp_file` gets written in the test method based on the metadata and single mapping set row that I declare in the Python code. Then, after writing, it gets read and additional assertions are run.

matentzn: Sorry, that was I was asking and of course I see it now! Thank you!

tests/test_writers.py
joeflack4: Wow, I'm surprised it wasn't part of the data model.

From an implementation standpoint, I was thinking that maybe `msdf.prefix_map` and `msdf.metadata['curie_map']` could point to the same Python dictionary object. Right now they are different objects and, as these assertions show, can get out-of-sync.

PR #541, opened 2024-06-15
Update validation data files

matentzn:
I remove all the current validation files, and replace them by new ones. This is a first step in building a better test suite to monitor subtle serialisation changes across SSSOM versions.

The updated files are generated by running the `update_validate_data.sh` script.


ehartley (approved):
Looks good!

PR #542, opened 2024-06-18
Add additional typecheck to curie validation

matentzn:
Due to the lack of type safety, i can pass a non string to the `_is_curie` method, which causes the method to fail hard when its trying to do regex matching.

Here, i add an additional safety net around the matching processing, making sure the value is not None, and is indeed a string (and not a number or some other kind of non-string).


ehartley (approved):
Looks good! Thanks for adding this fix!

PR #544, opened 2024-06-20
Fix numpy.nan for numpy 2.0 compatibility

### LINKED ISSUE #543: Numpy 2.0 compatibility
Numpy has just released 2.0 and removed `np.NAN` in favor of `np.nan`. This line with a type annotation is therefore erroring in all downstream dependencies where the numpy version is not explicitly limited:

https://github.com/mapping-commons/sssom-py/blob/master/src/sssom/util.py#L658

I assume we can just change all relevant instances to `np.nan` and make a new sssom-py release to avoid this.

bgyori:
Fixes #543.


matentzn (approved):

matentzn:
Thank you @bgyori !

PR #548, opened 2024-06-25
Addressed situation when assign_default_confidence() returns only dataframe with all NaN confidence values

### LINKED ISSUE #546: Make `filter_redundant_rows` resistant to the case that there is an empty `confidence` column
## Overview
https://github.com/mapping-commons/sssom-py/blob/550206721911f711ee678eb1a8da50591649bd04/src/sssom/util.py#L429

We had the problem that this was failing:

https://github.com/mapping-commons/sssom-py/blob/550206721911f711ee678eb1a8da50591649bd04/src/sssom/util.py#L449

with `AttributeError: 'Series' object has no attribute 'iterrows'`

<details><summary>Log / traceback</summary>
<p>

```
python ../scripts/[lexmatch-sssom-compare.py](http://lexmatch-sssom-compare.py/) extract_unmapped_matches doid gard icd10cm icd10who icd11foundation ncit omim ordo \
	--matches ../mappings/mondo-sources-all-lexical.sssom.tsv \
	--output-dir lexmatch \
	--summary lexmatch/[README.md](http://readme.md/) \
	 --exclusion reports/doid_term_exclusions.txt  --exclusion reports/gard_term_exclusions.txt  --exclusion reports/icd10cm_term_exclusions.txt  --exclusion reports/icd10who_term_exclusions.txt  --exclusion reports/icd11foundation_term_exclusions.txt  --exclusion reports/ncit_term_exclusions.txt  --exclusion reports/omim_term_exclusions.txt  --exclusion reports/ordo_term_exclusions.txt
/usr/local/lib/python3.10/dist-packages/sssom/[parsers.py](http://parsers.py/):428: ChainedAssignmentError: A value is trying to be set on a copy of a DataFrame or Series through chained assignment using an inplace method.
When using the Copy-on-Write mode, such inplace method never works to update the original DataFrame or Series, because the intermediate object on which we are setting values always behaves as a copy.

For example, when doing 'df[col].method(value, inplace=True)', try using 'df.method({col: value}, inplace=True)' instead, to perform the operation inplace on the original object.


  df2[CONFIDENCE].replace(r"^\s*$", np.NaN, regex=True, inplace=True)
/usr/local/lib/python3.10/dist-packages/sssom/[util.py](http://util.py/):168: FutureWarning: Downcasting behavior in `replace` is deprecated and will be removed in a future version. To retain the old behavior, explicitly call `result.infer_objects(copy=False)`. To opt-in to the future behavior, set `pd.set_option('[future.no](http://future.no/)_silent_downcasting', True)`
  df.replace("", np.nan, inplace=True)
/usr/local/lib/python3.10/dist-packages/sssom/[util.py](http://util.py/):168: FutureWarning: Downcasting behavior in `replace` is deprecated and will be removed in a future version. To retain the old behavior, explicitly call `result.infer_objects(copy=False)`. To opt-in to the future behavior, set `pd.set_option('[future.no](http://future.no/)_silent_downcasting', True)`
  df.replace("", np.nan, inplace=True)
/usr/local/lib/python3.10/dist-packages/sssom/[util.py](http://util.py/):447: FutureWarning: The provided callable <built-in function max> is currently using np.maximum.reduce. In a future version of pandas, the provided callable will be used directly. To keep current behavior pass the string np.maximum.reduce instead.
  dfmax = df.groupby(key, as_index=False)[CONFIDENCE].apply(max).drop_duplicates()
/usr/local/lib/python3.10/dist-packages/sssom/[util.py](http://util.py/):447: FutureWarning: The provided callable <built-in function max> is currently using np.maximum.reduce. In a future version of pandas, the provided callable will be used directly. To keep current behavior pass the string np.maximum.reduce instead.
  dfmax = df.groupby(key, as_index=False)[CONFIDENCE].apply(max).drop_duplicates()
/usr/local/lib/python3.10/dist-packages/sssom/[util.py](http://util.py/):447: FutureWarning: The provided callable <built-in function max> is currently using np.maximum.reduce. In a future version of pandas, the provided callable will be used directly. To keep current behavior pass the string np.maximum.reduce instead.
  dfmax = df.groupby(key, as_index=False)[CONFIDENCE].apply(max).drop_duplicates()
/usr/local/lib/python3.10/dist-packages/sssom/[util.py](http://util.py/):447: FutureWarning: The provided callable <built-in function max> is currently using np.maximum.reduce. In a future version of pandas, the provided callable will be used directly. To keep current behavior pass the string np.maximum.reduce instead.
  dfmax = df.groupby(key, as_index=False)[CONFIDENCE].apply(max).drop_duplicates()
/usr/local/lib/python3.10/dist-packages/sssom/[util.py](http://util.py/):447: FutureWarning: The provided callable <built-in function max> is currently using np.maximum.reduce. In a future version of pandas, the provided callable will be used directly. To keep current behavior pass the string np.maximum.reduce instead.
  dfmax = df.groupby(key, as_index=False)[CONFIDENCE].apply(max).drop_duplicates()
/usr/local/lib/python3.10/dist-packages/sssom/[util.py](http://util.py/):447: FutureWarning: The provided callable <built-in function max> is currently using np.maximum.reduce. In a future version of pandas, the provided callable will be used directly. To keep current behavior pass the string np.maximum.reduce instead.
  dfmax = df.groupby(key, as_index=False)[CONFIDENCE].apply(max).drop_duplicates()
  File "/work/src/ontology/../scripts/[lexmatch-sssom-compare.py](http://lexmatch-sssom-compare.py/)", line 403, in <module>
    main()
  File "/usr/local/lib/python3.10/dist-packages/click/[core.py](http://core.py/)", line 1157, in __call__
    return self.main(*args, **kwargs)
  File "/usr/local/lib/python3.10/dist-packages/click/[core.py](http://core.py/)", line 1078, in main
    rv = self.invoke(ctx)
  File "/usr/local/lib/python3.10/dist-packages/click/[core.py](http://core.py/)", line 1688, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/local/lib/python3.10/dist-packages/click/[core.py](http://core.py/)", line 1434, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/local/lib/python3.10/dist-packages/click/[core.py](http://core.py/)", line 783, in invoke
    return __callback(*args, **kwargs)
  File "/work/src/ontology/../scripts/[lexmatch-sssom-compare.py](http://lexmatch-sssom-compare.py/)", line 190, in extract_unmapped_matches
    unmapped_ont_df = get_unmapped_df(
  File "/work/src/ontology/../scripts/[lexmatch-sssom-compare.py](http://lexmatch-sssom-compare.py/)", line 299, in get_unmapped_df
    filtered_new_df = filter_redundant_rows(new_df)
  File "/usr/local/lib/python3.10/dist-packages/sssom/[util.py](http://util.py/)", line 449, in filter_redundant_rows
    for _, row in dfmax.iterrows():
  File "/usr/local/lib/python3.10/dist-packages/pandas/core/[generic.py](http://generic.py/)", line 6299, in __getattr__
    return object.__getattribute__(self, name)
AttributeError: 'Series' object has no attribute 'iterrows'
make[1]: *** [mondo-ingest.Makefile:447: lexmatch/[README.md](http://readme.md/)] Error 1
rm imports/ro_terms_combined.txt
make[1]: Leaving directory '/work/src/ontology'
make: *** [mondo-ingest.Makefile:333: build-mondo-ingest] Error 2
Command exited with non-zero status 2
```

</p>
</details> 

I patched this case here: https://github.com/monarch-initiative/mondo-ingest/pull/581, basically adding some dummy confidence values to the data frame. 

Again, the case was: there was a `confidence` column, but with no "legal" float values in there.

## Action items
- [ ] 1. Write a test that confirms that this is a problem (maybe I was wrong in the diagnosis - the "patch" worked in any case)
- [ ] 2. Fix whatever makes the function fail

hrshdhgd:
Ok, so here was the problem:

When the dataframe whose redundant rows had to be filtered out had all `NaN` values for confidence, the line 

https://github.com/mapping-commons/sssom-py/blob/550206721911f711ee678eb1a8da50591649bd04/src/sssom/util.py#L441

returned `df` = Empty dataframe and the entire source data frame = `nan_df`. 

Due to this, the following line:

https://github.com/mapping-commons/sssom-py/blob/550206721911f711ee678eb1a8da50591649bd04/src/sssom/util.py#L447

result in `dfmax = {}` which is of type `pandas.Series`. Hence the confusion. 

The correct way to handle this is simple adding an `if` statement:

https://github.com/mapping-commons/sssom-py/blob/ffa2109616020f994196cbb827d71bca17192014/src/sssom/util.py#L447-L469

I've added an explicit test and it passes.  Fixes #546

### DIFF
--- a/src/sssom/util.py
+++ b/src/sssom/util.py
@@ -444,27 +444,29 @@ def filter_redundant_rows(df: pd.DataFrame, ignore_predicate: bool = False) -> p
     else:
         key = [SUBJECT_ID, OBJECT_ID, PREDICATE_ID]
     dfmax: pd.DataFrame
-    dfmax = df.groupby(key, as_index=False)[CONFIDENCE].apply(max).drop_duplicates()
-    max_conf: Dict[Tuple[str, ...], float] = {}
-    for _, row in dfmax.iterrows():
+    if not df.empty:
+        dfmax = df.groupby(key, as_index=False)[CONFIDENCE].apply(max).drop_duplicates()
+        max_conf: Dict[Tuple[str, ...], float] = {}
+        for _, row in dfmax.iterrows():
+            if ignore_predicate:
+                max_conf[(row[SUBJECT_ID], row[OBJECT_ID])] = row[CONFIDENCE]
+            else:
+                max_conf[(row[SUBJECT_ID], row[OBJECT_ID], row[PREDICATE_ID])] = row[CONFIDENCE]
         if ignore_predicate:
-            max_conf[(row[SUBJECT_ID], row[OBJECT_ID])] = row[CONFIDENCE]
+            df = df[
+                df.apply(
+                    lambda x: x[CONFIDENCE] >= max_conf[(x[SUBJECT_ID], x[OBJECT_ID])],
+                    axis=1,
+                )
+            ]
         else:
-            max_conf[(row[SUBJECT_ID], row[OBJECT_ID], row[PREDICATE_ID])] = row[CONFIDENCE]
-    if ignore_predicate:
-        df = df[
-            df.apply(
-                lambda x: x[CONFIDENCE] >= max_conf[(x[SUBJECT_ID], x[OBJECT_ID])],
-                axis=1,
-            )
-        ]
-    else:
-        df = df[
-            df.apply(
-                lambda x: x[CONFIDENCE] >= max_conf[(x[SUBJECT_ID], x[OBJECT_ID], x[PREDICATE_ID])],
-                axis=1,
-            )
-        ]
+            df = df[
+                df.apply(
+                    lambda x: x[CONFIDENCE]
+                    >= max_conf[(x[SUBJECT_ID], x[OBJECT_ID], x[PREDICATE_ID])],
+                    axis=1,
+                )
+            ]
     # We are preserving confidence = NaN rows without making assumptions.
     # This means that there are potential duplicate mappings
     # FutureWarning: The frame.append method is deprecated and
--- a/src/sssom/util.py
+++ b/src/sssom/util.py
@@ -469,10 +469,7 @@ def filter_redundant_rows(df: pd.DataFrame, ignore_predicate: bool = False) -> p
             ]
     # We are preserving confidence = NaN rows without making assumptions.
     # This means that there are potential duplicate mappings
-    # FutureWarning: The frame.append method is deprecated and
-    # will be removed from pandas in a future version.
-    # Use pandas.concat instead.
-    # return_df = df.append(nan_df).drop_duplicates()
+
     confidence_reconciled_df = pd.concat([df, nan_df]).drop_duplicates()
 
     # Reconciling dataframe rows based on the predicates with equal confidence.
--- a/tests/test_reconcile.py
+++ b/tests/test_reconcile.py
@@ -27,7 +27,7 @@ class TestReconcile(unittest.TestCase):
 
         self.msdf1.df["confidence"] = np.NAN
         df3 = filter_redundant_rows(self.msdf1.df)
-        self.assertEqual(10, len(df3.index))
+        self.assertEqual(11, len(df3.index))
 
     def test_deal_with_negation(self):
         """Test handling negating returns the right number of rows."""
--- a/tests/test_reconcile.py
+++ b/tests/test_reconcile.py
@@ -22,6 +22,13 @@ class TestReconcile(unittest.TestCase):
         df2 = filter_redundant_rows(self.msdf2.df)
         self.assertEqual(18, len(df2.index))
 
+        # Create a new dataframe with the confidence column having NaN values
+        import numpy as np
+
+        self.msdf1.df["confidence"] = np.NAN
+        df3 = filter_redundant_rows(self.msdf1.df)
+        self.assertEqual(10, len(df3.index))
+
     def test_deal_with_negation(self):
         """Test handling negating returns the right number of rows."""
         df1 = deal_with_negation(self.msdf1.df)


src/sssom/util.py
```diff
@@ -444,33 +444,32 @@ def filter_redundant_rows(df: pd.DataFrame, ignore_predicate: bool = False) -> p
     else:
         key = [SUBJECT_ID, OBJECT_ID, PREDICATE_ID]
     dfmax: pd.DataFrame
-    dfmax = df.groupby(key, as_index=False)[CONFIDENCE].apply(max).drop_duplicates()
-    max_conf: Dict[Tuple[str, ...], float] = {}
-    for _, row in dfmax.iterrows():
+    if not df.empty:
```
twhetzel, line 447: Should this check be on `df` or `df["confidence"]` and check if `df["confidence"]` is `np.nan` or `''`? 

In the case where this failed in the mondo-ingest build iI believe t was due to the `df["confidence"]` being one of `np.nan` or `''`.

hrshdhgd: it is handled in case of `np.NAN` which was the reason of the error. In the case of `""`, I should not throw any error.

matentzn: Shouldnt we prevent the case that 

https://github.com/mapping-commons/sssom-py/blob/78eacb3a11b08500a1688fdbe4bfe1940fcfef8e/src/sssom/util.py#L441

returns an df.empty in the first place?

src/sssom/util.py
```diff
@@ -444,33 +444,32 @@ def filter_redundant_rows(df: pd.DataFrame, ignore_predicate: bool = False) -> p
...
+                )
+            ]
     # We are preserving confidence = NaN rows without making assumptions.
     # This means that there are potential duplicate mappings
-    # FutureWarning: The frame.append method is deprecated and
-    # will be removed from pandas in a future version.
-    # Use pandas.concat instead.
-    # return_df = df.append(nan_df).drop_duplicates()
+
     confidence_reconciled_df = pd.concat([df, nan_df]).drop_duplicates()
```
matentzn, line 473: What exactly happens `if df.empty` because

https://github.com/mapping-commons/sssom-py/blob/78eacb3a11b08500a1688fdbe4bfe1940fcfef8e/src/sssom/util.py#L441

returns an empty df?

hrshdhgd: if `df.empty:` then it just gets ignored and `nan_df` is the df of interest (empty or not).

src/sssom/util.py
hrshdhgd: Not really. The use case itself had all rows for the column `confidence` to be `np.NAN`. `nan_df` becomes the core df of interest an we move forward with it.

src/sssom/util.py
matentzn: so df.concat does not choke with am empty df? Just making sure, sorry.

src/sssom/util.py
matentzn: I am wondering about the case where the original df is not empty, but after cycling through the assign_default_confidence() becomes empty. This would suggest a bug in `assign_default_confidence`. You are saying this cannot happen?

hrshdhgd: It should not become empty. `assign_default_confidence` only adds value to the `confidence` column. There is no scenario where it would convert the entire df to be empty

src/sssom/util.py
hrshdhgd: It will not. I ran it locally just to make sure.

matentzn: Cheers.

src/sssom/util.py
matentzn: Ahhh ok!

matentzn (approved):
LGTM, thank you!

PR #552, opened 2024-08-07
Update for latest changes of the SSSOM schema

gouttegd:
This PR updates SSSOM-Py to prepare it for changes that either have already occurred or that will occur in the SSSOM schema:

* the `semantic_similarity_{score|measure}` slots have been renamed to `similarity_{score|measure}`;
* the `subject_id` and `object_id` slots will no longer be _unconditionally_ required, but instead will be required or not depending on the value of `subject_type` and `object_type` (something the generated Python class is unfortunately unable to enforce on its own, despite the proper `rules` expressed in the schema).

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -7,6 +7,7 @@ import pandas as pd
 import yaml
 from curies import Converter, Record, chain
 from sssom_schema import Mapping as SSSOM_Mapping
+from sssom_schema import slots as SSSOM_Slots
 
 from sssom.constants import (
     CREATOR_ID,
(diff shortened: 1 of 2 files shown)


tests/test_utils.py
```diff
@@ -409,14 +410,19 @@ def test_get_dict_from_mapping(self):
...
             "other": "",
             "comment": "",
         }
 
+        if hasattr(SSSOM_Slots, "similarity_score"):
+            expected_result["similarity_score"] = np.nan
+            expected_result["similarity_measure"] = ""
+        else:
+            expected_result["semantic_similarity_score"] = np.nan
+            expected_result["semantic_similarity_measure"] = ""
```
matentzn, line 424: Are you married to this? I can see why this would make sense in the framework code (to be backwards compatible with the 0.5.X lineage), but this is the test suite.. I'd be inclined to just fix the tests to use `similarity_score` instead?

matentzn (commented):
Awesome, I didnt believe you would even touch sssom-py!! Much much much appreciated!!

tests/test_utils.py
gouttegd: But if we do that, then the test will fail until (1) we publish the next version of the SSSOM schema (one that includes the `semantic_similarity_score` -> `similarity_score` renaming) and (2) SSSOM-Py’s dependencies are updated to depend on the new SSSOM schema.

And we can’t test the next version of the SSSOM schema until that changes is in, so we are in a kind of catch-22 here.

SSSOM-Py’s test suite currently can only pass with SSSOM-Schema <= 0.15.2. If we simply “fix the tests to use `similary_score` instead”, then it will only pass with the next version of the SSSOM-Schema, and will no longer pass with <= 0.15.2.

The way the PR does it, the test suite can accommodate both SSSOM-Schema <= 0.15.2 and the to-be-published version. It’s well worth the additional complexity of _one_ extra conditional instruction in my opinion.

gouttegd:
> I didnt believe you would even touch sssom-py

Not that I _like_ that (you know my opinion on Python… right? :p ), but I am the one who pushed for this method of representing literal mappings, so the least I can do is help sorting the resulting issues.

tests/test_utils.py
matentzn: Sold. Thanks for explaining!

matentzn (approved):

PR #559, opened 2024-11-06
Ensure empty last line isnt printed when writing TSV

matentzn:
when printing df.to_csv(sep=sep, index=False), a line break is added automatically;

print(x) is also adding a line break, so when the dataframe string is printed, two line breaks are printed, leading to an empty line.

I also refactored the `lines` part a bit moving it inside the conditional, as they are only needed there.


anitacaron (approved):
LGTM

PR #560, opened 2024-11-06
suppress the FutureWarning: Downcasting behavior in `replace` is depr…

matentzn:
…ecated

Previously we got

```
/Users/matentzn/ws/sssom-py/src/sssom/util.py:168: FutureWarning: Downcasting behavior in `replace` is deprecated and will be removed in a future version. To retain the old behavior, explicitly call `result.infer_objects(copy=False)`. To opt-in to the future behavior, set `pd.set_option('future.no_silent_downcasting', True)`
  df = df.replace("", np.nan)
```

Here I am hiding the root cause of the issue (which would require me to ensure the I am not type casting accidentally while replacing) by surpressing the warning.


matentzn:
Closing in favour of #561

PR #562, opened 2024-11-09
Ineffectual PR to add my commits messages back

matentzn:


gouttegd (approved):
This is a **weird** maneuver. Approved if you want, but frankly not convinced of the usefulness. It only makes the history even more complicated.

matentzn:
Better not make things even worse than I already did..

PR #563, opened 2024-11-09
NEVER MERGE Semra testing

matentzn:
Parking some stached files here for future


matentzn:
> 
FYI there were some major updates motivated by Jakob Voß's comments, the SeMRA I/O for SSSOM is now supporting reading all of SSSOM and outputting a valid subset of SSSOM


Very cool!

PR #564, opened 2024-11-29
Add more logging to validator and change some comments/msges

matentzn:


ehartley (approved):
Looks good!

matentzn:
Thank you @ehartley - best to use approve feature as we have set some strict rules here for protecting the main branch :) Merged using godmode

PR #565, opened 2024-12-07
Extend parse_sssom_table to report wrong prefixes and metadata

matentzn:
This also comes with the introduction of a strict mode, which allows draconian failing in cases where small violations are encountered (by default, we drop unknown metadata elements, and redefined built in prefixes).

So the idea is:

Look for some basic errors before the actual parsing occurs, fail if in strict mode, if not proceed to normal mode with repair.

@gouttegd feel free to check, I can already see three comments you will have, but its best I do small steps rather than none :)

### DIFF
--- a/src/sssom/parsers.py
+++ b/src/sssom/parsers.py
@@ -184,8 +184,7 @@ def _get_seperator_symbol_from_file_path(file):
 
 def _is_check_valid_extension_slot(slot_name, meta):
     extension_definitions = meta.get("extension_definitions", [])
-    return any(entry.get("slot_name") == slot_name for entry in extension_definitions)
-
+    return any("property" in entry and entry.get("slot_name") == slot_name for entry in extension_definitions)
 
 def _is_irregular_metadata(metadata_list: List[Dict]):
     fail_metadata = False
--- a/src/sssom/parsers.py
+++ b/src/sssom/parsers.py
@@ -182,6 +182,97 @@ def _get_seperator_symbol_from_file_path(file):
     return None
 
 
+def _is_check_valid_extension_slot(slot_name):
+    logging.warning(
+        f"'{slot_name}' could be a valid extension slot "
+        f"(https://mapping-commons.github.io/sssom/spec-model/#non-standard-slots), "
+        f"but the validator does not check that yet."
+    )
+    return False
+
+
+def _check_irregular_metadata(sssom_metadata, meta):
+    fail_metadata = False
+    for m in [sssom_metadata, meta]:
+        for key in m:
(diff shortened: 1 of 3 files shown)


src/sssom/parsers.py
```diff
@@ -182,6 +182,97 @@ def _get_seperator_symbol_from_file_path(file):
...
+                        f"to the required URI expansion: {builtin_uri}. The prefix will be ignored."
+                    )
+                    is_valid_prefixes = False
+            reverse_bimap = {value: key for key, value in builtin_converter.bimap.items()}
+            if builtin_uri in reverse_bimap:
+                if builtin_prefix != reverse_bimap[builtin_uri]:
+                    logging.warning(
+                        f"A built-in URI namespace ({builtin_uri}) was used in (one of) the provided prefix map(s), "
+                        f"but the provided prefix ({reverse_bimap[builtin_uri]}) does not correspond to the "
+                        f"standard prefix: {builtin_prefix}. The prefix will be ignored."
```
gouttegd, line 235: I’ll just note that this case (standard prefix associated to a different prefix name than the one listed in the spec, e.g. `som: https://w3id.org/sssom/`) is _not_ forbidden by the spec. The spec only forbids associated a builtin prefix name to a different prefix (e.g., `sssom: https://not-the-sssom-namespace.example.org/`).

The spec says:

> By exception, prefix names listed in the table found in the IRI prefixes section are considered “built-in”. As such, they MAY be omitted from the curie_map. If they are not omitted, they MUST point to the same IRI prefixes as in the aforementioned table.

There is nothing here that forbids using another prefix name than a builtin one to refer to one of the known namespaces.

Not requesting any change here because SSSOM-Py is free to be stricter than the spec if it wants to.

gouttegd (approved):
Looks good, but mind that you are being stricter than what the spec requires.

src/sssom/parsers.py
```diff
@@ -182,6 +182,95 @@ def _get_seperator_symbol_from_file_path(file):
     return None
 
 
+def _is_check_valid_extension_slot(slot_name, meta):
+    extension_definitions = meta.get("extension_definitions", [])
+    return any(entry.get("slot_name") == slot_name for entry in extension_definitions)
```
gouttegd, line 187: I would personally like a check that the extension definition does have a `property` (which it SHOULD have), and a warning if it does not.

matentzn: Done! https://github.com/mapping-commons/sssom-py/commit/57d24f9bbdaf7b640e0d9d7dc6e4ff04cdaf9ec2

PR #579, opened 2025-04-29
Return validation reports

cthoyt:
I'm currently trying to use `sssom.io.validate_path` to check that SSSOM files generated by SeMRA are valid, but it's very difficult to get access to the results of validation programmatically. There _should_ be some logging output to stderr, but this isn't so helpful in a unit testing environment.

This PR makes the validation functions return the reports generated by LinkML, so they can be returned and used downstream.

This PR also operationalizes the default validator list, meaning you don't have to import them or know what they are to get this function up and running.

Blocked by #580

Alternative to #387, but that PR uses custom results instead of returning the LinkML objects

### DIFF
--- a/src/sssom/io.py
+++ b/src/sssom/io.py
@@ -19,7 +19,6 @@ from sssom.validators import validate
 
 from .constants import (
     CURIE_MAP,
-    DEFAULT_VALIDATION_TYPES,
     PREFIX_MAP_MODE_MERGED,
     PREFIX_MAP_MODE_METADATA_ONLY,
     PREFIX_MAP_MODE_SSSOM_DEFAULT_ONLY,
--- a/src/sssom/io.py
+++ b/src/sssom/io.py
@@ -102,7 +102,7 @@ def parse_file(
 def validate_file(
     input_path: str,
     *,
-    validation_types: List[SchemaValidationType] | None = None,
+    validation_types: Optional[List[SchemaValidationType]] = None,
     fail_on_error: bool = True,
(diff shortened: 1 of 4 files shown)


src/sssom/io.py
```diff
@@ -98,17 +99,24 @@ def parse_file(
     write_table(doc, output, embedded_mode)
 
 
-def validate_file(input_path: str, validation_types: List[SchemaValidationType]) -> None:
+def validate_file(
+    input_path: str,
+    *,
```
matentzn, line 104: what happens with implementations that call `validate_file(a,b)` without the argument names, if the `*` is the second option?

matentzn (commented):
Just a minor comment but this looks great!  THANKS a ton and a million sorry for the delayed review!

src/sssom/io.py
cthoyt: I guess this is breaking, but it's good form to have arguments with defaults be keyword only. I can revert this

cthoyt: fixed in d96e3eb and 8b4011f

matentzn: Thank you!

cthoyt:
fyi this PR won't pass CI until #580 is merged since there are unrelated things that are broken now that #580 fixes

cthoyt:
sorry @matentzn I need a re-review, since there was a test I hadn't noticed had failed; related to checking the validate function returned none, and now it returns something that we can actually check in more depth!

matentzn (approved):
Thanks for the test fixes

PR #580, opened 2025-04-29
Clean up CURIE parsing

cthoyt:
- [x] update RDF parsing
- [ ] update prefix discovery

### DIFF
--- a/src/sssom/rdf_util.py
+++ b/src/sssom/rdf_util.py
@@ -3,6 +3,7 @@
 import logging
 from typing import Any, Dict, List, Optional
 
+import curies
 from linkml_runtime.utils.metamodelcore import URIorCURIE
 from rdflib import Graph, URIRef
 from sssom_schema import EntityReference, Mapping
@@ -26,7 +27,7 @@ def rewire_graph(
     if mdoc.mapping_set.mappings is None:
         raise TypeError
 
-    converter = mdoc.converter
+    converter: curies.Converter = mdoc.converter
     rewire_map: Dict[URIorCURIE, URIorCURIE] = {}
     for m in mdoc.mapping_set.mappings:
         if not isinstance(m, Mapping):
@@ -42,8 +43,8 @@ def rewire_graph(
                 curr_tgt = rewire_map[src]
                 logging.info(f"Ambiguous: {src} -> {tgt} vs {curr_tgt}")
                 if precedence:
-                    curr_pfx, _ = converter.parse_curie(curr_tgt)
-                    tgt_pfx, _ = converter.parse_curie(tgt)
+                    curr_pfx = converter.parse_curie(curr_tgt, strict=True).prefix
+                    tgt_pfx = converter.parse_curie(tgt, strict=True).prefix
                     if tgt_pfx in precedence:
                         if curr_pfx not in precedence or precedence.index(
                             tgt_pfx


src/sssom/rdf_util.py
```diff
@@ -42,8 +43,14 @@ def rewire_graph(
                 curr_tgt = rewire_map[src]
                 logging.info(f"Ambiguous: {src} -> {tgt} vs {curr_tgt}")
                 if precedence:
-                    curr_pfx, _ = converter.parse_curie(curr_tgt)
-                    tgt_pfx, _ = converter.parse_curie(tgt)
+                    curr_reference = converter.parse_curie(curr_tgt)
+                    if curr_reference is None:
+                        raise ValueError
```
matentzn, line 48: Would it make sense to provide a somewhat meaningful message here?

src/sssom/rdf_util.py
```diff
@@ -42,8 +43,14 @@ def rewire_graph(
...
+                    target_reference = converter.parse_curie(tgt)
+                    if target_reference is None:
+                        raise ValueError
```
matentzn, line 51: Would it make sense to provide a somewhat meaningful message here?

src/sssom/util.py
```diff
@@ -1118,20 +1118,20 @@ def _is_iri(string: str) -> bool:
 def get_prefix_from_curie(curie: str) -> str:
     """Get the prefix from a CURIE."""
     if _is_curie(curie):
-        return curie.split(":")[0]
+        return ReferenceTuple.from_curie(curie).prefix
```
matentzn, line 1121: Cool: https://curies.readthedocs.io/en/latest/api/curies.ReferenceTuple.html

Thanks for that!

src/sssom/util.py
```diff
@@ -1118,20 +1118,20 @@ def _is_iri(string: str) -> bool:
...
+def get_prefixes_used_in_table(df: pd.DataFrame) -> Set[str]:
     """Get a list of prefixes used in CURIEs in key feature columns in a dataframe."""
     prefixes = set(SSSOM_BUILT_IN_PREFIXES)
     if df.empty:
         return prefixes
     sssom_schema_object = _get_sssom_schema_object()
     entity_reference_slots = sssom_schema_object.entity_reference_slots & set(df.columns)
     new_prefixes = {
-        converter.parse_curie(row).prefix
+        ReferenceTuple.from_curie(row).prefix
```
matentzn, line 1134: Does this change the behaviour of the method slightly?

Did `converter.parse_curie` obtain a prefix _only_ of the specific converter actually had that prefix?

If so, the method `get_prefixes_used_in_table` was previously returning "get_valid_prefixes_used_in_table_according_to_prefix_map", right?

I am not objecting to the change, just wanted to make sure this was indeed happening.

cthoyt: the old implementation of `curies.Converter.parse_curie` just deferred to `        ReferenceTuple.from_curie(row).prefix`. Now, the implementation validates against the prefix map. So this change maintains current SSSOM functionality, where the goal was to find _all_ prefixes, not just valid ones

src/sssom/rdf_util.py
cthoyt: Good point. Better error messages are now generated by a newer version of `curies`, so I bumped the minimum version to take care of that in d355756

src/sssom/util.py
matentzn: Thank you!

matentzn (approved):

PR #582, opened 2025-04-29
Handle gzipped SSSOM files

### LINKED ISSUE #581: `sssom.io.parse_sssom_table` fails when passing path to gzipped file
It would be nice to update this to accept a gzipped file. Need to update the following to check the file path extension:

https://github.com/mapping-commons/sssom-py/blob/3a991034869bd40bcf0557d27d9f3d078b6f562c/src/sssom/parsers.py#L99-L103

cthoyt:
Closes #581

### DIFF
--- a/src/sssom/parsers.py
+++ b/src/sssom/parsers.py
@@ -1,5 +1,6 @@
 """SSSOM parsers."""
 
+import gzip
 import io
 import itertools as itt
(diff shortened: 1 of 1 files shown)


matentzn (approved):
I assume this was tested.

THANK YOU!

PR #586, opened 2025-05-22
Update poetry to 2.1.3

cthoyt:
This repository is currently using poetry 1.6.1, released in August 2023. A lot has changed in Python packaging since then, including the crystallization of the pyproject.toml standard metadata, dependency groups, and more.

This PR gets ready for revitalizing SSSOM-py by first updating Poetry to a newer version. This makes a small amount of diff in the poetry lock file.

Note: poetry fixed its resolver since 1.6.1 and it's also muchhhh faster now


matentzn (approved):
Yes, I was not aware that its best practice to hard code the poetry version but seems it is. I will approve.

cthoyt:
I don't know if it's best practice or not. I would rather not do it and rely on the most recent version, but I didn't want this update to be controversial

PR #591, opened 2025-05-24
Simplify reading CSV/TSV

cthoyt:
This PR reduces complexity in the way TSV files are read, requiring only a single pass through the file instead of two.

This PR does not need new tests, as the code that is changed is part of a private API, which is thoroughly tested by reading many SSSOM documents via the user-facing functions that call this.

### DIFF
--- a/src/sssom/parsers.py
+++ b/src/sssom/parsers.py
@@ -132,11 +132,11 @@ def _read_pandas_and_metadata(file_path: Union[str, Path, TextIO], sep: Optional
     if isinstance(file_path, (str, Path)):
         raise_for_bad_path(file_path)
 
-    stream = _open_input(file_path)
+    table_stream = _open_input(file_path)
 
     # consume from the top of the stream until there's no more preceding #
     header_yaml = ""
-    while (line := stream.readline()).startswith("#"):
+    while (line := table_stream.readline()).startswith("#"):
         line = line.lstrip("#").rstrip()
         if not line:
             continue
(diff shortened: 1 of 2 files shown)


src/sssom/parsers.py
```diff
@@ -119,57 +118,40 @@ def _open_input(p: PathOrIO) -> TextIO:
...
     if isinstance(file_path, (str, Path)):
         raise_for_bad_path(file_path)
 
     stream = _open_input(file_path)
-    table_stream, metadata_stream = _separate_metadata_and_table_from_stream(stream)
+
+    # consume from the top of the stream until there's no more preceding #
+    header_yaml = ""
+    while (line := stream.readline()).startswith("#"):
+        line = line.lstrip("#").rstrip()
```
matentzn, line 141: Should this be:

```suggestion
        line = line.lstrip("#").lstrip().rstrip()
```

To account for

```
# whitespace_after_has_and_before_w: a
```

tests/test_parsers.py
```diff
@@ -470,8 +470,11 @@ def test_strict_parsing(self):
             parse_sssom_table(stream, strict=True)
 
         # Make sure it parses in non-strict mode
-        msdf = parse_sssom_table(stream)
-        self.assertEqual(len(msdf.df), 2)
+        with open(input_path, "r") as file:
```
matentzn, line 473: Why open the file again? it seemed to have worked before?

matentzn (commented):
Thanks a ton! Two minor questions :)

tests/test_parsers.py
cthoyt: Now, we only consume the file once and don't put it into a buffer that we can seek to the top of. The old way of reading the same stream twice was also problematic since you shouldn't assume this is even possible

src/sssom/parsers.py
cthoyt: You can't do a full lstrip since that would break the indentation of the curie map. This is a good point, but I'm maintaining the status quo here. We could reinvestigate if it should be possible to have leading whitespace in a different thread

tests/test_parsers.py
matentzn: ok

cthoyt:
i'm not going to engage with half baked comments from an AI, sorry!

matentzn (approved):
Thank you @cthoyt for you great contribution again! :)

Issue #554, opened 2024-10-23
Mapping set inversion fails if the set contains a `subject_*` column but no corresponding `object_*` column


gouttegd:
Trying to `sssom invert` a mapping set that contains a `subject_label` column but no `object_label` yields an error because `sssom-py` somehow expects to find a `subject_label` column in the inverted set, even though the inverted set will (logically) only contain a `object_label` column.

Example: given the following minimalist set:

```
#curie_map:
#  COMENT: https://example.com/entities/
#  ORGENT: https://example.org/entities/
#mapping_set_id: https://example.org/sets/exo2c
subject_id      subject_label  predicate_id    object_id       mapping_justification
ORGENT:0001     alice      skos:closeMatch COMENT:0011     semapv:ManualMappingCuration
```

the following command:

```sh
sssom-dev invert sample.sssom.tsv -o inverted.sssom.tsv
```

yields the following error:

```
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/home/damien/.cache/pypoetry/virtualenvs/sssom-tFe7wmy3-py3.9/lib64/python3.9/site-packages/click/core.py", line 1157, in __call__
    return self.main(*args, **kwargs)
  File "/home/damien/.cache/pypoetry/virtualenvs/sssom-tFe7wmy3-py3.9/lib64/python3.9/site-packages/click/core.py", line 1078, in main
    rv = self.invoke(ctx)
  File "/home/damien/.cache/pypoetry/virtualenvs/sssom-tFe7wmy3-py3.9/lib64/python3.9/site-packages/click/core.py", line 1688, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/damien/.cache/pypoetry/virtualenvs/sssom-tFe7wmy3-py3.9/lib64/python3.9/site-packages/click/core.py", line 1434, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/damien/.cache/pypoetry/virtualenvs/sssom-tFe7wmy3-py3.9/lib64/python3.9/site-packages/click/core.py", line 783, in invoke
    return __callback(*args, **kwargs)
  File "/home/damien/dvlpt/semweb/python/sssom-py/src/sssom/cli.py", line 769, in invert
    msdf.df = invert_mappings(
  File "/home/damien/dvlpt/semweb/python/sssom-py/src/sssom/util.py", line 1467, in invert_mappings
[...]

PR #595, opened 2025-07-16
Fix inversion of mappings containing `subject_type` or `object_type` slots.

gouttegd:
Mapping inversion requires, among other things, to know which are the slots whose value must be inverted. For example, knowing that `subject_id` must be inverted with `object_id`, `subject_label` must be inverted with `object_label`, and so on.
    
This knowledge is provided, in SSSOM-Py, by the `COLUMN_INVERT_DICTIONARY` map in the `sssom.constants` module.
    
That map is currently missing the `subject_type` and `object_type` slots, resulting in SSSOM-Py [being unable](https://github.com/mapping-commons/sssom-py/issues/554#issuecomment-3078993519) to invert any mapping set that contains one of those slots.

This PR adds the missing slots to the `COLUMN_INVERT_DICTIONARY`.


matentzn (approved):
Thank you @gouttegd much appreciated!!

Issue #599, opened 2025-08-16
`scipy` should not be an optional dependency


gouttegd:
PR #394 made `scipy` an optional dependency, but the module is always _unconditionally_ imported by the main CLI module (`sssom/cli.py`):

```python
from scipy.stats import chi2_contingency
```

As a result, with a _default_ installation, the `sssom` command is completely unusable, since the entire `cli.py` module cannot be loaded. Even just trying to print the online help will fail:

```sh
$ poetry install
$ poetry run sssom --help
ModuleNotFoundError: No module named 'scipy'
```

One needs to take care of installing the `scipy` extra dependency group for the command to become usable:

```sh
$ poetry install --extras=scipy
$ poetry run sssom --help
Usage: sssom [OPTIONS] COMMAND [ARGS]...

```

If the `sssom` command cannot be used at all without `scipy`, then `scipy` is _not_ an optional dependency.

PR #600, opened 2025-08-16
Error out gracefully on missing optional module

gouttegd:
Three of the dependencies of `sssom` are considered optional:

* `scipy`,
* `pansql`,
* and `networkx`.

However, wherever they are needed, those modules are always imported unconditionally and without any code to deal with the possibility that they may very well be absent.

This is especially problematic for the `scipy` module, since it is imported at the root of the `sssom.cli` module, resulting in that module being unloadable if `scipy` itself cannot be found – resulting in turn in the entire `sssom` command being unusable (#599).

For the other twos, their absence does not prevent the `sssom` command to run, but any attempt to use one of the subcommands that need them will result in a loud crash (uncaught `ModuleErrorNotFoundError`).

Since those modules are _optional_, the `sssom` command-line tool should deal gracefully with their possible absence. That is, it should not spit out an exception stack trace, but instead print out a simple error message highlighting that the particular subcommand the user tried to run requires an optional module.

This is what this PR does.

For `scipy`, which is only used within the `correlations` subcommand, we `try` to import the module within that subcommand, and error out if the module cannot be found.

For `pansql` and `networkx`, we guard all calls to functions that use them in a `try` block to catch the specific error indicating that they are not present.

### DIFF
--- a/src/sssom/cli.py
+++ b/src/sssom/cli.py
@@ -423,7 +423,9 @@ def partition(inputs: List[str], output_directory: str):
         cliquedocs = split_into_cliques(doc)
     except ModuleNotFoundError as e:
         if e.name == "networkx":
-            raise click.ClickException("The partition command requires the optional networkx module.")
+            raise click.ClickException(
+                "The partition command requires the optional networkx module."
+            )
         raise
     for n, cdoc in enumerate(cliquedocs, start=1):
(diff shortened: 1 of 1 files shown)


cthoyt (approved):
this is great!

Issue #584, opened 2025-04-30
Support for slot propagation and condensation


gouttegd:
Some SSSOM metadata slots, when used at the level of a _mapping set_ (rather than at the level of individual mapping records), are intended to represent the fact that all mapping records in the set have the same value. Such slots are known as “propagatable slots”.

For example, in the following set:

```
#curie_map:
#  COMENT: https://example.com/entities/
#  ORGENT: https://example.org/entities/
#mapping_set_id: https://example.org/sets/exo2c-with-propagatable-slots
#license: https://creativecommons.org/licenses/by/4.0/
#mapping_provider: https://example.org/provider
subject_id	subject_label	predicate_id	object_id	object_label	mapping_justification
ORGENT:0001	alice	skos:closeMatch	COMENT:0011	alpha	semapv:ManualMappingCuration
ORGENT:0002	bob	skos:closeMatch	COMENT:0012	beta	semapv:ManualMappingCuration
ORGENT:0004	daphne	skos:closeMatch	COMENT:0014	delta	semapv:ManualMappingCuration
```

The `mapping_provider` slot is a propagatable slot and its presence in the set-level metadata is to be interpreted as indicating that all three mapping records have `https://example.org/provider` as their “mapping provider” – in effect, as if the set had been:

```
#curie_map:
#  COMENT: https://example.com/entities/
#  ORGENT: https://example.org/entities/
#mapping_set_id: https://example.org/sets/exo2c-with-propagatable-slots
#license: https://creativecommons.org/licenses/by/4.0/
subject_id	subject_label	predicate_id	object_id	object_label	mapping_justification	mapping_provider
ORGENT:0001	alice	skos:closeMatch	COMENT:0011	alpha	semapv:ManualMappingCuration	https://example.org/provider
ORGENT:0002	bob	skos:closeMatch	COMENT:0012	beta	semapv:ManualMappingCuration	https://example.org/provider
ORGENT:0004	daphne	skos:closeMatch	COMENT:0014	delta	semapv:ManualMappingCuration	https://example.org/provider
```

“Slot propagation” is the process of taking the values of all propagatable slots in the set-level metadata and setting them to each individual mappings (i.e., the process that transforms the first set shown above into the second set). “Slot condensation” is the reverse process (if all mappings have the same value for a given propagatable slot, then that value can be removed from the mapping records themselves and set once in the set-level metadata instead).

[...]

PR #603, opened 2025-08-24
Add support for propagating/condensing slots.

gouttegd:
This PR adds partial support for condensation and propagation of slot values (#584).

“Partial” because

(1) It is only supported for a mapping set represented with the `MappingSetDataFrame` class. This seems to be _the_ standard object to represent a mapping set across SSSOM-Py (in particular, that is the object returned by all the parsers, and consumed by all the writers). In fact it’s unclear to me what the `MappingSet` and `MappingSetDocument` classes are for.

So for now, if you have a mapping set as `MappingSet` object, it still cannot be propagated or condensed directly, it needs to be converted into a `MappingSetDataFrame` first (and converted back to a `MappingSet` afterwards), as in:

```python
msdf = MappingSetDataFrame.from_mapping_set(ms)
msdf.propagate()
ms = msdf.to_mapping_set()
```

(2) Condensation/propagation is not (yet?) wired into the `sssom` command-line tool, the feature is only available from Python code.

FWIW, my opinion is that propagation should be _systematically_ performed when reading a SSSOM file, and conversely condensation should be _systematically_ performed when writing a SSSOM file. That’s what the spec recommends, and what SSSOM-Java does. But I acknowledge this would be a breaking change compared to what has always been SSSOM-Py’s behaviour (since it has never supported either operation), so I’ll let others to decide if and how propagation/condensation should be enabled directly in the SSSOM parsers and readers.

### DIFF
--- a/src/sssom/util.py
+++ b/src/sssom/util.py
@@ -58,6 +58,7 @@ from .constants import (
     UNKNOWN_IRI,
     MetadataType,
     PathOrIO,
+    SSSOMSchemaView,
     _get_sssom_schema_object,
     get_default_metadata,
 )
@@ -298,6 +299,56 @@ class MappingSetDataFrame:
 
         self.df = self.df[self.df.columns.drop(list(self.df.filter(regex=r"_2")))]
         self.clean_prefix_map()
+
+    def propagate(self) -> List[str]:
+        """Propagate slot values from the set level down to individual records.
+
+        :return: The list of slots that were effectively propagated.
+        """
+        schema = SSSOMSchemaView()
+        propagated = []
+
+        for slot in schema.propagatable_slots:
+            if slot not in self.metadata:  # Nothing to propagate
+                continue
+            if slot in self.df.columns:  # Cannot propagate
+                continue
+
+            if schema.view.get_slot(slot).multivalued:
+                value = "|".join(self.metadata.pop(slot))
+            else:
+                value = self.metadata.pop(slot)
+
+            self.df[slot] = value
+            propagated.append(slot)
+
+        return propagated
+
+    def condense(self) -> List[str]:
+        """Condense record-level slot values to the set whenever possible.
+
+        :return: The list of slots that were effectively condensed.
+        """
+        schema = SSSOMSchemaView()
+        condensed = []
+
+        for slot in schema.propagatable_slots:
+            if slot not in self.df.columns:  # Nothing to condense
+                continue
+            values = self.df[slot].unique()
+            if len(values) > 1:  # Cannot condense
+                continue
+
+            if schema.view.get_slot(slot).multivalued:
+                value = values[0].split("|")
+            else:
+                value = values[0]
+
+            self.metadata[slot] = value
+            condensed.append(slot)
+
+        self.df.drop(columns=condensed, inplace=True)
+        return condensed
 
 
 def _standardize_curie_or_iri(curie_or_iri: str, *, converter: Converter) -> str:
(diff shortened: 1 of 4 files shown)


src/sssom/util.py
```diff
@@ -299,6 +300,56 @@ def remove_mappings(self, msdf: "MappingSetDataFrame") -> None:
...
+        propagated = []
+
+        for slot in schema.propagatable_slots:
+            if slot not in self.metadata:  # Nothing to propagate
+                continue
+            if slot in self.df.columns:  # Cannot propagate
+                continue
+
+            if schema.view.get_slot(slot).multivalued:
+                value = "|".join(self.metadata.pop(slot))
```
cthoyt, line 318: when we're in dataframe world, are the lists really stored as strings that have been joined? this seems like something that should only happen during I/O

cthoyt:
> In fact it’s unclear to me what the MappingSet and MappingSetDocument classes are for.

IIRC, these are for LinkML, but I am doubtful of their value

> FWIW, my opinion is that propagation should be systematically performed when reading a SSSOM file, and conversely condensation should be systematically performed when writing a SSSOM file. That’s what the spec recommends, and what SSSOM-Java does.

fully agree

src/sssom/util.py
gouttegd: > when we're in dataframe world, are the lists really stored as strings that have been joined?

Apparently so.

Try reading a SSSOM/TSV file that has multi-valued columns, you’ll find that in the dataframe returned by the `parse_sssom_table` function the values are stored as `|`-joined strings. Dealing with those values is apparently left to client code.

src/sssom/util.py
```diff
@@ -299,6 +300,56 @@ def remove_mappings(self, msdf: "MappingSetDataFrame") -> None:
...
+            values = self.df[slot].unique()
+            if len(values) > 1:  # Cannot condense
+                continue
+
+            if schema.view.get_slot(slot).multivalued:
+                value = values[0].split("|")
+            else:
+                value = values[0]
+
+            self.metadata[slot] = value
```
gouttegd, line 347: Note to self before I forget: this is wrong, [as per the spec](https://mapping-commons.github.io/sssom/spec-formats-tsv/#condensation) condensation is only allowed if there is not already a value at the set level (or if the set-level value happens to be the same as the values for all mapping records), so here we are missing a check (does the `self.metadata` dict already have a value for the slot?).

src/sssom/util.py
```diff
@@ -299,6 +300,58 @@ def remove_mappings(self, msdf: "MappingSetDataFrame") -> None:
...
+                continue
+
+            if schema.view.get_slot(slot).multivalued:
+                value = values[0].split("|")
+            else:
+                value = values[0]
+            if slot in self.metadata and self.metadata[slot] != value:
+                continue  # Set already has a different value
+
+            self.metadata[slot] = value
```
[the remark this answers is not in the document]
cthoyt: okay, thanks. Same suggestion as in the other comment - it would be great to summarize this in a code comment.

cthoyt (dismissed):
besides adding a few code comments to reflect discussion on the PR, this looks good. Thanks @gouttegd

src/sssom/util.py
```diff
@@ -299,6 +300,99 @@ def remove_mappings(self, msdf: "MappingSetDataFrame") -> None:
...
+                # Different values across the records, cannot condense
+                continue
+
+            if schema.view.get_slot(slot).multivalued:
+                value = values[0].split("|")
+            else:
+                value = values[0]
+
+            if slot in self.metadata:
+                if self.metadata[slot] != value:
```
cthoyt, line 380: can you add one test case for each branch of this conditional? Other than that, I checked that the rest of the code in this PR is fully tested :)

cthoyt (approved):
huge thanks @gouttegd for this well-done PR. This is ready for merge

PR #604, opened 2025-09-01
Add method to infer minimum compatible SSSOM version.

gouttegd:
This PR adds a new method to the `MappingSetDataFrame` class to automatically determine the minimum version of the SSSOM specification the set is compatible with – that is, the earliest version that defines all the slots and all the enum values present in the set.

This method could later be used to implement the behaviour recommended by the spec: automatically inserting the `sssom_version` slot when a set is written, if the set requires another version than 1.0. This could simply look like:

```python
# Assuming msdf is the mapping set we have to write
min_version = msdf.get_compatible_version()
if min_version != "1.0":
    msdf.metadata["sssom_version"] = min_version
# Now we can actually write the msdf
```

### DIFF
--- a/src/sssom/util.py
+++ b/src/sssom/util.py
@@ -393,6 +393,56 @@ class MappingSetDataFrame:
         self.df.drop(columns=condensed, inplace=True)
         return condensed
 
+    def get_compatible_version(self):
+        """Get the minimum version of SSSOM this set is compatible with."""
+        schema = SSSOMSchemaView()
+        versions = set()
+
+        # First get the minimum versions required by the slots present
+        # in the set; this is entirely provided by the SSSOM model.
+        for slot in self.metadata.keys():
+            version = schema.get_minimum_version(slot, "mapping set")
+            if version is not None:
+                versions.add(str(version))
+        for slot in self.df.columns:
+            version = schema.get_minimum_version(slot, "mapping")
+            if version is not None:
+                versions.add(str(version))
+
+        # Then take care of enum values; we cannot use the SSSOM model
+        # for that (enum values are not tagged with an "added_in"
+        # annotation the way slots are), so this has to be handled
+        # "manually" based on the informations provided in
+        # <https://mapping-commons.github.io/sssom/spec-model/#model-changes-across-versions>.
+        if (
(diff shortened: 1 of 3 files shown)


src/sssom/util.py
```diff
@@ -393,6 +393,56 @@ def condense(self) -> List[str]:
...
+            and "0:0" in self.df["mapping_cardinality"].values
+        ):
+            versions.add("1.1")
+
+        # Get the highest of the accumulated versions. We do a numerical
+        # sort, so that version 1.10 (if we ever get that far in the 1.x
+        # branch) does not get sorted before version 1.9.
+        def _version_to_compare_key(version):
+            major, minor = [int(s) for s in version.split(".")]
+            return (major * 100) + minor
```
cthoyt, line 442: Python is pretty happy to compare tuples in the way you'd expect them to, so this can be simplified to:

```suggestion
            return tuple(int(s) for s in version.split("."))
```

src/sssom/util.py
```diff
@@ -393,6 +393,56 @@ def condense(self) -> List[str]:
...
+                versions.add(str(version))
+
+        # Then take care of enum values; we cannot use the SSSOM model
+        # for that (enum values are not tagged with an "added_in"
+        # annotation the way slots are), so this has to be handled
+        # "manually" based on the informations provided in
+        # <https://mapping-commons.github.io/sssom/spec-model/#model-changes-across-versions>.
+        if (
+            self.metadata.get("subject_type") == "composed entity expression"
+            or self.metadata.get("subject_type") == "composed entity expression"
```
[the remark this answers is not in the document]
gouttegd: Absolutely, my bad. :facepalm:

src/sssom/util.py
gouttegd: Nice, thanks for the tip!

matentzn (dismissed):
I just want to say a huge THANK YOU for these PRs.. Much appreciated.

I dont see any red flags, lgtm!

gouttegd:
Marking as draft to prevent inadvertent premature merging.

matentzn:
Ah we still need the sssom schema version, sorry, just remembered.

PR #605, opened 2025-09-01
Add the `infer_cardinality` method.

gouttegd:
Add a new `infer_cardinality` method to the `MappingSetDataFrame` to fill the `mapping_cardinality` slot with computed cardinality values.

The approach used here is more or less a direct Python translation of my existing implementation in SSSOM-Java. (As such, it may not be as “Pythonic” as the rest of SSSOM-Py’s codebase, but it does the job.)

The gist of it is that we iterate over the entire set of records a first time to populate two hash tables: one that associates a subject to all the different objects it is mapped to, and one that associates an object to all the different subjects it is mapped to. Then we can iterate over the records a second time, and for every record we can immediately get (1) the number of different objects mapped to the same subject and (2) the number of different subjects mapped to the same object; the combination of those two values gives us the cardinality we are looking for.

To deal with the concept of "scope", the "subjects" and "objects" that we use to fill the hash tables are not made of only the "subject_id" slot or the "object_id" slot, but also of all the slots that define the scope. For example, if the scope is `["predicate_id"]`, then given the following records:

| subject_id | predicate_id | object_id |
| ---------- | ------------ | --------- |
| DO:1234    | skos:exactMatch | HP:5678 |
| DO:1234    | skos:broadMatch | MONDO:5678 |

* for the first record the "subject" string will contain both `DO:1234` and `skos:exactMatch`, and the "object" string will contain both `HP:5678` and `skos:exactMatch`;
* for the second record the "subject" string will contain both `DO:1234` and `skos:broadMatch`, and the "object" string will contain both `MONDO:5678` and `skos:broadMatch`.

This way, the "subject" and "object" strings of these records will occupy different entries in the hash tables, thereby ensuring that they are counted separately (as they should since they are in different "scopes").

### DIFF
--- a/src/sssom/util.py
+++ b/src/sssom/util.py
@@ -23,12 +23,15 @@ from sssom_schema import Mapping as SSSOM_Mapping
 from sssom_schema import MappingSet, slots
 
 from .constants import (
+    CARDINALITY_SCOPE,
     COLUMN_INVERT_DICTIONARY,
     COMMENT,
     CONFIDENCE,
+    MAPPING_CARDINALITY,
     MAPPING_JUSTIFICATION,
     MAPPING_SET_ID,
     MAPPING_SET_SOURCE,
+    NO_TERM_FOUND,
     OBJECT_CATEGORY,
     OBJECT_ID,
     OBJECT_LABEL,
@@ -437,10 +440,7 @@ class MappingSetDataFrame:
         # We iterate over the records a first time to collect the different
         # objects mapped to each subject and vice versa
         for _, row in self.df.iterrows():
-            if (
-                row.get("subject_id") == "sssom:NoTermFound"
-                or row.get("object_id") == "sssom:NoTermFound"
-            ):
+            if row.get(SUBJECT_ID) == NO_TERM_FOUND or row.get(OBJECT_ID) == NO_TERM_FOUND:
                 # Mappings to sssom:NoTermFound are ignored for cardinality computations
                 continue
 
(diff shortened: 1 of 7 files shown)


src/sssom/util.py
```diff
@@ -393,6 +393,100 @@ def condense(self) -> List[str]:
...
+                s = "E\0" + row.get(f"{side}_id", "")
+            for slot in scope:
+                s += "\0" + row.get(slot, "")
+            return s
+
+        # We iterate over the records a first time to collect the different
+        # objects mapped to each subject and vice versa
+        for _, row in self.df.iterrows():
+            if (
+                row.get("subject_id") == "sssom:NoTermFound"
```
cthoyt, line 441: can you use constants for the fields and sssom:NoTermFound? I think many are already in the `sssom.constants` module

gouttegd: (Aside: is the `sssom.constants` module entirely hand-written? Seems to me that most of those constant declarations could, and arguably should, be generated from the LinkML schema…)

matentzn (approved):
I love it - I did not review the python code in great detail but I checked all the test cases you have and these align with my understanding of cardinality_scope! Thank you!

cthoyt:
> @cthoyt Do we care about compatibility with Python 3.9?
> 
> ```
> src/sssom/util.py:399: error: X | Y syntax for unions requires Python 3.10
> ```

it's end of life in like 1 month. But this PR isn't about dropping py39 support, so let's keep compatibility here. Sorry i am in the habit of all of my code being 310+ already

src/sssom/util.py
```diff
@@ -393,6 +396,106 @@ def condense(self) -> List[str]:
         self.df.drop(columns=condensed, inplace=True)
         return condensed
 
+    def infer_cardinality(self, scope: List[str] | None = None) -> None:
```
cthoyt, line 399:
```suggestion
    def infer_cardinality(self, scope: Optional[List[str]] = None) -> None:
```

PR #608, opened 2025-09-05
Add alternate dataframe split implementations

cthoyt:
Related to #607

PR #609, opened 2025-09-05
Split dataframe optimization

### LINKED ISSUE #607: Unnecessary repeated iteration in `split_dataframe_by_prefix`
I've been trying to track some performance issues in the [mondo](https://github.com/monarch-initiative/mondo/) pipeline, and one bottleneck I've identified is in a function that calls [`sssom.parsers.split_dataframe_by_prefix`](https://github.com/mapping-commons/sssom-py/blob/cf75d07cfea233718331365675a27dfda7bf80ba/src/sssom/parsers.py#L980-L1022). Specifically, this part here:

```
    for subject_prefix, object_prefix, relation in itt.product(
        subject_prefixes, object_prefixes, relations
    ):
 ...
        df_subset = df[
            (df[SUBJECT_ID].str.startswith(subject_prefix + ":"))
            & (df[PREDICATE_ID] == relation)
            & (df[OBJECT_ID].str.startswith(object_prefix + ":"))
        ]
```

(full implementation 

The issue is a bit more clear before the refactor in 1fc45eceb776546c6ea028c81f37ca30cc6a8710:

```
    for pre_subj in subject_prefixes:
        for pre_obj in object_prefixes:
            for rel in relations:
...
                if df is not None:
                    dfs = df[
                        (df[SUBJECT_ID].str.startswith(pre_subj + ":"))
                        & (df[PREDICATE_ID] == rel)
                        & (df[OBJECT_ID].str.startswith(pre_obj + ":"))
                    ]
```

The issue is: the entire MappingSetDataFrame is iterated over for the product of `len(subject_prefix) * len(object_prefix) * len(relations)`. Each iteration involves creating a new string and calling `.startswith` twice. While these are small operations, they add up in the end. (It takes ~10 minutes in the case I'm looking at). There might be a big performance gain from only going through the MappingSetDataFrame once.

A quickly written attempt that I can formalize as a PR if the approach looks OK:

```python
    @dataclass
    class SSSOMTriple:
        subject_prefix: str
        object_prefix: str
        relation: str

        def __post_init__(self):
            relation_prefix, relation_id = self.relation.split(":")
            self.relation_prefix = relation_prefix
            self.relation_id = relation_id

        def as_identifier(self):
            return f"{self.subject_prefix.lower()}_{self.relation_id.lower()}_{object_prefix.lower()}"

    splits_by_triple: dict[SSSOMTriple, list[dict]] = {}

    # Still iterate through the product of SxPxO initially to pre-populate a dict
    for subject_prefix, object_prefix, relation in itt.product(
        subject_prefixes, object_prefixes, relations
    ):
        triple = SSSOMTriple(subject_prefix, object_prefix, relation)
        split = triple.as_identifier()
        if subject_prefix not in msdf.converter.bimap:
            logging.warning(f"{split} - missing subject prefix - {subject_prefix}")
            continue
        if object_prefix not in msdf.converter.bimap:
            logging.warning(f"{split} - missing object prefix - {object_prefix}")
            continue
        splits_by_triple[triple] = []

    # Iterate through `msdf.df` once, building up a list of matches in the pre-populated dict
    for entry in df.itertuples(index=False):
        _entry = entry._asdict()
        subject_prefix = entry[SUBJECT_ID].split(":")[0]
        object_prefix = entry[OBJECT_ID].split(":")[0]
        relation = entry[PREDICATE_ID]
        triple = SSSOMTriple(subject_prefix, object_prefix, relation)
        if triple in splits_by_triple:
            splits_by_triple[triple].append(_entry)

    # Iterate through the prepopulated dict
    for triple, values in splits_by_triple.items():
        split = triple.as_identifier()
        if len(values) == 0:
            logging.debug(f"{split} - No matches (0 matches found)")
            continue
        subconverter = msdf.converter.get_subconverter(
            [triple.subject_prefix, triple.object_prefix, triple.relation_prefix]
        )
        split_to_msdf[split] = from_sssom_dataframe(
            pd.DataFrame(values), prefix_map=dict(subconverter.bimap), meta=meta
        )

    return split_to_msdf
```

This would drastically reduce the performance of the algorithm as the size of `msdf.df` increases (if there are non-trival amounts of subject_prefix/object_prefix/relation iterables). In my case, it went from ~10m to nearly instant with a similar approach. The only added complexity is constructing the `SSSOMTriple` object for every row in the dataframe and every individual of the SxPxO product, but this is not computationally complex-- it should not add a performance penalty in any case.

ptgolden:
There was too much iteration going on in `parsers.split_dataframe_by_prefix`-- for every combination of prefixes\*relations\*objects, the entire input dataframe was being iterated over. This was a lot of overhead, when it only needs to be gone through once.

This is basically the [same approach used in sssom-java](https://github.com/gouttegd/sssom-java/blob/main/cli/src/main/java/org/incenp/obofoundry/sssom/cli/SimpleCLI.java#L646-L670), but there is some more work being done to keep logging consistent with the previous approach. If there were no logging, the dataclass included here (which reduces repetition), would not be necessary.

In testing splitting `mondo.sssom.tsv` (a 25mb file), the time to split went from ~5 minutes to ~50 seconds. Nearly all of the time spent is on calling `parsers.from_sssom_dataframe`, which is slow but inevitable.

Fixes #607

### DIFF
--- a/src/sssom/parsers.py
+++ b/src/sssom/parsers.py
@@ -9,7 +9,19 @@ import os.path
 import typing
 from collections import ChainMap, Counter
 from pathlib import Path
-from typing import Any, Callable, Dict, Iterable, List, NamedTuple, Optional, TextIO, Tuple, Union, cast
+from typing import (
+    Any,
+    Callable,
+    Dict,
+    Iterable,
+    List,
+    NamedTuple,
+    Optional,
+    TextIO,
+    Tuple,
+    Union,
+    cast,
+)
 from xml.dom import Node, minidom
 from xml.dom.minidom import Document
 
@@ -979,16 +991,20 @@ def split_dataframe(
 
 class SSSOMSplitGroup(NamedTuple):
     """A tuple to identify the boundary where a mapping set should be split."""
+
     subject_prefix: str
     object_prefix: str
     relation_curie: ReferenceTuple
 
     def as_identifier(self):
-        return "_".join([
-            self.subject_prefix.lower(),
-            self.relation_curie.identifier.lower(),
-            self.object_prefix.lower(),
-        ])
+        return "_".join(
+            [
+                self.subject_prefix.lower(),
+                self.relation_curie.identifier.lower(),
+                self.object_prefix.lower(),
+            ]
+        )
+
 
 def split_dataframe_by_prefix(
     msdf: MappingSetDataFrame,
--- a/src/sssom/parsers.py
+++ b/src/sssom/parsers.py
@@ -8,8 +8,9 @@ import logging as _logging
 import os.path
 import typing
 from collections import ChainMap, Counter
+from dataclasses import dataclass
 from pathlib import Path
-from typing import Any, Callable, Dict, Iterable, List, Optional, TextIO, Tuple, Union, cast
+from typing import Any, Callable, Dict, Iterable, List, NamedTuple, Optional, TextIO, Tuple, Union, cast
 from xml.dom import Node, minidom
 from xml.dom.minidom import Document
 
@@ -977,6 +978,27 @@ def split_dataframe(
     )
 
 
+@dataclass(frozen=True, eq=True)
+class SSSOMTriple:
+    subject_prefix: str
+    object_prefix: str
+    relation: str
+
+    def __post_init__(self):
+        relation_prefix, relation_id = self.relation.split(":")
+
+    @property
+    def relation_prefix(self):
+        return self.relation.split(":")[0]
+
+    @property
+    def relation_id(self):
+        return self.relation.split(":")[1]
+
+    def as_identifier(self):
+        return f"{self.subject_prefix.lower()}_{self.relation_id.lower()}_{self.object_prefix.lower()}"
+
+
 def split_dataframe_by_prefix(
     msdf: MappingSetDataFrame,
     subject_prefixes: Iterable[str],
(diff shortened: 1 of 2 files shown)


cthoyt:
@ptgolden that's an incredible speedup! Nice work! Do you have a link to this mondo.sssom.tsv for my own testing purposes?

src/sssom/parsers.py
```diff
@@ -977,6 +978,27 @@ def split_dataframe(
     )
 
 
+@dataclass(frozen=True, eq=True)
+class SSSOMSplitTriple:
```
cthoyt, line 982: why not use a named tuple here?

src/sssom/parsers.py
```diff
@@ -977,6 +978,27 @@ def split_dataframe(
...
+    subject_prefix: str
+    object_prefix: str
+    relation: str
+
+    def __post_init__(self):
+        relation_prefix, relation_id = self.relation.split(":")
```
cthoyt, line 988: use the `curies` package for CURIE parsing! Within the MSDF, you can access the converter directly with `msdf.converter` and use `msdf.converter.parse_curie`

tests/test_parsers.py
```diff
@@ -273,6 +274,21 @@ def test_parse_sssom_json(self):
             f"{self.json_file} has the wrong number of mappings.",
         )
 
+    def test_split_msdf(self):
```
matentzn, line 277: Thank you for the tests!!!

src/sssom/parsers.py
ptgolden: Definitely could do! Although that would preclude having methods on the class. (Those methods were for convenience, since they're used twice).

ptgolden:
> @ptgolden that's an incredible speedup! Nice work! Do you have a link to this mondo.sssom.tsv for my own testing purposes?

Just replied at https://github.com/mapping-commons/sssom-py/pull/608#issuecomment-3258780956

matentzn (commented):
Thank you so much!

My python reading is a bit rusty.

1. I checked the tests and they make sense
2. The code is well written. I do not know whether OO (i.e. `SSSOMSplitTriple`) is the way things are done still, but I dont see a problem.

@cthoyt do you have any major objections (since this is, or might be, an interim solution and it will help me shave A LOT of time of my existing pipelines).

cthoyt:
@matentzn yes there are major objections to the way this is written, I will gently guide @ptgolden though writing nicer python code before making a merge

src/sssom/parsers.py
ptgolden: Yes, true, a subclass of NamedTuple could add those methods. I am happy to change that- though what is the advantage over a hashable dataclass (i.e. `frozen=True`) at that point?

tests/test_parsers.py
ptgolden: I realized that the test for `split` in the CLI only checked if the command ran without error, without checking any output :face_with_spiral_eyes:

src/sssom/parsers.py
ptgolden: Oh wow, subclassing NamedTuple is a good deal more ergonomic than it used to be. Change incoming!

tests/test_parsers.py
cthoyt: Don’t get me on a rant about the cli testing here…

ptgolden:
Thanks, @cthoyt. I've refactored a bit and I think it's a good deal better now. Let me know if there's anything else you can see.

cthoyt (changes requested):
Try and get the tests to pass, then I can take another look and we can get to refactoring.

ptgolden:
@cthoyt d439b60 actually introduces a change in behavior. Previously, a mapping would only be added to the split if its subject prefix, object prefix, and relation were all in the iterables passed as arguments to the function. This was the purpose of building up a dict first (rather than using defaultdict)-- to have a known set of groups which were meant to be populated. After this change, all mappings are added to the split regardless of whether that's the case.

Note that this won't change behavior in the case of `split_dataframe` (since all of the object prefixes, subject prefixes, and relations are passed to `split_dataframe_by_prefix` there), but it would affect behavior when called with different arguments.

ptgolden:
Noting: whether that makes sense as an API design-- I don't know! But that's how the function was implemented when I started refactoring.

cthoyt:
I realized that almost immediately and reverted it. Should be back to what you had

though this can definitely use refactoring because it is hard to understand

ptgolden:
Understood. Most of the complexity derives from having to keep track of state for logging purposes. Is this logging even necessary?:

```
        if subject_prefix not in msdf.converter.bimap:
            logging.warning(f"{split_id} - missing subject prefix - {subject_prefix}")
            continue
        if object_prefix not in msdf.converter.bimap:
            logging.warning(f"{split_id} - missing object prefix - {object_prefix}")
            continue
```

This will never be triggered if someone calls `split_dataframe`, only if they call `split_dataframe_by_prefix` with other arguments, in which case they probably explicitly want to skip certain prefixes.

ptgolden:
I'm closing this one in favor of #611.

PR #610, opened 2025-09-06
Add reusable test for split dataframe function

cthoyt:


ptgolden:
Thank you! This was the kind of test I wanted to write but I instead matched the style of the existing ones. (The existing ones are... interesting!)

cthoyt:
@ptgolden that's a... nice way of putting it ;)

PR #611, opened 2025-09-07
Optimize split_dataframe_by_prefix

ptgolden:
Replaces #609. I started from scratch with a total rewrite rather than adding to the commit history of that one. The output here is identical, but it's hopefully easier to follow.

The point is the same: when splitting a MappingSetDataFrame by `l` subject prefixes, `m` object prefixes, and `n` relations, iterate over `msdf.df` once rather than `l * m * n` times.


src/sssom/parsers.py
```diff
@@ -996,34 +1017,53 @@ def split_dataframe_by_prefix(
...
+            subject_prefixes, relations, object_prefixes
+        )
+    ]
+
+    for mapping in msdf.df.itertuples(index=False):
+        group = SSSOMSplitGroup(
+            parse_curie(getattr(mapping, SUBJECT_ID), strict=True).prefix,
+            parse_curie(getattr(mapping, OBJECT_ID), strict=True).prefix,
+            parse_curie(getattr(mapping, PREDICATE_ID), strict=True),
+        )
```
cthoyt, line 1041:
```suggestion
    for subject_id, predicate_id, object_id in df[[SUBJECT_ID, PREDICATE_ID, OBJECT_ID]].values:
        group = SSSOMSplitGroup(
            parse_curie(subject_id, strict=True).prefix,
            parse_curie(object_id, strict=True).prefix,
            parse_curie(predicate_id, strict=True),
        )
```

cthoyt:
@ptgolden I just added a test case to make sure we don't break backwards compatibility with handling of CURIEs that aren't registered properly in the internal converter. We'll also want to address this

src/sssom/parsers.py
ptgolden: Will change. I used `itertuples()` on purpose because I thought it would save some memory/processing, but after profiling- it doesn't!

cthoyt:
I just merged all of the code together so we could get a better look at the time:

```
$ wget https://github.com/monarch-initiative/mondo/raw/refs/heads/master/src/ontology/mappings/mondo.sssom.tsv
$ mkdir results
$ time sssom split mondo.sssom.tsv -d results
________________________________________________________
Executed in   14.16 secs    fish           external
   usr time   14.16 secs    0.19 millis   14.16 secs
   sys time    3.49 secs    2.25 millis    3.48 secs
$ time sssom split mondo.sssom.tsv -d results --method patrick
________________________________________________________
Executed in   12.76 secs    fish           external
   usr time   12.80 secs    0.20 millis   12.80 secs
   sys time    3.44 secs    2.35 millis    3.43 secs
$ time sssom split mondo.sssom.tsv -d results --method disjoint-indexes
________________________________________________________
Executed in   12.74 secs    fish           external
   usr time   12.68 secs    0.12 millis   12.68 secs
   sys time    3.59 secs    1.49 millis    3.58 secs
$ time sssom split mondo.sssom.tsv -d results --method dense-indexes
________________________________________________________
Executed in   12.62 secs    fish           external
   usr time   12.64 secs    0.19 millis   12.64 secs
   sys time    3.51 secs    2.29 millis    3.50 secs
```

It might also be worth doing a memory profiling. If they're all comparable, then I would definitely opt for more idiomatic, reusable code rather than using a new custom implementation. @ptgolden can you run the same profiles I did above and let me know what it looks like?

Issue #612, opened 2025-09-08
Achieve strict static type checking


cthoyt:
Add `--strict` to the mypy configuration in tox. This can be done incrementally, by adding more comprehensive static typing throughout the package

PR #613, opened 2025-09-08
Add more type checking

cthoyt:
On the way towards achieving strict static type checking #612

This PR does not change any functionality, it only adds type annotations.

### DIFF
--- a/src/sssom/cli.py
+++ b/src/sssom/cli.py
@@ -123,7 +123,7 @@ predicate_filter_option = click.option(
 @click.option("-v", "--verbose", count=True)
 @click.option("-q", "--quiet")
 @click.version_option(__version__)
-def main(verbose: int, quiet: bool):
+def main(verbose: int, quiet: bool) -> None:
     """Run the SSSOM CLI."""
     logger = _logging.getLogger()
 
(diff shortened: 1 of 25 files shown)


src/sssom/cli.py
```diff
@@ -660,14 +661,20 @@ def sort(input: str, output: TextIO, by_columns: bool, by_rows: bool):
 #     write_table(msdf=filtered_msdf, file=output)
 
 
-def dynamically_generate_sssom_options(options) -> Callable[[Any], Any]:
+P = ParamSpec("P")
+T = TypeVar("T")
+
+
+def dynamically_generate_sssom_options(
+    options: List[str],
```
gouttegd, line 669: Minor nitpick: Some type hints use `List[...]` (as on this line), whereas some other type hints use `list[...]`.

My understanding is that the capitalized version is only required for Python 3.8 and earlier, while Python 3.9+ can accommodate both forms (capitalized or not).

Since we are already no longer supporting Python 3.8, maybe we could uniformly use the lowercase forms, instead of a mix of both forms?

gouttegd (approved):
Since we use type hints, we should use them fully, so no objection to adding them where they are missing.

(I do question a bit the usefulness of type hints in test methods, but I don’t mind. If anything, seeing type hints in those methods merely reinforces the idea that type hints should be used everywhere, which is good.)

PR #620, opened 2025-09-18
Update minimum version to Python 3.10

cthoyt:
Python 3.9 is end of life at the end of October. This PR gets a bit ahead on it, because I keep running into type annotation backwards compatibility issues and wasting Damien's valuable review time


.github/workflows/qc.yml
```diff
@@ -12,7 +12,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, windows-latest]
-        python-version: [ "3.9", "3.10", "3.11", "3.12" ]
+        python-version: [ "3.10", "3.11", "3.12" ]
```
gouttegd, line 15: Shouldn’t we use that opportunity to add `3.13`?

gouttegd (approved):
Obviously no objection, just wondering if we should start testing with 3.13 instead?

.github/workflows/qc.yml
gouttegd: Strange that the test should fail under 3.13, though. The test suite runs fine under 3.13 on my machine.

cthoyt: I don't want to start working on fixing (removing) poetry in this PR.

gouttegd: > fixing (removing) poetry

🤣 Fair enough!

PR #624, opened 2025-10-04
Fix support for literal mappings.

gouttegd:
Since the LinkML-generated `Mapping` class does not, in itself, validates the post-conditions specified in the schema, it cannot validate that a Mapping object is created with either a `subject_id` or (if `subject_type` is `rdfs literal`) a `subject_label` (likewise for the object side).

Instead, that validation is performed by custom code executed as part of the parser (after the parsing proper has been done, but before return a `MappingSetDataFrame` to the caller).

However, the present validation is bogus because the check on the value of `subject_type` and `object_type` is trying to compare the value of the slot to a literal string, instead of a EntityTypeEnum object. We fix that comparison here.

### DIFF
--- a/src/sssom/parsers.py
+++ b/src/sssom/parsers.py
@@ -42,7 +42,7 @@ from linkml_runtime.loaders.json_loader import JSONLoader
 from linkml_runtime.loaders.rdflib_loader import RDFLibLoader
 from pandas.errors import EmptyDataError
 from rdflib import Graph
-from sssom_schema import Mapping, MappingSet
+from sssom_schema import EntityTypeEnum, Mapping, MappingSet
 from typing_extensions import Literal, TypeAlias
 
 from sssom.constants import (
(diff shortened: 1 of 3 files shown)


src/sssom/parsers.py
```diff
@@ -1187,12 +1187,12 @@ def _ensure_valid_mapping_from_dict(mdict: Dict[str, Any]) -> Optional[Mapping]:
 
     try:
         m = Mapping(**mdict)
-        if m.subject_type == "rdfs literal":
+        if m.subject_type == EntityTypeEnum(EntityTypeEnum["rdfs literal"]):
```
gouttegd, line 1190: Note: As strange as it may look like, this is the _normal_ way of comparing LinkML-defined enum values.

A comparison like this:

```python
m.subject_type == EntityTypeEnum["rdfs literal"]
```

would _not_ work because `EntityTypeEnum["rdfs literal"]` is _not_ the actual enum value, it is the LinkML _PermissibleValue_ object that represents the enum value (that’s not the same thing!). We must give that value to the `EntityTypeEnum` constructor to obtain an actual enum value that we can compare to the value of the slot.

See [here](https://github.com/linkml/linkml/issues/723) for details, and [here](https://github.com/linkml/linkml/issues/1203) for a request for an easier way of working with LinkML enums).

(Of course in our case this is make even weirder by the fact that our enum values contain space characters, and LinkML does not allow to distinguish between the actual value to be used in data and the symbol used in Python code.)

PR #625, opened 2025-10-08
Use latest pre-release of the SSSOM schema.

gouttegd:
This PR updates the `sssom_schema` dependency to use the latest 1.1.0 pre-release, so that we can start using the new slots and enum values in the upcoming version of the specification and ensure that SSSOM-Py is ready to work with them.


matentzn (approved):
Very awesome, thank you!

PR #632, opened 2025-10-20
Use custom code for RDF (de)serialisation

### LINKED ISSUE #628: Support for “new” RDF serialisation
Now that the [normalized form of the RDF serialisation](https://github.com/mapping-commons/sssom/pull/469) is about to get merged, we should think about how it is going to be supported by SSSOM-Py.

The normalized form is purposefully very close to the form that SSSOM-Py has been producing and reading since the beginning, but it does deviates from it on some aspects – it deviates enough for the current SSSOM/RDF parser in SSSOM-Py to _reject_ a file compliant with the normalized form.

Both the current SSSOM/RDF parser and the current SSSOM/RDF writer in SSSOM-Py are heavily reliant on the LinkML runtime: the parser is mostly a thin wrapper around LinkML’s `RDFLibLoader` and the writer is mostly a thin wrapper around LinkML’s `rdflib_dumper`.

Unfortunately, that approach is only valid for a RDF serialisation that is exactly conform to LinkML’s [RDF translation rules](https://linkml.io/linkml-model/latest/docs/specification/06mapping/#translation-to-rdf), which the new normalized form of SSSOM/RDF is _not_.

Possible new approaches to support the form:

**(A)** Forget about LinkML’s RDF-related methods and write entirely custom code to convert to and from RDF, using SSSOM’s own translation rules.

In my (not-so-)humble opinion the best solution, simultaneously the cleanest and the most flexible, but possibly the one that will require the most work (though it’s unclear to me that the other approaches below are actually any better on that front).

**(B)** Keep using the LinkML runtime, with appropriate pre- and post-processing steps. That is:

* when converting _from_ RDF, add a pre-processing step to transform the graph into the form expected by the LinkML runtime, then feed the transformed graph to the runtime;
* when converting _to_ RDF, use the LinkML runtime to produce a first graph, then add a post-processing step to transform that graph into the new normalized form.

_Possibly_ quicker to code, but that’s a lot of completely avoidable complexity.

**(C)** Going through JSON-LD, as suggested in #592 (at least for the conversion _to_ RDF).

The problem here is the JSON-LD serialisation of SSSOM is still not formally defined (which incidentally is also the case of the non-LD JSON serialisation). Now I think it would make much sense for SSSOM/JSON-LD to be defined in such a way that is exactly equivalent to SSSOM/RDF, but we have not decided that yet.

gouttegd:
This PR implements option ”A” mentioned in #628. That is, it replaces the LinkML-based code for serialising/deserialising to/from RDF by a completely custom code that implements the new SSSOM/RDF spec exactly as we want it.

All the new code is contained within the new `sssom.rdf_internal` module, which as the name implies is not really supposed to be used outside of SSSOM-Py (though of course people can still do whatever they want!). The public interface to read from RDF is still the `from_sssom_rdf` method in `sssom.parsers`, and the public interface to write to RDF is still the `to_rdf_graph` method in `sssom.writers`. Both methods are rewritten to use the `MappingSetRDFConverter` class of the `sssom.rdf_internal` module, instead of methods from the LinkML runtime.

closes #628

### DIFF
--- a/.github/workflows/qc.yml
+++ b/.github/workflows/qc.yml
@@ -60,4 +60,5 @@ jobs:
         run: |
           poetry add "pandas<2.0.0"
           poetry add "numpy<1.24"
+          poetry add "scipy<1.16.3"
           poetry run tox -e py


gouttegd:
Thanks!

gouttegd:
Support for extension slots is still missing, but as the rest of SSSOM-Py does not support them either, this is not really important – and I don’t want to start working on extension slots until there has been a discussion on _how_ to support them in all of SSSOM-Py, not just in the RDF parser/writer.

PR #635, opened 2025-10-29
Use custom code for RDF (de)serialisation

### LINKED ISSUE #628: Support for “new” RDF serialisation

gouttegd:
This PR implements option ”A” mentioned in #628. That is, it replaces the LinkML-based code for serialising/deserialising to/from RDF by a completely custom code that implements the new SSSOM/RDF spec exactly as we want it.

All the new code is contained within the new `sssom.rdf_internal` module, which as the name implies is not really supposed to be used outside of SSSOM-Py (though of course people can still do whatever they want!). The public interface to read from RDF is still the `from_sssom_rdf` method in `sssom.parsers`, and the public interface to write to RDF is still the `to_rdf_graph` method in `sssom.writers`. Both methods are rewritten to use the MappingSetRDFConverter class of the `sssom.rdf_internal` module, instead of methods from the LinkML runtime.

closes #628

### DIFF
--- a/src/sssom/rdf_internal.py
+++ b/src/sssom/rdf_internal.py
@@ -446,7 +446,7 @@ class ObjectConverter:
     #
 
     def dict_from_rdf(
-        self, g: Graph, subject: Node, dest: Optional[Dict[str, Any]] = None
+        self, graph: Graph, subject: Node, dest: Optional[Dict[str, Any]] = None
     ) -> Dict[str, Any]:
         """Parse a SSSOM object from a RDF node.
 
@@ -454,7 +454,7 @@ class ObjectConverter:
         constructs a dictionary from all the triples where the node is
         the subject.
 
-        :param g: The graph to parse the object from.
+        :param graph: The graph to parse the object from.
         :param subject: The root node of the object to extract.
         :param dest: The dictionary into which the parsed object will be
             stored, if specified. Client code may use this argument to
@@ -470,16 +470,16 @@ class ObjectConverter:
         """
         if dest is None:
             dest = {}
-        self._init_dict_from_rdf(g, subject, dest)
+        self._init_dict_from_rdf(graph, subject, dest)
 
-        for pred, obj in g.predicate_objects(subject):
+        for pred, obj in graph.predicate_objects(subject):
             pred = cast(URIRef, pred)
             if pred == RDF.type:
                 if obj != self.object_uri:
                     raise ValueError(f"Invalid type {obj} for a {self.name} object")
                 continue
 
-            if self._process_triple(g, cast(Triple, [subject, pred, obj]), dest):
+            if self._process_triple(graph, cast(Triple, [subject, pred, obj]), dest):
                 continue
 
             slot = self.slots_by_uri.get(pred)
@@ -499,7 +499,7 @@ class ObjectConverter:
 
     # Helper methods for conversion from RDF
 
-    def _init_dict_from_rdf(self, g: Graph, subject: Node, dest: Dict[str, Any]) -> None:
+    def _init_dict_from_rdf(self, graph: Graph, subject: Node, dest: Dict[str, Any]) -> None:
         """Initialize a dictionary representing a parsed object.
 
         This method is used to check that the subject node is of a
@@ -508,7 +508,7 @@ class ObjectConverter:
         any additional specific operations at the beginning of the
         conversion.
 
-        :param g: The graph to parse the object from.
+        :param graph: The graph to parse the object from.
         :param subject: The root node of the object to extract.
         :param dest: The dictionary in which to store the parsed object.
 
@@ -518,14 +518,14 @@ class ObjectConverter:
         if not isinstance(subject, BNode):
             raise ValueError(f"Invalid node type for a {self.name} object")
 
-    def _process_triple(self, g: Graph, triple: Triple, dest: Dict[str, Any]) -> bool:
+    def _process_triple(self, graph: Graph, triple: Triple, dest: Dict[str, Any]) -> bool:
         """Process an individual triple associated with a SSSOM object.
 
         This method is intended to allow subclasses to implement custom
         processing of some RDF triples, before letting the triple being
         handled by the common logic in the `dict_from_rdf` method.
 
-        :param g: The graph to parse the object from.
+        :param graph: The graph to parse the object from.
         :param triple: The triple to process.
         :param dest: The dictionary in which to store the parsed object.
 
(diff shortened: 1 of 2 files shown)


src/sssom/rdf_internal.py
```diff
@@ -0,0 +1,992 @@
...
+            use for CURIE expansion/contraction.
+        """
+        super().__init__(primary_type=RDFS.Resource, allowed_types=[XSD.string])
+        self.ccp = ccp
+
+    @override
+    def from_rdf(self, obj: Node) -> str:
+        """Convert a RDF node into an entity reference value."""
+        value = super().from_rdf(obj)
+        return self.ccp().compress(value, passthrough=True)
```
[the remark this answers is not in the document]
gouttegd: I principle I agree, and in fact I initially did not use `passthrough=True`, but then this caused the test suite to fail, because some of the very sets that we use in our own test suite apparently contain poorly formatted data – which was accepted without errors or warnings by the current RDF parser and writer.

I think the new code should be, at least initially, as tolerant as the old one. If the new code refuses a set that the old one let pass, I’d consider that a bug of the new code (and this is why I had to use `passthrough=True`).

I wouldn’t be opposed to making the code more strict, but that would belong to another PR IMHO.

src/sssom/rdf_internal.py
```diff
@@ -0,0 +1,992 @@
...
+            return EntityReferenceValueConverter(ccp)
+        elif ctor is not None:
+            return ctor()
+        else:
+            # This should only happen if a brand new type of slot has
+            # been introduced in the SSSOM schema
+            raise NotImplementedError(f"Range {range_name} is not supported")
+
+
+class ObjectConverter(object):
```
cthoyt: inheriting from `object` isn't needed since Python 2

```suggestion
class ObjectConverter:
```

src/sssom/rdf_internal.py
```diff
@@ -0,0 +1,992 @@
...
+        for slot in self.schema.view.class_induced_slots(class_name):
+            self.slots_by_name[slot.name] = slot
+            self.slots_by_uri[self._get_slot_uri(slot)] = slot
+            ranges.append(slot.range)
+
+        # ... and the scalar converters table
+        self.value_converters = {}
+        factory = ValueConverterFactory()
+        for rng in set(ranges):
+            vc = factory.create(rng, self.schema.view, ccp)
```
cthoyt: idk why the type checker isn't catching this, but `rng` is not a string as the create() function is expecting

src/sssom/rdf_internal.py
```diff
@@ -0,0 +1,992 @@
...
+
+        :returns: True if the value is empty.
+        """
+        return value is None or (hasattr(value, "__len__") and len(value) == 0)
+
+    def _init_dict_to_rdf(self, g: Graph, obj: DictOrSeries) -> Node:
+        """Create the root node representing a SSSOM object.
+
+        Subclasses should override this method to customize the way
+        their SSSOM object is represented in RDF. The default bhaviour
```
matentzn: bhaviour

src/sssom/rdf_internal.py
```diff
@@ -0,0 +1,992 @@
...
+
+        :param g: The graph to export to.
+        :param subject: The node representing the exported object.
+        :param name: The name of the extension slot.
+        :param value: The value of the extension slot.
+
+        :returns: True if `name` is the name of a valid extension,
+            otherwise False.
+        """
+        # FIXME: Not implemented yet, ignore all for now
```
matentzn: I assume you want a sep. PR for this. IN the meantime, maybe logging.warning()?

matentzn (approved):
Sorry it took so long to look at this. 

I have reviewed mostly everything _apart from_ src/sssom/rdf_internal.py (I scanned it from an architectural viewpoint). This all looks clean, well designed and performant!

gouttegd:
> I've made a PR onto this PR gouttegd#3 that improves the control flow

Commenting here to keep the discussion in a single place.

Your PR “improves” the control flow by making the `MappingSetRDFConverter` class in effect not reusable (unless for sets that all use the same prefix map), thereby negating one of the main interests of that class.

The class is designed so that you can do something like this:

```python
from rdflib import Graph
from sssom.rdf_internal import MappingSetRDFConverter

# Creating a converter object once and for all
converter = MappingSetRDFConverter()

# Extracting a set from a graph
msdf1 = converter.msdf_from_rdf(Graph().parse("my-set-1.ttl"))

# Extracting another set from another graph
msdf2 = converter.msdf_from_rdf(Graph().parse("my-set-2.ttl"))

# Extracting yet another set from yet another graph
msdf3 = converter.msdf_from_rdf(Graph().parse("my-set-3.ttl"))

# and so on...
```

and likewise for RDF export (`msdf_to_rdf`).

The “improved” version requires instead to create a whole new converter for every single set to extract:

```python
g1 = Graph().parse("my-set-1.ttl")
msdf1 = MappingSetRDFConverter(Converter.from_rdflib(g1)).msdf_from_rdf(g1)

g2 = Graph().parse("my-set-2.ttl")
msdf2 = MappingSetRDFConverter(Converter.from_rdflib(g2)).msdf_from_rdf(g2)

# and so on...
```

And then there is no longer any real point in having a `MappingSetRDFConverter` class to begin with...

At the _very least_, such a design would absolutely need a clear warning in the documentation of the `MappingSetRDFConverter` constructor that any new instance can be used to (de)serialize several mapping sets _if and only if_ all the sets share the same prefix map.

gouttegd:
> from your perspective this is ready to merge right?

Assuming there is agreement on the overall design, yes.

If there is _no_ agreement on the design: I personally think supporting the RDF serialisation as described in the spec is more important than the exact design used to achieve that aim. I am not wed to my initial design and could consider changing it, but _not_ for the design proposed in https://github.com/gouttegd/sssom-py/pull/3, at least not in its current state.

Issue #657, opened 2026-04-16
Support record hashing and "mapping sameness identifier"


gouttegd:
SSSOM-Py should include functions to compute, for any given mapping:

* the “SSSOM standard hash”, as [defined by the SSSOM specification](https://mapping-commons.github.io/sssom/spec-support-hashing/);
* and while we’re at it the “Mapping Sameness Identifier” as defined by the [proposed standard of the same name](https://ts4nfdi.github.io/mapping-sameness-identifier/).

PR #658, opened 2026-04-17
[Demo] Implement mapping sameness identifier

cthoyt:
Part of #657

This demo PR shows how the mapping sameness identifier could be implemented.

However, it needs a design. How would someone using sssom-py use this, considering that the primary data structure is a dataframe, whose columns correspond to SSSOM fields?

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -19,7 +19,7 @@ requires-python = ">=3.10,<4.0.0"
 
 dependencies = [
     "click>=8.1.6",
-    "curies>=0.10.23",
+    "curies>=0.13.4",
     "linkml-runtime>=1.10.0",
     "linkml>=1.10.0",
     "packaging>=20.0",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -19,7 +19,7 @@ requires-python = ">=3.10,<4.0.0"
 
 dependencies = [
     "click>=8.1.6",
-    "curies>=0.13.4",
+    "curies>=0.13.5",
     "linkml-runtime>=1.10.0",
     "linkml>=1.10.0",
     "packaging>=20.0",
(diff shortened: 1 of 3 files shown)


gouttegd:
> However, it needs a design. How would someone using sssom-py use this, considering that the primary data structure is a dataframe, whose columns correspond to SSSOM fields?

Well, your proposed design (i.e. a method on the `MappingSetDataFrame` class which returns a list of hashes) doesn’t look too bad to me. It’s consistent with the idea that the `MappingSetDataFrame` is, in SSSOM-Py, the primary way of manipulating SSSOM sets.

For convenience, maybe the method could take an optional parameter to filter the rows of the dataframe, so that if the caller only wants the hash for one particular record (or a particular group of records), the method would only return the hash(es) for the selected record(s).

Something like:

```python
def get_mapping_sameness_identifiers(self, selector=None) -> list[str]:
    """Get mapping sameness identifiers for all records (selector == None) or all selected records."""
    if selector is not None:
        df = self.df.loc[selector]
    else:
        df = self.df
    expand = partial(self.converter.expand, strict=True)
    return [
        encode_uri_triple(
            (expand(subject_curie), expand(predicate_curie), expand(object_curie)),
            negate=predicate_modifier == "Not",
        )
        for subject_curie, predicate_curie, object_curie, predicate_modifier in df[
            ["subject_id", "predicate_id", "object_id", "predicate_modifier"]
        ].values
    ]
```

If you want _all_ the hashes, just call `msdf.get_mapping_sameness_identifiers()`.

If you only want the hash for the second record, call `msdf.get_mapping_sameness_identifiers(selector=[2])` (beware that the method will still return a list, even if it contains a single hash).

If you only want the hash for the records whose subject ID is `EX:123`, call `msdf.get_mapping_sameness_identifiers(selector=msdf.df["subject_id"] == "EX:123")`.

Etc.

PR #660, opened 2026-05-04
Preserve metadata-only prefixes when splitting

### LINKED ISSUE #573: `write_table()` ignores `curie_map` in `msdf.metadata`
## Overview
I'm creating a `MappingSetDataFrame` (`msdf`) and then saving it:

```python
    msdf = MappingSetDataFrame(matches_df, metadata={
        'curie_map': {
            'rdfs': 'http://www.w3.org/2000/01/rdf-schema#'
        },
    })
    with open(outpath, 'w') as f:
        write_table(msdf, f)
```

When I inspect `msdf`, I see that it does have my correct `curie_map` set inside the `metadata` property.

However, when I `write_table()` and inspect the resulting TSV, the `curie_map` is empty within the metadata comments at the top of the file:

```
# curie_map: {}
```

matentzn:
Fixes #573 

@cthoyt problem was actually ONLY in the split dataframe by prefix method. Would appreciate review.

### DIFF
--- a/src/sssom/parsers.py
+++ b/src/sssom/parsers.py
@@ -86,6 +86,7 @@ from .util import (
     ExtensionLiteral,
     MappingSetDataFrame,
     get_file_extension,
+    get_prefixes_used_in_metadata,
     is_multivalued_slot,
     propagate_condensed_slots,
     raise_for_bad_path,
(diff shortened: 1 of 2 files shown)


src/sssom/parsers.py
```diff
@@ -1075,10 +1076,11 @@ def split_dataframe_by_prefix(
         object_prefixes=object_prefixes,
         method=method,
     )
+    metadata_prefixes = get_prefixes_used_in_metadata(msdf.metadata) & msdf.converter.get_prefixes()
```
cthoyt, line 1081: can you comment that the reason for `& msdf.converter.get_prefixes()` is to subset to only _valid_ prefixes? this isn't obvious on a first glance

cthoyt (dismissed):
this looks good and the test seems to make sense.

you could consider extending the test with the common workflow that writes the split dataframes (however this is done in mondo) to make sure that the right metadata comes along for the ride

PR #662, opened 2026-05-13
Add option to print prettier validation reports

matentzn:
Replaces #387

@cthoyt I think the pretty rendering makes more sense because the default linkml is to bulky for normal humans to read..

- Add `format_report()`: one line per unique JSON path, drops the verbose
  `instance` blob, collapses paired pattern/anyOf errors from LinkML.
- Stop validators from logging as a side effect; extract `_raise_if_errors`
  so callers choose their own rendering. `print_linkml_report` retained
  for log-style output.
- Respect severity: only FATAL/ERROR cause raises and non-zero CLI exit.
- CLI: add `--report-format/-r {pretty,raw}`, default `pretty`.

### DIFF
--- a/src/sssom/cli.py
+++ b/src/sssom/cli.py
@@ -15,6 +15,7 @@ import curies
 import pandas as pd
 import yaml
 from curies import Converter
+from linkml.validator.report import Severity
 from rdflib import Graph, URIRef
 from typing_extensions import ParamSpec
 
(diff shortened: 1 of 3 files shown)


cthoyt:
I’m afk for a few days but I think the idea in general is good. If this could be reusable you could consider upstreamjng for other linkml users

PR #664, opened 2026-05-23
Enable running tests when pansql isn't installed

cthoyt:
If pansql isn't installed, tests would fail. This adds some guard rails to skip the tests for pansql if it's not installed, since it's an optional dependency

### DIFF
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -1,5 +1,6 @@
 """Tests for the command line interface."""
 
+import importlib.util
 import os
 import subprocess  # noqa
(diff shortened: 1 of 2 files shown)


matentzn (approved):
Fair enough :) I am assuming that for sssom-py CI / CD pansql is installed (else it would have failed all along)

cthoyt:
Correct, all optional dependencies are installed during normal CI/CD but this surprisingly fails if you just do a regular install and try and run pytest (I ran into this when testing the sssom improvements)

PR #668, opened 2026-06-21
Update uv.lock, pin some vulnerable dependencies and remove pandas compatibility code

matentzn:
To fix the security issues with some of the transitive dependencies


cthoyt:
Does this come with some explicit changes to the pyproject.toml? I don’t know how to review a lock file. I don’t think they are meant to be read by humans

cthoyt (approved):

PR #669, opened 2026-07-20
Remove `sssom:superClassOf`.

### LINKED ISSUE #559: What is `sssom:superClassOf`?
Wandering through SSSOM-Py’s code, I serendipitously found (in the [`inverse_map.yaml`](https://github.com/mapping-commons/sssom-py/blob/master/src/sssom/inverse_map.yaml) file) that SSSOM-Py considers that the `rdfs:subClassOf` mapping predicate has an inverse predicate, `sssom:superClassOf`.

There is _no_ mention of `sssom:superClassOf` in the entire SSSOM specification and documentation. Even the page about chaining rules, which has a section about [mapping inversion](https://mapping-commons.github.io/sssom/dev/chaining-rules/#inverse-rules), is silent about it. Even worse, [sssom:superClassOf](https://w3id.org/sssom/superClassOf) does not resolve to anything (the URI just fall backs to the SSSOM homepage).

I question the usefulness of a `sssom:superClassOf` predicate. Yes, I can imagine that some people would like to invert a `rdfs:subClassOf` predicate, and may regret the absence of a suitable inverse predicate in the `rdfs:` namespace, but then I would argue that those people should not use `rdfs:subClassOf` in the first place. They should use something like `skos:broadMatch` instead, which does have a standard inverse predicate (`skos:narrowMatch`).

The existence of that predicate is mentioned in a couple of issues in this repo, notably #38, where @matentzn [says](https://github.com/mapping-commons/sssom/issues/38#issuecomment-771094202) he is “using it in COB”. That may have been true back then, but nowadays the only trace of `sssom:superClassOf` in COB is in the [`obo-bridge.md`](https://github.com/OBOFoundry/COB/blob/master/docs/obo-bridge.md) document – a document that seems outdated given that it references a `cob-to-external.tsv` ”source of truth” that no longer exists in COB since [commit f2782d8](https://github.com/OBOFoundry/COB/commit/f2782d893faa87a2a9de41a1884e849d8bf628b9#diff-f0b5ca2afb1a42008859060884755c9c298d43bc2dcae87484dea52d06043011), in October 2024.

If `sssom:superClassOf` was a temporary hack that was solely needed by COB and that COB no longer needs it, then I believe it should be removed from SSSOM-Py.

If, on the contrary, people truly think a `sssom:superClassOf` predicate is useful, this should be discussed somewhere in this repo, and then the predicate should be clearly documented somewhere.

gouttegd:
The `rdfs:subClassOf` predicate has no standard inverse predicate, but for a while SSSOM-Py has been treating it as being the inverse of a completely made-up `sssom:superClassOf` predicate, which has never been documented anywhere.

This used to be needed specifically for the COB ontology, but this is no longer the case and this predicate should now disappear. It is completely unknown outside of SSSOM-Py.

Closes https://github.com/mapping-commons/sssom/issues/559

### DIFF
--- a/src/sssom/cliques.py
+++ b/src/sssom/cliques.py
@@ -18,7 +18,6 @@ from sssom.constants import (
     SKOS_CLOSE_MATCH,
     SKOS_EXACT_MATCH,
     SKOS_NARROW_MATCH,
-    SSSOM_SUPERCLASS_OF,
     SSSOM_URI_PREFIX,
 )
 
@@ -58,8 +57,6 @@ def to_digraph(msdf: MappingSetDataFrame) -> "networkx.DiGraph[str]":
                 pi = 0
             elif p == SKOS_BROAD_MATCH:
                 pi = 0
-            elif p == SSSOM_SUPERCLASS_OF:
-                pi = 1
             elif p == SKOS_NARROW_MATCH:
                 pi = 1
             elif p == OWL_DIFFERENT_FROM:
--- a/src/sssom/constants.py
+++ b/src/sssom/constants.py
@@ -37,7 +37,6 @@ EXTENDED_PREFIX_MAP = HERE / "obo.epm.json"
 OWL_EQUIV_CLASS_URI = "http://www.w3.org/2002/07/owl#equivalentClass"
 RDFS_SUBCLASS_OF_URI = "http://www.w3.org/2000/01/rdf-schema#subClassOf"
 RDF_TYPE_URI = "http://www.w3.org/1999/02/22-rdf-syntax-ns#type"
-SSSOM_SUPERCLASS_OF_URI = "http://w3id.org/sssom/superClassOf"
 SKOS_EXACT_MATCH_URI = "http://www.w3.org/2004/02/skos/core#exactMatch"
 SKOS_CLOSE_MATCH_URI = "http://www.w3.org/2004/02/skos/core#closeMatch"
 SKOS_BROAD_MATCH_URI = "http://www.w3.org/2004/02/skos/core#broadMatch"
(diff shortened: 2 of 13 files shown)


matentzn (dismissed):
There is quite a bit of boilerplate code in sssom-py referencing sssom:superClassOf: https://github.com/search?q=repo%3Amapping-commons%2Fsssom-py%20sssom%3AsuperClassOf&type=code

Just FYI. Happy with this to be merged.

gouttegd:
Removed all traces of `sssom:superClassOf`. This notably involved the complete removal of the `_flip_subclass_assertions()` method in `src/sssom/parsers.py`, since that method makes no sense without `sssom:superClassOf`. It was a “private” method anyway (according to Python conventions), so its removal shouldn’t impact any downstream code.

This also involved re-generating all the validation test files derived from `cob-to-external.tsv`.

FWIW, I now question the usefulness of all tests that use this `cob-to-external.tsv` file, since they are testing for a very specific use case that no longer exists.

matentzn (changes requested):
I am not requesting changes right now - just blocking the merge to give the community a chance to intervene. Lets revisit this PR a little later towards the end of August; I am not against its contents, just against introducing breaking changes like this w/o warning.

If you are in a rush, I can be convinced to merge faster.

gouttegd:
> Just blocking the merge to give the community a chance to intervene.

Fine with me, but anyone who will object to removing all traces of `sssom:superClassOf` will have to

* make a case that `sssom:superClassOf` is actually useful and should _not_ be removed;
* commit to properly document this predicate in SSSOM documentation.

> just against introducing breaking changes

This will only break things for people who have been using `sssom:superClassOf`. That is the point.

matentzn:
> Fine with me, but anyone who will object to removing all traces of sssom:superClassOf will have to

Full agreed

> This will only break things for people who have been using sssom:superClassOf. That is the point.

I concede the wider point as being the logical consequence of merging the PR but object to "That is the point" :P The point is to ensure supported features are well documented and motivated.

gouttegd:
> The point is to ensure supported features are well documented and motivated.

Hence the removal of a feature that is neither well documented nor motivated.

matentzn:
I concede! 😁

cthoyt (approved):

cthoyt:
After this is done, there is only one public usage of sssom:superClassOf and it's in the COB docs (not the actual data) which should also get updated:

https://github.com/search?q=sssom%3AsuperClassOf%20-repo%3Amapping-commons%2Fsssom-py%20-is%3Afork&type=code

gouttegd:
@cthoyt That COB document is completely obsolete. It is entirely about a `cob-to-external.tsv` component (which the document claims is the “source of truth” for where root terms from OBO ontologies should be placed in the COB hierarchy) that **no longer exists**  since 2024.

cthoyt (approved):

matentzn (approved):

Issue #670, opened 2026-07-24
Is the `cliquesummary.yaml` schema still needed?


gouttegd:
SSSOM-Py’s source code contains a small LinkML schema called [`schema/cliquesummary.yaml`](https://github.com/mapping-commons/sssom-py/blob/master/schema/cliquesummary.yaml), that supposedly defines a “data dictionary for clique summaries”. That schema is seemingly used to auto-generate the Python source file [`src/sssom/cliquesummary.py`](https://github.com/mapping-commons/sssom-py/blob/master/src/sssom/cliquesummary.py).

However:

(1) That Python file is not used from anywhere within SSSOM-Py. In particular, the code in [`src/sssom/cliques.py`](https://github.com/mapping-commons/sssom-py/blob/master/src/sssom/cliques.py) (which is behind the `sssom cliquesummary` subcommand) makes absolutely no use of it.

(2) The data structure _actually_ used by the `cliquesummary` command (and therefore the columns in the generated TSV file) differs from what is described in the schema. For example, it contains keys such as `num_mappings`, `sources`, `num_sources` that are _not_ present in the schema.

Overall, both the `cliquesummary.yaml` schema and the derived `cliquesummary.py` Python file seem to be completely obsolete. Maybe the Python file does no harm, but I would argue that providing a schema that does no longer match the data structure produced by the `cliquesummary` command is misleading at best, actively harmful at worst.

The schema should either be updated to correctly reflect the output of the `cliquesummary` command, or be removed altogether.

gouttegd:
Answered by #672.

PR #672, opened 2026-07-25
Cleanup obsolete files.

gouttegd:
This PR removes:

* the `odk.bat` script, which should not be needed for anything (even if someone needed to run some commands from this repo through the ODK – and I don’t see why would anyone ever need to do that –, `odkrun` is a far better option);
* the `environment.sh` script, which should no longer be needed especially after the migration to `uv`;
* the `cliquesummary.yaml` LinkML schema, which is obsolete (#670);
* the corresponding `cliquesummary.py` Python file, which is no longer used (#670).

The `MANIFEST.in` file is updated accordingly.


cthoyt (dismissed):
I agree with removing these files, which don't appear to be used/useful. However, getting nico's approval is a good idea since he might know why they're there in the first place

matentzn (dismissed):
I assume you made sure src/sssom/cliquesummary.py is really not referenced anywhere in the lib, if so, ok.

cthoyt:
check https://github.com/search?q=repo%3Amapping-commons%2Fsssom-py%20cliquesummary&type=code to find some more things that can be removed

cthoyt (approved):

